PR #11, opened 2016-12-19
Allow passing custom className to FormInput

chibicode:
Sometimes, I'd like `FormInput` to be an inline DOM element (by default, it uses `div` tag, so it's a block element) - I can add CSS targeting `.FormInput` class, but I think allowing users to customize the class name of `FormInput` seems like a more flexible solution.


tannerlinsley:
Nicely done!

PR #21, opened 2017-01-06
Fix an issue where `state` and `props` didn't pass through `validate` function

jihchi:
In our defined `validate` method in `Form`, we couldn't get `state` and `props` argument.


tannerlinsley:
Awesome find 👍 Thanks @jihchi

PR #24, opened 2017-01-09
initial stab at HOC exports

SeeThruHead:
I haven't tested anything out but the Form Component so far, will do that shortly. But I thought I'd put this up earlier rather than later.

This exports Higher Order Component versions of Form, FormDefaultProps, FormField, FormInput for better interop with an HOC based workflow like recompose.

### DIFF
--- a/src/hoc.js
+++ b/src/hoc.js
@@ -1,0 +1,36 @@
+import React from 'react';
+import Form, { FormDefaultProps } from './form'
+import FormField from './formField'
+import FormError from './formError'
+import FormInput from './formInput'
+// Inputs
+import Select from './formInputs/select'
+import Checkbox from './formInputs/checkbox'
+import Textarea from './formInputs/textarea'
+import NestedForm from './formInputs/nestedForm'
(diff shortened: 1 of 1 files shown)


tannerlinsley:
This is super cool and very lightweight.  I wonder if you would need anything special at all for the form inputs, since they don't expose an api at all.  They consume it under the hood.  Just pass props.

SeeThruHead:
Do you mean the FormInput component or the Select, Checkbox etc components?
I didn't alter the latter at all, just re-export

tannerlinsley:
Perfect. 👌 so do you think this is good to commit?

SeeThruHead:
i need to write some tests and docs first i think. Really busy right now but i'll keep the pr up to date

tannerlinsley:
I think this would be great to include.  Let's get this stuff on this PR as well:
- [ ] Documentation
- [ ] Storybook Entry w/ code

tannerlinsley:
@SeeThruHead Closing this for now. Feel free to reopen anytime :)

Issue #31, opened 2017-02-01
Missing Import RadioGroup in Annotated Example


h9h:
Hi Tenner,

the annotated example needs the import RadioGroup together with Radio:
`import { Form, ..., RadioGroup, Radio, ... } from 'react-form'
`

And in the jsx the element needs to be changed from

`<radiogroup ... />` to `<RadioGroup ... />`.

Nice lib! Thanx

Michael

tannerlinsley:
Thanks for the suggestion!  Could you perhaps submit a PR to have those changes merged?

PR #45, opened 2017-03-13
Add missing import on RadioGroup/Radio example

AverageZ:
#31 

Noticed this when implementing some radios in a form, thought it might mitigate some questions. Thanks for the great library 👍  Saves me a ton of time.


tannerlinsley:
Nice catch!

PR #48, opened 2017-03-21
Add support for classes (React Components) as children

Qrysto:
If `FormField`, `FormInput` and `Form` can accept a function that returns a JSX element as `children`, why couldn't it also accept a React Component as `children`? "A function that returns JSX element" **is** exactly a stateless functional component, isn't it?

### DIFF
--- a/src/form.js
+++ b/src/form.js
@@ -213,8 +213,8 @@ export default React.createClass({
       ...this.state,
       ...this.getAPI()
     }
-    const { component, children, ...rest } = props
-    const resolvedChild = typeof children === 'function' ? children(rest) : children
+    const { component, children: Child, ...rest } = props
+    const resolvedChild = typeof children === 'function' ? <Child {...rest} /> : Child
     const RootEl = component
     if (!RootEl) {
       return resolvedChild
--- a/src/form.js
+++ b/src/form.js
@@ -214,7 +214,7 @@ export default React.createClass({
       ...this.getAPI()
     }
     const { component, children, ...rest } = props
-    const resolvedChild = _.normalizeComponent(children, rest)
+    const resolvedChild = _.normalizeComponent(children, formAPI)
     const RootEl = component
     if (!RootEl) {
       return resolvedChild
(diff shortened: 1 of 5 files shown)


Qrysto:
With the old way (`children(props)`) I got the error `Cannot call a class as a function` (from Babel if I remember correctly). You can test it again to confirm.

Qrysto:
@tannerlinsley is the update ok?

tannerlinsley:
Can you confirm that it works for all three component types? (stateless function, React.createClass(), and es6 class syntax)

Qrysto:
After looking more carefully, I realized that `Object.getPrototypeOf(Comp).isReactComponent` is kinda weird. I think `Object.getPrototypeOf(Comp)` will return `Function.prototype` and there probably won't be an `isReactComponent` property. Have you ever get this method working? I think my previous commit already works for all cases

Qrysto:
I've just tested, the new way with `normalizeComponent` method doesn't work. I've reverted to the old way and fix a bug and it works well with all 4 cases: normal function, arrow function, ES6 class, and `React.createClass`

Qrysto:
I have tested and confirm this works with all cases. Could it be merged?

tannerlinsley:
I would like to see a `normalizeComponent` utility be made to handle this, much how it is done in React-Table.  This way, the user can pass pretty much any representation of DOM and it will work, whether it is a class, stateless functional component, function returning a primitive, or even JSX that has already been rendered or compiled.

Qrysto:
@tannerlinsley as I have mentioned above, my PR works with all 4 cases:

- normal function,
- arrow function,
- ES6 class
- `React.createClass`

If you want it to also work with React elements, I could add some code. But the `normalizeComponent` utility you've referenced doesn't work for me, not sure if it works for you somehow.

Qrysto:
@tannerlinsley please check my new commit. You can now also pass a React element

tannerlinsley:
Where are we at on this?  I would like to see some examples using this before we merge :)

PR #53, opened 2017-03-30
Pass checked true/false to Checkbox

sukharenko:


tannerlinsley:
Why is this needed? I am pretty sure react prefers a boolean here.

Issue #54, opened 2017-04-13
React.createClass and React.PropTypes are deprecated


thchia:
In React 15.5.0, createClass and PropTypes are deprecated.

https://github.com/facebook/react/blob/master/CHANGELOG.md#1550-april-7-2017

PropTypes should now come from 
https://www.npmjs.com/package/prop-types

abhimanyu1289:
Looks like this is part of v1.2.6. 
But I am still getting warnings for createClass and PropTypes.

thchia:
I believe 1.2.6 is still using createClass and React.propTypes. It's been mentioned that the next version will handle the deprecations.

danielo515:
1.2.6 is indeed using prototypes . Please release 1.2.7 so we can get rid of this messages

PR #55, opened 2017-04-13
Update PropTypes and createClass.

thchia:
Quick and dirty fix for #54 to prevent console errors, but not sure what future support plans for the create-react-class library are.


codinronan:
@thchia Long term, this should use standard Javascript classes, as the browser warning notes. @tannerlinsley I don't know if you have a plan to migrate to that.

thchia:
@codinronan Yup you're right. I don't have the time to rewrite using the standard classes, this PR is just to smooth the transition.

tannerlinsley:
I plan on rewriting to es6 classes as soon as possible. But this looks good for the interim.

Issue #56, opened 2017-04-17
Checkbox component does not handle getValue properly


joshuamanns:
I'm unable to create a controlled checkbox using any approach....
Setting the `checked` property does nothing.
Setting the `value` property does nothing.
Setting `defaultChecked` conflicts with controlled/uncontrolled react requirements.

The only solution I've found is modifying the source so that line 47: `checked: getValue(false),` is changed to `checked: getValue(),` and managing the Checkbox's `checked` prop with state and an onChange handler.

joshuamanns:
@tannerlinsley I'm getting mixed results here. If I double click the checkbox it will sometimes toggle, but otherwise it does not function... Could you please advise on why this is not behaving as expected? Are there known compatibility issues with other libraries/dependencies? It appears to only happen within the context of my project, even when copy/pasting your storybook code verbatim.

tannerlinsley:
I will take a look at this tomorrow morning as I run through issues :)

joshuamanns:
@tannerlinsley After some cursory glances, it looks like your library might take issue with Fastclick's synthetic events. It's strange, because this shouldn't be an issue on Desktop where I'm experiencing the problem. Do you have any insight here?

bruceharris:
@joshuamanns @tannerlinsley I'm experiencing similar trouble; any update on this issue? thanks

tannerlinsley:
I've added fast click to the documentation and it seems to be working just fine :)

PR #57, opened 2017-04-17
Fixed getValue for Checkbox. Updated README for how to retrieve check…

joshuamanns:
Please accept this PR to get Checkbox functioning properly. See issue #56


SeeThruHead:
Can you go into a little more detail behind the motivation for this change? None of the other inputs have a defaults prop as defaults are handled by defaultValues on the form itself.

joshuamanns:
@SeeThruHead The current implementation seems to have a conflict where defaultValue is respected for initial state, but does not allow the Checkbox to be a controlled component. You can see it in origin/master stories/Checkbox.js - at least on Chrome latest, checkboxes are not functioning.

tannerlinsley:
Please don't commit `lib/` or other dist files :)

SeeThruHead:
Am I correct in thinking you cannot make any of the shipped inputs "controlled"? I think that is the case. Do you have a concrete use-case that you're trying to meet? I think there's likely a way to do what you want without this PR, but I would need to know more about your use-case to advise.

tannerlinsley:
`/lib` files are still present in the PR

tannerlinsley:
This is fixed in the latest version.  `defaultValues` is now used to determine if a Checkbox is checked by default now.

PR #62, opened 2017-05-03
Added inputRef prop

hosmelq:


tannerlinsley:
What do you use this ref for?

hosmelq:
For example add focus to an input programmatically.

tannerlinsley:
Good point :)  Could you please add this to the rest of the input types?

hosmelq:
Sure, when i have some time i'll update the PR :)

3bola:
Can we get this in?

hhff:
I have a use-case for this too!

warrenmcquinn:
Hey, wondering if this inputRef will be added to the Styled components at any point (maybe in 3.0)

sanderfish:
Is there any way to use this?

PR #64, opened 2017-05-09
Fix getValue() for textarea

### LINKED ISSUE #41: ResetForm not clear value of Textarea
I found a bug in the resetForm() method that does not reset Textarea value, apparently due to a missing  empty string in the component getValue() method

```
  return _react2.default.createElement('textarea', _extends({}, rest, {
        value: getValue(), // should an empty string go here to for the method to work?
        onChange: (0, _util.buildHandler)(onChange, function (e) {
          return setValue(e.target.value, noTouch);
        }),
        onBlur: (0, _util.buildHandler)(onBlur, function () {
          return setTouched();
        })
      }));
```

DjLeChuck:
- Use an empty string as for text
- Fix #41

### DIFF
--- a/src/formInputs/textarea.js
+++ b/src/formInputs/textarea.js
@@ -19,7 +19,7 @@ export default function FormInputTextarea ({
         return (
           <textarea
             {...rest}
-            value={getValue()}
+            value={getValue('')}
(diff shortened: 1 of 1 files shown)


tannerlinsley:
This is implemented in the latest release

PR #68, opened 2017-05-12
Resolves #67 - Add strict value support for select options

### LINKED ISSUE #67: Select options should have explicit value attribute
Given the following options:

```
const options = [{
  'label': 'A',
  'value': 'a'
}, {
  'label': 'B',
  'value': 'b'
}]
```

The resulting Select would look something like:

```
<select>
  <option value="0">A</option>
  <option value="1">B</option>
</select>
```

Whereas the expected and proper result looks like:

```
<select>
  <option value="a">A</option>
  <option value="b">B</option>
</select>
```

It's a very minimal alteration for which I already have a branch, and will PR momentarily.

thomas-p-wilson:
Select options now prefer to use the provided value for both the key and value attributes. If no value is given, defaults to ordinal position within the list.

### DIFF
--- a/src/formInputs/select.js
+++ b/src/formInputs/select.js
@@ -37,8 +37,8 @@ export default function FormInputSelect ({
             {resolvedOptions.map((option, i) => {
               return (
                 <option
-                  key={i}
-                  value={i}
+                  key={option.value || i}
+                  value={option.value || i}
                   disabled={option.disabled}
                 >
                   {option.label}


tannerlinsley:
This is implemented in the latest version :)

thomas-p-wilson:
That is only partially correct. https://github.com/tannerlinsley/react-form/blob/master/src/formInputs/select.js#L54 shows that to be the case for the `key` attribute but not the `value` attribute

tannerlinsley:
This is to support complex values.  You cannot use an object as a value, then utilize that value in the onChange handler.  Using the option index is a reliable way to lookup which option value should be passed in the onChange handler, regardless if it is a primitive or complex value.

thomas-p-wilson:
Right, but should the option not exist to do so? Sure, it needn't be
utilized if using options whose values are complex objects, but when using
scalars then sure it makes sense. Conversations are useful. Being told
we're wrong when there is a definite use case for something makes folks
less willing to help. I'll probably be moving on, not that it'll affect
your stats any.

thomas-p-wilson:
That is to say, being told we're wrong, without any discussion whatsoever,
in a terse manner, is not cool.

Issue #96, opened 2017-08-07
nestedform objects change defaultValues object


rowdyroad:
Now, if we have some related object or collection, it will be copy (by extend) to defaultValues by reference. So, when we update nestedform,  defaultValues updates too.

Need to make a deep copy.

PR is coming.

joepuzzo:
React form 2.0.0 makes deep copies at each Redux update. Therefore this is not an issue as we will never override the default values passed in.

PR #97, opened 2017-08-07
* #96 fix. make deep copy for defaultValues;

rowdyroad:


sboudrias-aa:
I think this bug also happens with deep defaultValues (like `Array<Object>`) when you call `resetForm()`

PR #134, opened 2017-10-31
Use is shallow equal instead of JSON.stringify for change detection

tannerlinsley:


tannerlinsley:
Closed in favor of #136

Issue #133, opened 2017-10-27
Converting circular structure to JSON


arolson101:
I am passing a react element as a prop to a custom component like this:
```js
<TextField addonBefore={<div>addon</div>}/>
```

this results in:
```
Uncaught TypeError: Converting circular structure to JSON
    at JSON.stringify (<anonymous>)
    at t.value (preview.bundle.js:formatted:382618)
    at checkShouldComponentUpdate (preview.bundle.js:formatted:135775)
    at updateClassInstance (preview.bundle.js:formatted:136091)
    at updateClassComponent (preview.bundle.js:formatted:136311)
    at beginWork (preview.bundle.js:formatted:136691)
    at performUnitOfWork (preview.bundle.js:formatted:138635)
    at workLoop (preview.bundle.js:formatted:138744)
    at HTMLUnknownElement.callCallback (preview.bundle.js:formatted:127420)
    at Object.invokeGuardedCallbackDev (preview.bundle.js:formatted:127459)
```

in ```checkShouldComponentUpdate ``` it's testing ```JSON.stringify(a) !== JSON.stringify(u)```.  I'm not quite sure why the structure is circular but it would be nice if this was supported in version 2 (it worked in version 1)

joepuzzo:
Hmm you are not alone. Someone else is having similar problem.. i need to revisit why i added that check in shouldComponentUpdate.

joepuzzo:
@arolson101 Hey turns out that optimization was not needed anymore because i updated react form to use react-redux and it no longer used state. You should be all set now, get the latest version 2.1.1  and let me know it this problem is solved.

joepuzzo:
Hmm although removing this may cause some performance issues in IE i think i will disable it by default and add an optimization flag.

joepuzzo:
Ok so i removed this optimization by default but if you need it then pass in optimize prop. V2.1.2

joepuzzo:
Update: working on a solution that adds this optimization by default and does not have circular json issue.

joepuzzo:
Ok so im now taking advantage of https://www.npmjs.com/package/circular-json so dont even worry about that temp solution. Your issue should be fixed now :)

arolson101:
Thank you for your responsiveness!
In my scenario, calling stringify is not an optimization; significant time is spent in the stringify.  Is there a way to disable, or better yet change it to opt-in?  I'd rather see a reference or equality test and be done with it

joepuzzo:
[...]

PR #135, opened 2017-10-31
Support FAC for FormField component

tannerlinsley:


tannerlinsley:
The other thing to consider here (related to #133), is if there is an easier way syntactically to detect changes in what the user is render inside of their form field.

tannerlinsley:
Closed in favor of #136

PR #136, opened 2017-10-31
Form field is now an HOC

tannerlinsley:
This significantly cleans up the syntax for FormField, and really benefits from these changes:

- There is no more "child" component in a FormField
- FormField is solely responsible for providing the `fieldApi` and handling logic for updating performance
- No more laborious proxying of props.
- sCU now uses `Utils.isShallowEqual` instead of `CircularJSON` to detect changes in props. Perf is amazing and follows React's best practices.


tannerlinsley:
Apologies, it looks like Prettify had it's way with some of the files haha. Maybe we should eslint --fix the whole thing.

tannerlinsley:
Fixes #133

joepuzzo:
Hey @tannerlinsley this looks sick.. i only have two things. 
1. Maybe turn off prettify so we can see the actual things that changed in this PR. and we can talk about maybe doing a prettify commit after. That way its easier to look back at this PR and see the core changes instead of the files being filled with "changes" lol.
2. I really like this using the connected pattern. However, it will break many peoples projects ( including mine ) until i make the switch. So we have a few options here. 1: we make it backwards compatible... this may be actually really easy because we could just say `if( formField has children ) then do it the old way` ( we would not call this out in docs, just leave for backwards compatibility ) 2: we make this a minor version update ( but again that has breaking changes soo prob not good )

src/redux/utils.js
```diff
@@ -67,9 +67,19 @@ function get(obj, path, def) {
   return typeof val !== 'undefined' ? val : def;
 }
 
+function isShallowEqual(obj1, obj2) {
+  for (var prop in obj1) {
```
arolson101: what about props in obj2 that aren't present in obj1?

tannerlinsley:
Yeah we should check the length of the keys first.

tannerlinsley:
I've removed the prettified content.  Actually, I straight up had to revert and just do it again. But no biggie. Github diffing still doesn't handle indentation changes very well though...

PR #137, opened 2017-10-31
Normalize input fields' callbacks and props

### LINKED ISSUE #132: React 16 issue with Text field
```
Failed form propType: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`
```

On wonder if its because of the onInput rather than onChange usage.

 https://github.com/react-tools/react-form/blob/master/src/components/form-fields/Text.js#L35

tannerlinsley:
- All inputs now internally use `onChange` instead of `onInput` as per recommended via React best practices.
- All inputs now consistently use and proxy `onBlur`.
- All inputs now consistently use a hybrid `onChange(newValue, event)` callback, instead of a separate `fieldDidUpdate` and `onChange` prop.
- All inputs now apply `...rest` props so as to not accidentally clobber internal props.

Fixes #132

### DIFF
--- a/src/components/form-fields/Checkbox.js
+++ b/src/components/form-fields/Checkbox.js
@@ -17,6 +17,7 @@ class CheckboxWrapper extends Component {
 
     const {
       fieldApi,
+      fieldDidUpdate,
       onChange,
       onBlur,
       ...rest
@@ -34,6 +35,12 @@ class CheckboxWrapper extends Component {
         checked={!!getValue()}
         onChange={(e) => {
           setValue(e.target.checked);
+          if (fieldDidUpdate) {
+            if (process.env.NODE_ENV === 'development') {
+              console.warn('fieldDidUpdate has been deprecated in favor of onChange. Please check the latest docs and update your app!')
+            }
+            fieldDidUpdate(e.target.checked);
+          }
           if (onChange) {
             onChange(e.target.checked, e);
           }
--- a/src/components/form-fields/Checkbox.js
+++ b/src/components/form-fields/Checkbox.js
@@ -17,7 +17,6 @@ class CheckboxWrapper extends Component {
 
     const {
       fieldApi,
-      fieldDidUpdate,
       onChange,
       onBlur,
       ...rest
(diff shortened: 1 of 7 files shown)


joepuzzo:
Hey @tannerlinsley This one we may have a big issue with that im happy to work through because you may know more about this than me. 

So we had a huge problem with using onChange instead of onInput in one of our projects. The problem was this: 

We had a very complex form, with multiple iterations that generated nested forms within nested forms etc. This very large form that was built from all these iterations was generated fine, but we noticed a very strange bug. Only the first text field in every nested form would trigger its on change event. Right off that bat this sounded like some sory of issue with react `keys`, but nope! Turns out it was an issue with onChange. I forget exacly where i read up on this, but somehow the way react is taking over the event handlers is causing onChange to not trigger for all those other fields in the nested forms and was a huge blocker for functionality. 

So in summary, i spent a lot of time debugging this strange issue and determined it was the use of on change over onInput, onInput fixed that issue so unless we can figure out another solution im hesitant to change this.

joepuzzo:
Also for backwards compatibility, I like to leave things like `fieldDidUpdate` but just hide them in the docs.

tannerlinsley:
I think the case for normalization and alignment with the React is stronger here. I would be more than happy to help you debug the problem you were having with `onChange` in large forms to make this change happen. `onInput` used to be a token of discussion a while back in older version of react, but as of react 15, they've made it pretty clear that `onChange` is blessed syntax for form inputs. It's really the only way to avoid seeing this:
```Failed form propType: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly```

Can you confirm with these changes that you're still experiencing that bug?

tannerlinsley:
I just noticed that `fieldDidUpdate` is actually undocumented currently. So I would still suggest removing it and documenting the `onChange` prop.

PR #138, opened 2017-10-31
Support React 16

tannerlinsley:


tannerlinsley:
Fixes #138

joepuzzo:
Just want to double check that doing this should not cause issues if projects are still using 15 correct ?

tannerlinsley:
Shouldn't cause any problems at all.

PR #146, opened 2017-11-02
Add getApi prop and docs

tannerlinsley:
allows retrieving an instance of the api from a parent component, much like using a `ref`. :)


joepuzzo:
This looks good. Awesome idea

mulmulovich:
this feature is really useful. is it possible to add it to FormProps type definition?

PR #154, opened 2017-11-07
Add getFieldName to fieldApi for getting field name of custom input

hnq90:
I'm writing a custom input which based on `RadioGroup` component, I couldn't get `field` property from props because `field` was destructed from Component props.

https://github.com/react-tools/react-form/blob/5c7c76cda719256f1401d2aa06c75a1174a9b025/src/components/FormField.js#L65

Here's my implementation:
```
const RadioGroup = FormField(({ options, fieldName, disabled }) => {
  if (!options || options.length === 0) {
    return false
  }

  return (
    <_RadioGroup field={fieldName}>
      {group =>
        options.map((option, i) => (
          <label key={i}>
            <Radio
              group={group}
              className='radio'
              value={option.value}
              disabled={disabled}
            />
            <span>{option.label}</span>
          </label>
        ))}
    </_RadioGroup>
  )
})
```

After adding `getFieldName` to `FormField`, we can retrieve the field name by:
```
const { fieldApi: { getFieldName } } = props
const fieldName = getFieldName()
```

Other solution is changing `<FormComponent fieldApi={fieldApi} {...rest} />` to `<FormComponent fieldApi={fieldApi} {...this.props} />` but I'm not sure which is better.

Please let me know your opinion. Thank you in advance.

### DIFF
--- a/src/components/FormField.js
+++ b/src/components/FormField.js
@@ -84,6 +84,7 @@ function FormField(FormComponent) {
         setSuccess: ( success ) => {
           formApi.setSuccess( field, success );
         },
+        getFieldName: ( ) => field,
         getValue: ( ) => formApi.getValue( field ),
         getTouched: ( ) => formApi.getTouched( field ),
         getError: ( ) => formApi.getError( field ),


joepuzzo:
Im not sure i understand the issue. I think there may be some confusion on how to use custom inputs. Above you are trying to write your own custom FormField.. but you are wrapping it with an existing one. I think this is the main issue. When you return the JSX with the default `RadioGroup` you are returning a `FormField`. If you want to write a custom `FormField` then you should not be wrapping the default ones but instead overriding it from scratch. This way.. you will not need the form field name. The field name should only be used internally by `react-form`. 

Doing what you are doing above is probs going to have some strange side affects because you are wrapping a FormField in a FormField.

hnq90:
@joepuzzo Thanks for explanation. In my PR, I've just added a getter function which could be useful for some cases.
Example: When I create a custom input which has both `formApi` and `fieldApi` and I need to use `addValue`, `removeValue` then `fieldName` must be retrievable from the `fieldApi`.

joepuzzo:
Hmm now im just curious.. can you send code snippet of this example use case. I will probs merge this in and put in my next release.. but i just want to really understand the use case. ( honestly just want to make sure exposing the field name does not promote "bad practices" ) that will lead to people having issues.

hnq90:
@joepuzzo Here's my implementation of Checkbox Group:
```js
onst CheckboxGroup = FormField(({ fieldName, options, fieldApi, formApi }) => {
  const { addValue, removeValue, values } = formApi
  const selectedOptions = values[fieldName] || []

  if (!options || options.length === 0) {
    return false
  }

  return (
    <div>
      {options.map((option, i) => (
        <span
          key={i}
          onChange={e => {
            e.target.checked
              ? addValue(fieldName, option.value)
              : removeValue(fieldName, selectedOptions.indexOf(option.value))
          }}
        >
          <label className='item-check'>
            <input
              type='checkbox'
              className='checkbox'
              defaultChecked={selectedOptions.indexOf(option.value) > -1}
            />
            <span>{option.label}</span>
          </label>
        </span>
      ))}
    </div>
  )
})
```

which requires `addValue` & `removeValue`.

Actually, it has just re-written to:
```js
const CheckboxGroup = FormField(({ options, onChange, fieldApi }) => {
  const { getValue, setValue } = fieldApi
  let selectedOptions = getValue() || []

  if (!options || options.length === 0) {
    return false
  }

  return (
    <div>
      {options.map((option, i) => (
        <span
          key={i}
          onChange={e => {
            if (e.target.checked) {
              selectedOptions.push(option.value)
              setValue(selectedOptions)
            } else {
              const removedIndex = selectedOptions.indexOf(option.value)
              selectedOptions = [
                ...selectedOptions.slice(0, removedIndex),
                ...selectedOptions.slice(removedIndex + 1)
              ]
              setValue(selectedOptions)
            }
            onChange && onChange(e)
          }}
        >
          <label className='item-check'>
            <input
              type='checkbox'
              className='checkbox'
              defaultChecked={selectedOptions.indexOf(option.value) > -1}
            />
            <span>{option.label}</span>
          </label>
        </span>
      ))}
    </div>
  )
})
```

For exposing a getter function for field name is up to you. If not necessary, please close this PR.

joepuzzo:
Another person just asked about this too in an issue. Would it be ok to change this to pass the field down as prop instead of getter? @hnq90

joepuzzo:
or maybe even `fieldApi.field`? Thoughts

joepuzzo:
I just merged it in.. will release in a sec.. we can look to add fieldApi.field or somthing down the road but getter is fine.

joepuzzo:
Working on adding some docs for fieldApi as well.

hnq90:
Thank you. I think it’s better to keep getter instead of property.

joepuzzo:
Ok so i merged the pull request and added some docs... you now have access to  `fieldApi.getFieldName()`! https://react-form.js.org/#/custom-input . This is in latest version 2.7.0

joepuzzo:
And agreed on the getter 👍

PR #168, opened 2017-11-14
Add a DefaultValue prop

SebastiaanNijland:
This allows for a DefaultValue to be passed to text inputs

### DIFF
--- a/src/components/form-fields/Text.js
+++ b/src/components/form-fields/Text.js
@@ -10,14 +10,24 @@ class TextWrapper extends Component {
   render() {
     //console.log('RENDER');
 
-    const { fieldApi, onChange, onBlur, ...rest } = this.props;
+    const {
+      onChange,
+      onBlur,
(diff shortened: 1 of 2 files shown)


joepuzzo:
So I don't think this does what you think it does. The problem with doing this is that the react formApi will not actually register that you set that default value. If you tried to log the `formApi.values` without touching anything, you would actually get `{}` in the logs. The reason it has the `|| ''`is to mitigate an issue where react will throw an uncontrolled input error. I would recomend using the `defaultValues` prop on the form.

hnq90:
I agree. We should use defaultValues for form initialization.

SebastiaanNijland:
Ah yes, I hadn't seen that option. Is it possible to open-source the docs, so the community can add to it?

joepuzzo:
The docs are open sourced. They live right in this project. And default values are already documented :)

Issue #169, opened 2017-11-14
Specifying a string value in fields should not attempt to autoconvert to integer


danbopes:
See this codepen: https://codesandbox.io/s/oq899oy076

You'll notice that when I type something into the third field, it attempts to turn friends into an array. Strict type-checking should occur, and if a number was not passed (1 !== '1'), it should not attempt to convert to an array.

joepuzzo:
Haha yup i knew about this and forgot to update docs. As it stands you cant use numbers as field names. We have had some discussion about this exact thing and i dont know if we are going to support it.

tannerlinsley:
It's considered bad practice generally in javascript to use sequenced numbers as keys in objects. 99% of the time, you should be using an array for the data, or if you absolutely need to keep track of random numbers, it's easy enough to simply prefix your key with something like `item_1` or `n1`.

danbopes:
We're able to pass javascript numbers to the field via react, so ideally, I think we should typecheck to see if it's being passed as a string or as a number. (This ideally should be the default stance).

For compatibility: I suggest we add a strict parameter (or similar), that if set to true, will typecheck as intended.
Ideally: If it's a string, it's an object. If it's a number, it's an array. Perhaps add a strict parameter, that will default to true, but when set to false will fallback to current functionality. How many people are currently passing '0' to this? Every example mostly uses idx or similar, so this shouldn't be a huge impact.

You say 99% of the time, but this was my first project and I'm already running into this issue. We have a number system that was changed at a certain point to allow for letters prepended. Having to prepend 'item_' or similar seems like a wonky workaround, as I then need to strip it out later down the line.

joepuzzo:
Hmm.. while I do agree with @tannerlinsley the number system is a good example of where numbers could make sense for field names.

tannerlinsley:
I would like to see your use-case, as though your example shows a technical use-case, that doesn't provide the whole story.

I will attempt to refactor the internal field path generating functions to allow this. Stay tuned.

tannerlinsley:
Fixe via #170

joepuzzo:
This is now supported in the latest version!

PR #170, opened 2017-11-15
Better field path creation

tannerlinsley:
Field paths now respect using integers and bracket notation for array indices, and likewise, treat dot notation as a property location, even when parseable to an integer.


tannerlinsley:
Fixes #169

tannerlinsley:
Looks good

joepuzzo:
I for-see a few issues getting logged because someone is bound to be using `friends.0.firstName` lol

PR #179, opened 2017-11-18
Avoid to merge old and new errors/warnings/successes

alexmaroz:
This changes to prevent the merging old and new errors/warnings/successes.

For example,
My validation framework (modified `JOI`) for `validateError` callback returns object with only fields with errors:
`{
  firstName: 'First Name is required'
}`
not like 
`{
  firstName: 'First Name is required',
  lastName: null
}`

If form is valid then `validateError` callback returns empty object `{}`.

Original implementation does not remove old error messages.

Thanks

### DIFF
--- a/src/redux/ReducerBuilder.js
+++ b/src/redux/ReducerBuilder.js
@@ -325,10 +325,10 @@ const removeAsyncSuccess = ( state, action ) => {
 
 const validate = ( state, action, validateError, validateWarning, validateSuccess ) => {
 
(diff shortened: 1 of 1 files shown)


joepuzzo:
Hmm this is actually a big functional change.. well this may be a breaking change for your implementation it was a decision that was made when moving to V2. I cant remember off the top of my head, but there were reasons for doing it this way. I can take another look at why when i get a chance but I don't know if im to merge this pull request as it may break other users code that use V2 the way it is now.

PR #194, opened 2017-12-05
Add Typescript definitions

Arcath:
Adds Typescript definitions to replace the current ones at `@types/react-form` which are for 1.3

I have a PR open on Definitely Typed to update `@types/react-form` here https://github.com/DefinitelyTyped/DefinitelyTyped/pull/21981 but I'm not sure if @cameron-mcateer is around to approve the update.

I think in the long run providing types in a package is better than having an external dependency that can fall behind.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "React Form is a lightweight framework and utility for building powerful forms in React applications.",
   "main": "dist/index.js",
   "files": [
-    "dist/"
+    "dist/",
+    "types.d.ts"
   ],
   "scripts": {
     "test": "NODE_ENV=test mocha-webpack --opts test/mocha-webpack.opts",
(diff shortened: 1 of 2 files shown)


joepuzzo:
I apologize as i am unfamiliar with type script so bare with me. Im more of a vanilla JS / ECMA script guy :). Have you tested this with your code and does it work fine? and also should I have any concerns with this addition affecting existing users? if all that is good i would be happy to merge this but im def not the guy to review it as im not a typescripter lol.

Arcath:
Yeah I wrote these types for my own code after I found that `@types/react-form` was out of date.

If you look at the PR for DefinitelyTyped I created a test file that has all the code from the docs in it which should ensure that everything is typed correctly. I didn't add that here as adding typescript a dev dependency seemed a little over the top.

It shouldn't affect existing users unless something else uses the `types` key in `package.json`

Arcath:
Things are moving on the types PR now, I will close this off and focus my efforts there instead.

Issue #204, opened 2017-12-09
Async validator does not work after error


zalishchuk:
Why async validator stops working after resolving first error?

https://github.com/zalishchuk/react-form-async-validators

Try to type anything into field, it run validator every time, but if you will type `error` (to resolve object with error) it will fail and die then, validator will never run again. Bug or feature and how can I disable this logic?

**Update:** fixed that by adding synchronous validator.

---

Another question is how can I do things in custom text input `onChange` method after field finished validating. Something like `waitValidation` method in `fieldApi` to use it like this:

```js
const onChange = async event => {
  event.persist();

  const {
    onChange,
    fieldApi: { setValue, waitValidation },
  } = this.props;

  setValue(event.target.value);
  
  await waitValidation();

  onChange(event); // will be executed after field finished validating
};
```

zalishchuk:
Since async validations are [running](https://github.com/react-tools/react-form/blob/master/src/components/ReduxForm.js#L203) on `setTouched` event, method should be asynchronous. This feature can be used in custom inputs which have async validations.

```jsx
import React, { Component } from 'react';
import { FormField } from 'react-form';

@FormField
class Text extends Component {
  render() {
    const {
      fieldApi: {
        getValue,
        setValue,
        setTouched,
        getError,
        getWarning,
        getSuccess,
      },
      onChange,
      onBlur,
      ...rest
    } = this.props;

    return (
      <input
        {...rest}
        value={getValue() || ''}
        onChange={async event => {
          event.persist();

          setValue(event.target.value);

          await setTouched(); // this will trigger and wait for validations

          // actual values after validation ended
          console.log(getError(), getWarning(), getSuccess());

          if (onChange) {
            onChange(event);
          }
        }}
        onBlur={event => {
          setTouched();

          if (onBlur) {
            onBlur(event);
          }
        }}
      />
    );
  }
}

export default TextAsync;
```

joepuzzo:
Hmm i dont think this is what we want. `setTouched` should only care about one thing. Setting the field to touched. And on change should only care about one thing, did the field change. A side affect, yes is that async validation should get triggered, but that does not change the fact that the field changed.

zalishchuk:
What do you propose to do? How can we get errors, warnings, successes in fields which have asynchronous validation? My solution does not break anything, just gives the opportunity to wait before validation is complete and get actual information about field.

joepuzzo:
[...]

PR #205, opened 2017-12-10
setTouched async support

zalishchuk:
#204


joepuzzo:
setTouched should only care about one thing. Setting the field to touched. And on change should only care about one thing, did the field change. A side affect, yes is that async validation should get triggered, but that does not change the fact that the field was touched.

PR #206, opened 2017-12-11
Make it faster for big forms

krizka:
This try-catch impacts hugely on performance for big forms. After profiling and removal, my input updates become faster about 3 times. But performance is still not very good due to full form re-rendering

### DIFF
--- a/src/redux/utils.js
+++ b/src/redux/utils.js
@@ -77,9 +77,7 @@ function get(obj, path, def) {
   }
   const pathObj = makePathArray(path);
   let val;
-  try {
-    val = pathObj.reduce((current, pathPart) => current[pathPart], obj);
-  } catch (e) {}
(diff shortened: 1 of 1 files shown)


joepuzzo:
Hey thanks for pointing this out. I will verify that it all works in the AM and pull.

krizka:
Thanks!

tiffwu:
This update broke my form.

krizka:
Please explain, what happened? Some code?

tiffwu:
Sometimes `current` is null, which results in 
```
Uncaught TypeError: Cannot read property '0' of null```

LyhjeHylje:
This happened to me also: 2.14.1 works but 2.14.2 throws intermittent "Cannot read property of null" errors.

krizka:
@LyhjeHylje @tiffwu yeah, im really sorry, I led you to this. let we wait a bit for this to be fixed with @tiffwu PR.

PR #209, opened 2017-12-14
Prevent error that had been previously swallowed by try/catch

tiffwu:
https://github.com/react-tools/react-form/pull/206 removed a try/catch in the `get` utility function. This try/catch had been previously swallowing errors when it tried to access a value on `null`.


joepuzzo:
Thaks for pointing this out. I will merge and publish when i get to my laptop

joepuzzo:
Fixed in latest version!

tiffwu:
👍

PR #223, opened 2018-01-07
Fix a few documentation typos

isc:


isc:
Didn't think about taking a look at the other open pull requests before opening mine. I'm closing it as other are more complete.

PR #250, opened 2018-02-24
fix preventDefault behavior

mjhm:
It looks like the `preventDefault` behavior got confused with the update to 3.0.

### DIFF
--- a/demo/src/Intro.js
+++ b/demo/src/Intro.js
@@ -778,16 +778,16 @@ const FormProps = () => (
         </tr>
         <tr>
           <th scope="row">
-            <code>dontPreventDefault</code>
+            <code>preventDefault</code>
           </th>
           <td>
             <pre>bool</pre>
           </td>
           <td>no</td>
           <td>
-            Pass this in if you want to prevent the form submission from "preventingDefault". You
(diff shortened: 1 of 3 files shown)


mjhm:
@joepuzzo 

> cool and yeah it looks like all your pull request does is flip the logic of the prop.. not sure we need that

Actually the logic was flipped on this previous commit https://github.com/react-tools/react-form/commit/894296f86828b5d1f99c16c9231b4ade5d42a61c#diff-07a63ede6fe55fca89398d42da737369R272 by @tannerlinsley 

My fix was for a problem where `defaultValues.preventDefault` was being set here: https://github.com/react-tools/react-form/blob/master/src/components/Form.js#L34 
but was being clobbered here: https://github.com/react-tools/react-form/blob/master/src/components/Form.js#L456

fabienjuif:
Good catch @mjhm I was getting confused 😅

joepuzzo:
So sorry i just realized i screwed up lol thought you were just flipping hte logic and realized i should have looked closer.. i will pull in right away

joepuzzo:
@mjhm Thanks for the fix!

PR #277, opened 2018-03-13
Add MultiSelect Component.

passionInfinite:
@joepuzzo I have added MultiSelect component to handle the multiple selection. I have added two test cases but want your help to add more and check the problems.
The <MultiSelect/> component takes options (all the options for select)  as props and it will show the selected options if we have defined in defaultValues. @joepuzzo let me know your thoughts!

### DIFF
--- a/demo/src/Examples.js
+++ b/demo/src/Examples.js
@@ -19,6 +19,7 @@ import MainContent from './MainContent'
 import DynamicForms from './DynamicForms'
 import ArrayOfNestedFields from './ArrayOfNestedFields'
 import Tester from './Tester'
+import MultiSelectExample from "./MultiSelectExample";
 
 class Examples extends Component {
   render () {
@@ -66,6 +67,9 @@ class Examples extends Component {
               <Link to="/nested-async" className="nav-link">
                 Nested async validation
               </Link>
+              <Link to="/multi-select" className="nav-link">
+                Multi-Select
+              </Link>
               {/* <Link to="/big-complex" className="nav-link">
                 Big complex form
               </Link> */}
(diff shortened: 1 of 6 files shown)


joepuzzo:
Hey taking a look now.. sorry for the delay

joepuzzo:
So i pulled this down and started playing with it and im not sure it woks as you expect. I would recomend adding this as an example in the docs under `FieldInputs` and verify it does what you think it does. Also make sure you keep the interface the same as the other select. Your select, accepts an array of options that differs from react-forms select. I have been trying to play around with this idea of multiselect myself and cant figure out how to detect if something has been deselected.

passionInfinite:
Hey buddy, I am out right now but see I have added one test case that
default values is ["xyz"] and options has ["xyz", "pqr"] so I simulated on
change event and selected pqr so in value of select is should have two
values now pqr and xyz. Similarly, I have done it again to deselect it. So
basically it toggles, if it is already selected then it will unselect if
selected again. I will try to add to the doc.

passionInfinite:
@joepuzzo I have added a simple demo to the docs. Check it out, what I was thinking related to multiselect.

passionInfinite:
@joepuzzo any updates on this?

joepuzzo:
Hey so i just pulled the code down and it looks like the example is not working as expected... I try and select things and it is jumping around a lot.

joepuzzo:
I did this in my new lib `informed` I will update this when i release V1

joepuzzo:
Check out informed. I added a multiselect!!!  https://joepuzzo.github.io/informed/?selectedKind=Inputs&selectedStory=Multi%20Select%20Input&full=0&addons=0&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel

PR #278, opened 2018-03-13
Add values prop

tannerlinsley:


src/components/Form.js
```diff
@@ -242,7 +245,13 @@ class Form extends Component {
     })
   }
 
-  setAllValues = values => this.props.dispatch(actions.setAllValues(values))
+  setAllValues = (values = {}) =>
+    this.props.dispatch(
+      actions.setAllValues({
+        ...this.props.defaultValues,
```
joepuzzo: Why did we have this laying them on top of default values again? Does it make sense to remove?

src/components/NestedField.js
```diff
@@ -59,25 +62,41 @@ class NestedField extends React.Component {
       getFullField: field => [fullField, field]
     }
 
+    const proxySubField = method => (subField, ...args) =>
+      method([fullField, subField].filter(Boolean), ...args)
```
joepuzzo: Can you explain this? we never needed subField before? where does that come from. I thought that was the job of overidding the getFullField on line 62 above?

src/components/Form.js
tannerlinsley: I think it makes sense to keep. Default values while providing a `values` prop is still useful in case the `values` prop doesn't contain required default values.

src/components/NestedField.js
tannerlinsley: The reason this is necessary is because the api for a NestedField should resemble that of a root form, not a field. Since we don't have any areas where we actually use the form api from a nested field, we missed that. This makes interacting with the nested field api actually useful.
```javascript
<NestedField field='profile' render={(api => (
  <button onClick={() => api.setValue('name', 'Tanner')} />
)} />
```
or
```javascript
<NestedField field='users' defaultValue={[]} render={(api => (
  <button onClick={() => api.addValue(undefined, {...} )}>Add User</button>
)} />
```

joepuzzo: Ahh ok cool

joepuzzo:
Do you think you could just add this to docs before we merge ?

tannerlinsley:
Just added some much needed logic to keep Form/nested level errors from overrideing nested errors. Basically, even if you registered a validator (or had a validator return undefined) at the nested/form level, it would clobber all of the sub-field errors that may have occured. Very dumb. Now form level and nested level errors check to make sure they won't clobber lower-level errors before they even run. Super cool and solves a lot of problems I was having over the last few days.

As for that other api error you slacked me, I'm still working on it. But go ahead and test this out :)

src/components/FieldApi.js
```diff
@@ -7,32 +7,41 @@ class FieldApi extends React.Component {
...
     // Get the full field name
     const fullField = formApi.getFullField(field)
 
     // Get the node of that field
     const node = formApi.getNodeByField(fullField)
 
-    // Get the field state and api
+    // Get the field api
     const fieldApi = node ? node.api : {}
-    const fieldState = node ? node.getState() : {}
```
joepuzzo: I had already fixed this issue in master, i think the issue was that before nested fields was not implimenting the getState funciton and in master i think its added

FredyC:
@tannerlinsley Would you be that kind and explain why is `fieldApi.removeValue` using `formApi.addValue` internally? It was there correctly before this PR, but for some reason, it was changed here and there is no explanation why is that.

PR #280, opened 2018-03-15
Allow passing false to select placeholder option

AlexandreBernard:
Hi,

I'd like to have the possibility not to have any placeholder on selects for fields that always have a default and no possible empty values

I think passing placeholder={false} to selects makes sense

if this is ok for you can we publish it as a 2.x release?


joepuzzo:
This is in V3.4.0! I also added docs for it

AlexandreBernard:
3.4? I really would need something under a 2.x release because I've built many things on top of that and can't really switch now

PR #283, opened 2018-03-18
Radio input value

EscoCreeks:
Hello, while testing Radio inputs using capybara, we could see that the inputs all have `value="on"`.

screen shot 2018-03-18 at 20 08 30

We would like to have the value passed to the input vanilla.


joepuzzo:
Sorry it took so long to reapond to this. I will validate and pull it in the second i get back to my laptop 👍🏼

joepuzzo:
Looks good.. look for it in latest release

PR #284, opened 2018-03-18
Fixed module references in index.js

szazo:
In 'src/index.js' the 'withNestedField' was imported from
'./components/FieldApi' and 'FieldApi' was imported from './components/withNestedField'


joepuzzo:
Wow nice catch i will get this in the second i get back to my computer !

szazo:
Thank you :)

PR #288, opened 2018-03-24
Add landing page contents and folder

ryanmunger:
Just head into the landing-page dir, run npm i and then npm start. This will spin up the static files using http-server. This can all change eventually, just getting it out there so you guys can pull down and check locally!


joepuzzo:
Hey Ryan. So it looks like you basically added your project directory to our project ( I see a package.json in there lol ). Do you think you could add this as another component in the project instead and actually add it to the demo page as the first page loaded? In other words for this to be megred i should be able to pull down this branch, run the demo normally, and see your page instead of the current one.

ryanmunger:
Good idea. I will just wrap all of the code in this page in a React Component that way it's just plug and play for you guys!

PR #296, opened 2018-03-31
select field: support optgroups

gaetano-guerriero:
Support `optgroup` in `Select` field by accepting an annidated `options` property:
```
<Select options={[
  { label: 'l1', value: 'v1' },
  { label: 'og1', options: [
    { label: 'l2', value: 'v2' },
    { label: 'l3', value: 'v3' },
    { label: 'l4', value: 'v4' }
  ] },
  { label: 'og2', options: [
    { label: 'l5', value: 'v5' },
    { label: 'l6', value: 'v6' }
 ] }
]} 
```

### DIFF
--- a/src/components/form-fields/Select.js
+++ b/src/components/form-fields/Select.js
@@ -5,36 +5,74 @@ import React, { Component } from 'react'
 import withField from '../withField'
 
 class SelectWrapper extends Component {
+
+  constructor (props) {
+    super(props)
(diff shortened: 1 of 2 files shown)


joepuzzo:
Try out https://joepuzzo.github.io/informed. I rewrote this library and will now be supporting `informed` instead. Select was refactored and has a much more intuitive interface

PR #307, opened 2018-04-12
Fixing defaultProps Warning in Select Component

### LINKED ISSUE #306: React 16 Warning on defaultValue props
If I set the `defaultValue` on the Field props, I get the error below. Since all of the form fields are controlled components, we probably should make sure to not pass the `defaultValue` prop down to the actual input field.

Warning: TextWrapper contains an input of type undefined with both value and defaultValue props. Input elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both). Decide between using a controlled or uncontrolled input element and remove one of these props. More info: https://fb.me/react-controlled-components

joshuaai:
This fixes #306 
The approach is to provide an `initialValue` prop for setting the
initial value rather than `defaultValue` since the component is not
uncontrolled. I then check the `selectedIndex` and `initialValueIndex`
(if provided) when setting the `value` prop of the HTML <select> element.

### DIFF
--- a/src/components/form-fields/Select.js
+++ b/src/components/form-fields/Select.js
@@ -30,6 +30,10 @@ class SelectWrapper extends Component {
     const nullIndex = resolvedOptions.findIndex(d => d.value === '')
     const selectedIndex = resolvedOptions.findIndex(d => d.value === value)
     const initialIndex = resolvedOptions.findIndex(d => d.value === initalValue)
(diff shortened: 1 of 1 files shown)


joepuzzo:
Try out https://joepuzzo.github.io/informed. I rewrote this library and will now be supporting `informed` instead.

Issue #268, opened 2018-03-05
Form validation errors don't always clear out


triketora:
I have a `validate` function that gets called when the submit button is clicked. It returns a object with error messages as per the spec. In my form, I check `formApi.errors` and render the error messages if they exist. For some reason, `formApi.errors` doesn't always update after the `validate` function runs. If I do a `console.log` inside the `validate` function, I can see it running and that the `errors` object is updating, but this doesn't get reflected in the form itself. In the screencap below, at the very end I've changed the field so everything validates and there should be no errors, but the error message still shows. Thoughts?

![formapi errors](https://user-images.githubusercontent.com/757690/36995563-6c473a6c-2082-11e8-8d27-86f846b393b0.gif)

```javascript
<Form
  defaultValues={this.state.data}
  validate={this.validate}
  validateOnSubmit={true}
  onSubmit={this.handleSubmit}
  getApi={this.setFormApi}
>
  {formApi =>
    <form
      onSubmit={e => {
        e.preventDefault();
        console.log('submitting');
        return formApi.submitForm(e);
      }}
    >
      <div>
        formApi.submits: {JSON.stringify(formApi.submits)}<br />
        formApi.errors {JSON.stringify(formApi.errors)}
      </div>
      <div className="preferred-name-wrapper">
        <label htmlFor="preferred-name">Preferred name</label>
        <br />
        <Text
          field="preferredName"
          id="preferredName"
          className={`${formApi.errors && formApi.errors.preferredNa
            ? 'has-error'
            : ''}`}
          placeholder="e.g., Goldilocks"
        />
        <p className="helper-text error preferred-name-error">
          {formApi.errors && formApi.errors.preferredName}
        </p>
      </div>
      <div className="email-wrapper">
        <label htmlFor="email">Email</label>
        <br />
        <Text
          field="email"
          id="email"
          className={`${formApi.errors && formApi.errors.email
            ? 'has-error'
            : ''}`}
        />
        <p className="helper-text error email-error">
          {formApi.errors && formApi.errors.email}
        </p>
      </div>
      <div className="password-wrapper">
        <label htmlFor="password">Password</label>
        <br />
        <Text
          field="password"
          id="password"
          type="password"
          className={`${formApi.errors && formApi.errors.password
            ? 'has-error'
            : ''}`}
        />
      </div>
      <div className="password-helper-text-wrapper">
        {(formApi.errors &&
          formApi.errors.password &&
          <p className="helper-text error password-error">
[...]

PR #309, opened 2018-04-18
Field: validate on mount for dynamic forms

Hyperkid123:
Recently, we started moving our project to React and we want to use this library to build our forms. We have a lot of dynamic forms and most of the inputs have "validate on mount" property, which is basically the same as  `validateOnMount` on the Form component. 

I added the `validateOnMount` property to generic `Field` component. 

I have also encountered a situation, when errors will remain in state, after some `Field` is removed from DOM. I've added simple action, that will remove the error from form state (similar to #268).

It would be nice if we could get this in :wink:


Hyperkid123:
@tannerlinsley @joepuzzo what do you think about this?

PR #315, opened 2018-04-26
Check the input values on componentDidMount

### LINKED ISSUE #314: Form values are lost when using SSR and hydrate
This is a timing issue, the timeline looks a bit like this:

1. server renders a form using `ReactDOMServer.renderToString`
2. browser autofills the form
3. *at some later point*, react on the client kicks in via `ReactDOM.hydrate`
4. `onSubmit` will return empty form values, although the input fields do have values.

I have a PR coming up shortly with a proposed fix.

Swatinem:
This way, react-form will correctly get form changes that happened
*before* react is initialized via `ReactDOM.hydrate`

Fixes #314 

I would primarily like to get some feedback on my approach:
* I created a test that manually uses jsdom, because enzyme does not yet support tests of reacts `hydrate` api. That makes things a little less nice.
* If you like the `ref`/`componentDidMoint` approach, I can also extract that into a different file and also add it to the other available components. So far, it was just a PoC.

### DIFF
--- a/src/components/form-fields/Text.js
+++ b/src/components/form-fields/Text.js
@@ -5,12 +5,32 @@ import React, { Component } from 'react'
 import withField from '../withField'
 
 class TextWrapper extends Component {
+  setElem = elem => {
+    const { ref } = this.props
+    this.elem = elem
+    if (ref) {
+      ref(elem)
+    }
+  }
+  elem = null
+
+  componentDidMount () {
+    const { value, fieldApi: { setTouched, setValue } } = this.props
+    const expected = !value && value !== 0 ? '' : String(value)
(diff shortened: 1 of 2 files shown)


joepuzzo:
Im sorry i guess i have never had this issue before so i dont understand. Can you elaborate on the issue? What are you seeing occur?

Swatinem:
The main problem is described in #314:

> 1. server renders a form using `ReactDOMServer.renderToString`
> 2. browser autofills the form
> 3. *at some later point*, react on the client kicks in via `ReactDOM.hydrate`
> 4. `onSubmit` will return empty form values, although the input fields do have values.

PR #319, opened 2018-05-01
Add dirty/pristine state to form fields

bringel:
I was looking for a way to show/enable a save button only if the user had actually modified a field instead of tabbing to the field. Most of the changes are the same as the code in `setTouched` to make sure all the appropriate validation happens


joepuzzo:
Try out https://joepuzzo.github.io/informed. I rewrote this library and will now be supporting `informed` instead. It has dirty and pristine at the form level

Issue #322, opened 2018-05-08
Trigger validation manually


laurenz-glueck:
It would be really helpful if there would be a method on the formApi to trigger the form validation manually. 

I have seen there is a "validateAll" function in the Form.js - why not making this available through the formApi? :)

joepuzzo:
Try out https://joepuzzo.github.io/informed. I rewrote this library and will now be supporting `informed` instead. And if you check out section on validation control you can do what you want!!!

laurenz-glueck:
@joepuzzo thanks for your response - just to be clear, does that mean that react-from is not longer maintained?

joepuzzo:
So I personally will no longer maintain it. But i think the plan is to depricate it. React form became a little to bloated so i decided to rewrite it for the 90% use case. Rather than try to pack every feature under the sun in it haha

PR #323, opened 2018-05-08
added formApi method to trigger form validation

laurenz-glueck:
according to my issue #322 this is a possibile implementation of a manually triggerable validation


joepuzzo:
Try out https://joepuzzo.github.io/informed. I rewrote this library and will now be supporting `informed` instead.

PR #345, opened 2019-10-05
Add TypeScript types

BrianMitchL:
This adds very rough initial typings for react-form. I based it off of the API docs and looking through some of the source. There are many uses of `any` throughout the types, particularly around getting and setting values and meta. I've never created a types for a library, so I'm not sure what is desired for tooling around testing, bundling, or linting the types. Some review would be much appreciated 🙂

Here's what I think needs to be worked on still

- [x] Refactor out as many uses of `any` as we can
- ~[ ] Find a type pattern for string object accessing, or can we change that to be functions to access nested data, a-la `mapStateToProps` in redux in order to be able to statically type check it?~ (I discussed this with Tanner, and don't this is going to be able to happen)
- [ ] Bundle types in dist
- [ ] Add typescript eslant plugins to lint the declarations
- [ ] Some testing or validation of the types?
- [ ] Docs and examples around the top-level functions (`useForm`, `useField`, etc.)

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -10,7 +10,7 @@ declare module 'react-form' {
     SetStateAction,
   } from 'react'
 
-  type Debounce = <T>(fn: (...args: any[]) => T, wait: number) => Promise<T>
+  type Debounce = <T>(fn: () => T, wait: number) => Promise<T>
   type ValidatorReturn = string | false | undefined
   type OptionalPromise<T> = Promise<T> | T
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -107,7 +107,7 @@ declare module 'react-form' {
   >(): FormInstance<TValues, CustomFormMeta>
 
   interface FieldMeta {
-    error: string | false
+    error: ValidationError
     isTouched: boolean
     isValidating?: boolean
   }
@@ -155,7 +155,7 @@ declare module 'react-form' {
 
   interface FieldOptions<TValue, CustomFieldMeta, TFormValues, CustomFormMeta> {
     defaultValue?: TValue
-    defaultError?: string
+    defaultError?: ValidationError
     defaultIsTouched?: boolean
     defaultMeta?: FieldMeta & CustomFieldMeta
     validate?: (
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -95,7 +95,7 @@ declare module 'react-form' {
     | Partial<FieldMeta>
     | ((previousMeta: FieldMeta) => FieldMeta)
 
-  interface FieldInstance<T = any, F = any> {
+  interface FieldInstance<T = any, F = {}> {
     form: FormInstance<F>
     fieldName: string
     value: T
@@ -130,9 +130,9 @@ declare module 'react-form' {
     defaultMeta?: FieldMeta
     validate?: (
       value: T,
-      instance: FieldInstance
+      instance: FieldInstance<T>
     ) => OptionalPromise<ValidatorReturn>
-    filterValue?: <T = any>(values: T, instance: FieldInstance) => T
+    filterValue?: (value: T, instance: FieldInstance<T>) => T
     validatePristine?: boolean
   }
 
@@ -141,8 +141,7 @@ declare module 'react-form' {
     options?: FieldOptions<T>
   ): FieldInstance
 
-  interface FieldOptionProps<T = any, Form = any>
-    extends FieldOptions<T> {
+  interface FieldOptionProps<T = any, Form = any> extends FieldOptions<T> {
     onSubmit?: (value: T, instance: FormInstance<Form>) => OptionalPromise<void>
     defaultValues?: Form
     debugForm?: boolean
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -10,12 +10,23 @@ declare module 'react-form' {
     SetStateAction,
   } from 'react'
 
+  type SerializableObject = { [key: string]: Serializable }
+  interface SerializableArray extends Array<Serializable> {}
+  type Serializable =
+    | string
+    | number
+    | boolean
+    | null
+    | SerializableObject
+    | SerializableArray
+
   type Debounce = <T>(fn: () => T, wait: number) => Promise<T>
   type ValidatorReturn = string | false | undefined
   type OptionalPromise<T> = Promise<T> | T
+  type ValidationError = string | false | null | undefined
 
   interface FormMeta {
-    error: string | any
+    error: ValidationError
     isSubmitting: boolean
     isDirty: boolean
     isSubmitted: boolean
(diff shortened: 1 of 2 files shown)


BrianMitchL:
@SleeplessByte Ping per your request on Twitter 🙂

sw-yx:
these look really good! 🛳️

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+    values: T
+    meta: FormMeta
+    formContext: FormInstance<T>
+    reset: () => void
+    setMeta: (value: SetFormMeta) => void
+    handleSubmit: FormEventHandler<HTMLFormElement>
+    debounce: Debounce
+    setValues: Dispatch<SetStateAction<T>>
+    runValidation: () => void
+    getFieldValue: <V = any>(fieldPath: keyof T | string) => V
```
SleeplessByte: Whilst this allows the _path of least resistance_, I would suggest considering:

```suggestion
    getFieldValue: <V = unknown>(fieldPath: keyof T) => V
```

This forces the _consumer_ to do dynamic type checks on the field value. I also suggest _not_ allowing a _random string_, because that doesn't work with type broadning. `'a' | 'b' | string` will always allow _any_ string, even if `T` is set correctly.

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+    getFieldMeta: (
+      fieldPath: keyof T | string
+    ) => { error: string | null; [k: string]: any }
```
SleeplessByte:
```suggestion
    ) => FieldMeta
```

I suggest extracting this to have parity between `FormMeta` and `FieldMeta`. Also makes it easier to extend if more _react-form_ fields are added down the line.

**Edit**: There is a `FieldMeta`. Should this not be `FieldMeta`?

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+    getInputProps: (
+      value: Partial<HTMLProps<HTMLInputElement>> &
+        Pick<HTMLProps<HTMLInputElement>, 'onSubmit'> &
+        Pick<HTMLProps<HTMLInputElement>, 'onBlur'>
+    ) => {
+      value: T
+      onChange: ChangeEventHandler<HTMLInputElement>
+      onBlur: FocusEventHandler<HTMLInputElement>
+    } & Partial<HTMLProps<HTMLInputElement>>
+    setValue: <V = any>(
```
SleeplessByte: Can a value _really_ be `any`? Or can it only be something that is _serializable_. The problem with `any` is that is also dictates `updater`, which now becomes:

```typescript
declare AnyValueSetter = SetValue<any> = any | ...
// AnyValueSetter  = any
```

Meaning it absolutely does nothing.

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+      updater: SetValue<V>,
+      options?: { isTouched: boolean }
+    ) => void
+    setMeta: (value: SetFieldMeta) => void
+    pushValue: <V = any>(value: V) => void
```
SleeplessByte: Same as above (re: `value` and `any`)

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+    insertValue: <V = any>(insertIndex: number, value: V) => void
```
SleeplessByte: Same as above (re: `value` and `any`)

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+    isTouched: boolean
+    [k: string]: any
+  }
+
+  type SetValue<S> = S | ((prevState: S) => S)
+  type SetFieldMeta =
+    | Partial<FieldMeta>
+    | ((previousMeta: FieldMeta) => FieldMeta)
+
+  interface FieldInstance<T = any, F = any> {
```
SleeplessByte: - `T`, see note below about `value` types.
- `F` is set to `any` by default here, but it's `T extends {} = {}` in `useForm`.

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  interface FieldOptions<T> {
+    defaultValue?: T
+    defaultError?: string
+    defaultIsTouched?: boolean
+    defaultMeta?: FieldMeta
+    validate?: (
+      value: T,
+      instance: FieldInstance
```
SleeplessByte: Type is not passed to `FieldInstance<T = any, F = any>`. This probably can be forwarded? `FieldInstance<T>`.

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+    ) => OptionalPromise<ValidatorReturn>
+    filterValue?: <T = any>(values: T, instance: FieldInstance) => T
```
SleeplessByte: - Why is there another `T` here? T comes in via `FieldOptions`. If its the same, remove the `<T = any>` here, otherwise rename.
- Type is not passed to `FieldInstance<T = any, F = any>`. This probably can be forwarded? `FieldInstance<T>`.

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+    validatePristine?: boolean
+  }
+
+  export function useField<T = any>(
```
SleeplessByte: The default for the _entire form values_ is `object` which would make this `never`. I suggest using `unknown` instead? If you want to use something broad, perhaps it's only _serializable types_?

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+    fieldPath: string,
+    options?: FieldOptions<T>
+  ): FieldInstance
+
+  interface FieldOptionProps<T = any, Form = any>
```
SleeplessByte: Assuming `Form` is the _entire shape_ of _all values_, it defaults to `{}` in `useForm`. Should probably match that.

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+    ComponentType,
+    Dispatch,
+    FocusEventHandler,
+    FormEventHandler,
+    HTMLProps,
+    Provider,
+    SetStateAction,
+  } from 'react'
+
+  type Debounce = <T>(fn: (...args: any[]) => T, wait: number) => Promise<T>
```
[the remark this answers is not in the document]
BrianMitchL: Oh yeah, good call, we definitely want to have the types reflect that.

src/index.d.ts
```diff
@@ -0,0 +1,158 @@
...
+  type ValidatorReturn = string | false | undefined
+  type OptionalPromise<T> = Promise<T> | T
+
+  interface FormMeta {
```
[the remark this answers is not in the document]
BrianMitchL: Oh yeah, I was trying to think about a good way to allow this to be typed. I've never used `Record`, cool! In your first snippet, trying to extend `C` emits an error, I'm guessing because it's being declared after being extended. What do you think of keeping `FormMeta` an interface with no generics, and instead using `FormMeta & C` whenever it's referenced?

```ts
interface FormMeta {
  error: string | any
  isSubmitting: boolean
  isDirty: boolean
  isSubmitted: boolean
  submissionAttempts: number
  isValid: boolean
  fieldsAreValidating: boolean
  fieldsAreValid: boolean
  canSubmit: boolean
}

type SetFormMeta<C = Record<string, any>> = Partial<FormMeta & C> | ((previousMeta: FormMeta & C) => FormMeta & C)

interface FormInstance<T, C = Record<string, any>> {
  Form: ComponentType<Omit<HTMLProps<HTMLFormElement>, 'onSubmit'>>
  values: T
  meta: FormMeta & C
  ...
}
```

src/index.d.ts
BrianMitchL: If we did this, we'd be saying that you can't access nested values when giving the function a string path to access a sub-value (in the case that the forms values are a series of nested objects and/or arrays). This kind of pattern ( implemented here: https://github.com/tannerlinsley/react-form/blob/master/src/utils.js#L35) is fundamentally incompatible with TypeScript, so that's why I left the `string` in there. But, that also makes the types for this function not really useful.

So, stricter type checking and skip that feature, or looser but support it? I almost lean towards the former. @tannerlinsley, do you have any thoughts?

src/index.d.ts
BrianMitchL: You're totally right that type exists already. I believe that it's not guaranteed to have `isTouched` on the `FieldMeta` object, specifically when calling `getFieldMeta` though. The function returns the current meta, but that is initialized to an empty object https://github.com/tannerlinsley/react-form/blob/master/src/hooks/useForm.js#L256
https://github.com/tannerlinsley/react-form/blob/master/src/hooks/useForm.js#L17

Maybe `FieldMeta` should mark it as optional, just in case? Then we can reuse the interface.

src/index.d.ts
BrianMitchL: The docs say any, but serializable definitely makes sense to me. Would that type be something like the following? Or is there a built-in?

```ts
type BaseSerializable = string | number | boolean
interface Seriealizable {
  [k: string]: BaseSerializable | BaseSerializable[] | Seriealizable
}
```

src/index.d.ts
BrianMitchL: Good catch, thanks!

src/index.d.ts
BrianMitchL: I was going with looser types if a generic isn't specified, but I agree, Serializable would for sure be better. If you have strong opinions about `unknown`, I would be open to that too. I usually see looser types vs `unknown` in the libraries I use though, maybe that's because it's relatively new, but it's fine to use now?

src/index.d.ts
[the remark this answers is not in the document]
SleeplessByte: Yeah, I absolutely would prefer separating out the "react-form"-meta and "custom meta".

src/index.d.ts
SleeplessByte: I'm personally *not* a fan of the path.access.thing, because it's so easy to do wrongly, and you won't get type safety anywhere. There is no way to do it. 

That said, if the looser variant is chosen, I still recommend `unknown`.

src/index.d.ts
SleeplessByte: I don't know of a built-in type. I think `type-fest` also has a `JSONSerializableThingType` so that makes me think it doesn't exist yet.

src/index.d.ts
SleeplessByte: Right! I use `unknown` to force the consumer of the library to do dynamic run-time type checks! With `any` things can just _magically break_.

tannerlinsley:
You guys are teaching me a lot about types as I watch this from the sidelines. Thanks :)

src/index.d.ts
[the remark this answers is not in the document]
BrianMitchL: I committed a start to this, but then stopped before committing the rest. I ultimately came up with this: https://gist.github.com/BrianMitchL/8e121cfb1633bb16924c81b6c264d88a

My example app that uses these types started breaking, and I'm having a hard time figuring out where. Using four generics everywhere, especially when we're splitting props out and creating field components, makes things really complex. Is there a better way than how I did it in the gist? I did move all of the default values for the generics to the functions. All of the internal interfaces require the generics.

src/index.d.ts
BrianMitchL: I added Serializable types to the gist I linked above, use the value type for value setters, and inferred types for the insert and push value functions if the value is an array (things like that make me looove TypeScript 😄).

BrianMitchL:
I just committed a bunch of changes, most of which adds generics all the way through, but it's getting pretty messy. Is this a good approach?

I'm also working on adapting Tanner's examples to use TypeScript so we have some implementations to test with, I'll comment back with some links once I have something up and running.

src/index.d.ts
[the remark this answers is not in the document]
BrianMitchL: I pushed all of the changes in the gist, so looking here at the PR is a better place to look at it 🙂

BrianMitchL:
Here's the Basic Form (Hooks Only) example: https://codesandbox.io/s/react-form-basic-form-hooks-only-typescript-z51fn?fontsize=14&hidenavigation=1&theme=dark It seems to work, but I think we need to make sure all of the generics default to `{}` so consumers don't need to explicitly set generics all the way through unless they are using custom meta. I'll start working on another example and see where that takes me 🙂

BrianMitchL:
I'm having problems with the InputField example. `getInputProps` from the `useField` hook forwards the entire form options object, which breaks attribute types for an `HTMLInputElement`. Here's what I have that illustrates this: https://codesandbox.io/s/react-form-basic-form-with-inputfield-wlh68?fontsize=14&hidenavigation=1&theme=dark Additionally, creating property typed generic components with forwardRef seems quite tricky (see https://stackoverflow.com/questions/58469229/react-with-typescript-generics-while-using-react-forwardref for some more details). This means the value of an input cannot be set per use of each InputField, so type checks fail because the input is set to accept any serializable type, but, for example, validation functions only check for a string. I'm not sure what the best approach is here.

SleeplessByte:
Does the SO solution for forwardRef not work?

waynevanson:
Looking forward to seeing these added.

What is currently being worked on with this? I'd be happy to look over anything if it helps.

BrianMitchL:
I got stuck with creating generic helper form inputs using forwardRef that use form hooks internally.

I've been quite busy and won't have time to keep working on porting examples over for another few weeks. If you'd like to give that a swing, that'd be super helpful 🙂

Alternatively, if you want to copy these types into a project and see how they work and let us know if you have any feedback, that'd also be helpful.

Lukas-Kullmann:
Today I tried using these typescript definitions and came across an issue with the typings of `getInputProps`. Currently, the return type is bound to `HTMLProps<HTMLInputElement>`. But I would suggest using generics. The implementation does not require `HTMLProps<HTMLInputElement>`. With this restriction, it is not possible to use this togother with material-ui's TextInput.

I would suggest this change:

```diff
-     getInputProps: (
-      value?: Partial<HTMLProps<HTMLInputElement>> &
-        Pick<HTMLProps<HTMLInputElement>, 'onSubmit'> &
-        Pick<HTMLProps<HTMLInputElement>, 'onBlur'>
-    ) => {
-      value: TValue
-      onChange: ChangeEventHandler<HTMLInputElement>
-      onBlur: FocusEventHandler<HTMLInputElement>
-    } & Partial<HTMLProps<HTMLInputElement>>
+    getInputProps: <
+      T extends { onChange?: ChangeEventHandler<HTMLInputElement>; onBlur?: FocusEventHandler<HTMLInputElement> }
+    >(
+      value?: T
+    ) => {
+      value: TValue
+      onChange: ChangeEventHandler<HTMLInputElement>
+      onBlur: FocusEventHandler<HTMLInputElement>
+    } & Omit<T, 'onChange' | 'onBlur'>
```

src/index.d.ts
```diff
@@ -0,0 +1,238 @@
...
+    | ((previousMeta: FormMeta & C) => FormMeta & C)
+
+  export interface FormInstance<TValues, CustomFormMeta> {
+    Form: ComponentType<Omit<HTMLProps<HTMLFormElement>, 'onSubmit'>>
+    values: TValues
+    meta: FormMeta & CustomFormMeta
+    formContext: FormInstance<TValues, CustomFormMeta>
+    reset: () => void
+    setMeta: (value: SetFormMeta<CustomFormMeta>) => void
+    handleSubmit: FormEventHandler<HTMLFormElement>
```
Vinlock, line 55: I seem to have been getting type issues until I modified this to:
```suggestion
    handleSubmit: FormEventHandler<HTMLElement>
```

meman:
Any news on this? I would love to start using them

Toliak:
There are `@types/react-form` package, by the way. You can use it, I believe the types are up to date :)

PR #352, opened 2019-11-03
Changing 'touched' state to reflect the common notion of 'touched'

cpepin:
@tannerlinsley want to get your thoughts on this. Generally, `touched` in other frameworks, libraries, etc. represents a field that has been focused/blurred. The `touched` state in `react-form` seems to represent `dirty`, but without the `touched` state, it's difficult to have a UX that needs to show validation on blur. Currently, with `touched` representing `dirty`, validation messaging will immediately populate as soon as a user types one character. For things like email fields this isn't great, as the user likely hasn't finished typing. I understand there is a concept of `debounce`, but I feel like that is solving something different.

### DIFF
--- a/src/hooks/useForm.js
+++ b/src/hooks/useForm.js
@@ -279,6 +279,11 @@ export default function useForm({
             ? updater(old.__fieldMeta[fieldID])
             : { ...old.__fieldMeta[fieldID], ...updater }
 
+        // If field has been blured, update form touched status
+        if (newFieldMeta.isTouched) {
+          apiRef.current.setMeta({ isTouched: true })
+        }
+
         return {
           ...old,
           // Any errors in fields should visually stop
@@ -297,11 +302,11 @@ export default function useForm({
         }
       })
     },
-    [setState]
+    [setState, apiRef]
   )
 
   const setFieldValue = React.useCallback(
-    (field, updater, { isTouched = true } = {}) => {
+    (field, updater, { isTouched = false } = {}) => {
       const fieldInstances = apiRef.current.__getFieldInstances(field)
 
       setState(old => {


tannerlinsley:
Agreed. This is good. What if the user wants to actually validate as you type tho?

cpepin:
Thinking we’ll need to add a “dirty” state. I can add it to this PR, just wanted to make sure you agreed first 👍

tannerlinsley:
Sounds good to me. Try to keep it all non breaking?

Tanner Linsley

cpepin:
Btw, really dig the approach you took with all of this.

cpepin:
For sure 👍

tannerlinsley:
Aww thanks man!

Tanner Linsley

tmilewski:
I just wanted to bump this as it would be incredibly helpful to our team. Thanks all!

cpepin:
@tmilewski I definitely don't mind pitching in, the more I think about it though - I'm not sure we can do this without a breaking change. It fundamentally changes how `touched` is working.

tmilewski:
@cpepin No worries, I appreciate all the work! I agree. I don't think that there is a way this pans out without a breaking change.

That said, I feel that as long as there's a different way to identify `touched` it as people have come to expect, doing that and bumping a major version number should be okay.

timcosta:
Tossing in another opinion/bump here - It seems appropriate to bump the major version with this change. I've also struggled with this usage of `touched` vs `dirty` where my forms are showing errors earlier than i want them to, but i have no way to defer the showing of errors until a blur occurs

cpepin:
@tannerlinsley what are your thoughts on a major release for this? I don’t mind contributing the code but I’ve been avoiding jumping on it, as I didn’t want to introduce any breaking changes.

GuySerfaty:
anything new about this topic? do you have any workaround for this one?

hamonCordova:
It's the biggest problem of the lib, dosn't has `dirty`, because the error is showing earlier... so make the libary useless

crutchcorn:
Closing this PR for now, as we're working on a substantial rework of the codebase that's expected to undergo more rapid changes as I work on stabilizing things.

Rest assured that we'll have `isDirty` and `isTouched` as two distinct items :) 

If you'd like to get involved in the rework, [join our Discord](https://discord.gg/tanstack-719702312431386674) and reach out in the #form channel

PR #354, opened 2019-11-11
feat: do not transpile async

### LINKED ISSUE #353: Do not transpile async functions
Thanks for providing `react-form@4` :heart:.

I'm working with front-end requirements that would allow me to use async functions out of the box. Unfortunately I still need to bundle the regenerator runtime in order to use `react-form@4.0.0...`.

This leads me to the question if you could drop transpilation of async functions. 

At this point, all browsers except for IE 11 (dead) support async functions, cf. [caniuse](https://caniuse.com/#search=async). 

Using `babel@7` people that still need to support IE 11 or other (older) browsers could transpile `async` via overrides.

nikku:
Closes #353

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,7 +3,11 @@
     [
       "@babel/env",
       {
-        "modules": false
+        "modules": false,
+        "exclude": [
+          "@babel/plugin-transform-async-to-generator",
+          "@babel/plugin-transform-regenerator"
+        ]
       }
     ],
     "@babel/react"


tannerlinsley:
Released as 4.0.0. In fact, that's the first stable release of 4.0! Better now than never, right!?

nikku:
Good to have it. Do you mind pushing a `v4.0.0` tag to GitHub, too?

tannerlinsley:
Pushed!

Tanner Linsley

PR #389, opened 2021-12-29
Add React 17 support

tm1000:
This library (though officially not maintained) works fine in React 17

The reasoning for this is because NPM v7 installs peerDependencies by default. The only way to get around the conflict when using this library and React 17 is to append `--legacy-peer-deps`

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "peerDependencies": {
     "prop-types": "^15.5.4",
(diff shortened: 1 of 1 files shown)


tm1000:
@tannerlinsley would you be able to publish this? Or is that risky (Last publish to npm was pre-covid Nov 2019!)

PR #397, opened 2023-05-30
Add github starts for other libraries in comparison page

bhattrajat:
image


crutchcorn:
Thanks for the contribution! ✨

Any way you might be game to add in the Bundle Size as well? :)

PR #402, opened 2023-07-04
Fix user input

crutchcorn:
> This PR builds on top of this PR: #401, please merge it first

This PR (specifically, this commit: https://github.com/TanStack/form/commit/2f7ec1d87e85d55498581e9ee716c302727c3b00) aims to fix user input from it's original broken state.

It includes tests to validate the fix.

However, I'm not a huge fan of this fix, as it duplicates the method of setting a value between the Form and Field. I'd much rather either:

- Add a listener to the form from the field that updates the field when the form is updated
- Abstract that away into a computed item in `@tanstack/store` that allows us to do this consistently with derived state


tannerlinsley:
Happy to merge whatever you see fit. Just go for it Corbin!

PR #420, opened 2023-08-30
fix: cannot update a component/bad setState

GabrielModog:
What this PR aims to:
- fixes warning: "Cannot update a component while rendering a different component warning"

image

### DIFF
--- a/packages/react-form/src/useForm.tsx
+++ b/packages/react-form/src/useForm.tsx
@@ -78,7 +78,7 @@ export function useForm<TData>(opts?: FormOptions<TData>): FormApi<TData> {
 
   useEffect(() => {
     formApi.update(opts)
-  }, [])
+  }, [opts])
 
   return formApi as any
 }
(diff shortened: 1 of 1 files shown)


packages/react-form/src/useForm.tsx
```diff
@@ -75,7 +75,10 @@ export function useForm<TData>(opts?: FormOptions<TData>): FormApi<TData> {
   })
 
   formApi.useStore((state) => state.isSubmitting)
-  formApi.update(opts)
+
+  useEffect(() => {
+    formApi.update(opts)
+  }, [])
```
crutchcorn, line 81: We probably want to have `opts` in the dep array here

crutchcorn (changes requested):
Minor change, but this looks good otherwise :)

packages/react-form/src/useForm.tsx
GabrielModog: I saw the wouldn't be needed because this method is called one time. This way it's called when the app is mounted.

If `opts` be a dep will be called always the forms submitted. This `update` method is only for `defaultValues`.

crutchcorn (approved):
This is great stuff, thank you for the contribution!

PR #424, opened 2023-08-31
fix(form-core): allow set default field meta

joaom00:
While writing tests for onChange validation in the react adapter, I noticed that the default meta wasn't being passed to the field. This PR aims to fix that

### DIFF
--- a/packages/form-core/src/tests/FieldApi.spec.ts
+++ b/packages/form-core/src/tests/FieldApi.spec.ts
@@ -19,21 +19,31 @@ describe('field api', () => {
     expect(field.getValue()).toBe('test')
   })
 
-  it('should set a value correctly', () => {
-    const form = new FormApi({
-      defaultValues: {
-        name: 'test',
-      },
-    })
-
+  it('should get default meta', () => {
+    const form = new FormApi()
     const field = new FieldApi({
       form,
       name: 'name',
     })
 
-    field.setValue('other')
+    expect(field.getMeta()).toEqual({
+      isTouched: false,
+      isValidating: false,
+    })
+  })
 
-    expect(field.getValue()).toBe('other')
+  it('should allow to set default meta', () => {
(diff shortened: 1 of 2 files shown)


packages/form-core/src/tests/FieldApi.spec.ts
```diff
@@ -19,21 +19,31 @@ describe('field api', () => {
     expect(field.getValue()).toBe('test')
   })
 
-  it('should set a value correctly', () => {
```
crutchcorn, line 22: I think we may have accidentally deleted this test. Can we make sure to add it back? :)

crutchcorn (changes requested):
Amazing fix as usual @joaom00 - thank you so so so much for joining us over from HouseForm!!

Just one small thing before we merge

packages/form-core/src/tests/FieldApi.spec.ts
joaom00: This is actually a [duplicate test](https://github.com/TanStack/form/blob/main/packages/form-core/src/tests/FieldApi.spec.ts#L22-L54)

joaom00: This is actually a [duplicate test](https://github.com/TanStack/form/blob/main/packages/form-core/src/tests/FieldApi.spec.ts#L22-L54)

crutchcorn: Oh! You're 100% right. Must've been my bad haha

crutchcorn (approved):

crutchcorn:
Dunno what's going on with `Lint` or `Typecheck` but it's clearly unrelated to the code. We'll merge and fix later if need-be

crutchcorn:
There are tests failing, but they're not related to your work. Merging

PR #432, opened 2023-09-02
[Form Core] Add tests to the validations API and remove `asyncDebounceMs`

joaom00:
This PR adds tests to the validations API (`onChangeAsync`, `onBlur`, `onBlurAsync`, `onSubmitAsync`) and with debounce as well. Also remove old `asyncDebounceMs` API

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -30,7 +30,6 @@ export interface FieldOptions<
   name: TName
   index?: TData extends any[] ? number : never
   defaultValue?: TData
-  asyncDebounceMs?: number
   asyncAlways?: boolean
   onMount?: (formApi: FieldApi<TData, TFormData>) => void
   onChange?: ValidateFn<TData, TFormData>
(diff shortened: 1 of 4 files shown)


packages/form-core/src/FieldApi.ts
```diff
@@ -322,19 +320,19 @@ export class FieldApi<TData, TFormData> {
...
         ? 0
         : (cause === 'change'
-            ? onChangeAsyncDebounceMs
-            : onBlurAsyncDebounceMs) ??
-          asyncDebounceMs ??
-          500
+          ? onChangeAsyncDebounceMs
+          : onBlurAsyncDebounceMs) ??
+        asyncDebounceMs ??
+        500
```
joaom00, line 335: Should we remove this? or is this right?

crutchcorn: Yeah, I'd say let's remove it

joaom00: Done

PR #438, opened 2023-09-03
Update quick-start.md to .Provider

spookyuser:
I think this is correct form.Form didn't work for me


crutchcorn:
Whoops! Thanks for catching this

PR #439, opened 2023-09-03
Make runtime agnostic/Support React Native

crutchcorn:
This PR aims to solve React Native support and make the code less runtime/UI reliant by removing all methods and types that utilized the DOM. This includes:

- Removing `APIForm.getFormProps()`
- Removing `APIField.getInputProps()`
- Removing `APIField.getChangeProps()`
- Changing `APIForm.handleSubmit()` to remove `e` argument
- Adding a new `APIField.handleChange()` to call on an input by: `onChange={e => field.handleChange(e.target.value)}`
- Adding a new `APIField.handleBlur()` to call on an input by: `onBlur={field.handleBlur}`

While this might seem like a more verbose syntax, it brings a few additional benifits:

- Avoids headaches with typings for UI libraries like React Aria or ShadCN/UI that might not use exact `input` or similar under-the-hood.
- Allows us more flexibility if UI libraries or runtimes have different props names
- Reduces our API surface area, which makes cognitive load easier in the end
- Detaches our implementation from React, which should be easier to adopt to other frameworks

### DIFF
--- a/docs/framework/react/quick-start.md
+++ b/docs/framework/react/quick-start.md
@@ -32,7 +32,12 @@ export default function App() {
           <form.Field
             name="fullName"
             children={(field) => (
-              <input name={field.name} {...field.getInputProps()} />
+              <input
+                name={field.name}
+                value={field.state.value}
+                onBlur={field.handleBlur}
+                onChange={(e) => field.handleChange(e.target.value)}
+              />
             )}
           />
         </div>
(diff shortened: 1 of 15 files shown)


tannerlinsley (approved):

joaom00:
Do you think it would be worth adding a `getInputProps` to the adapters for quick configuration? but keep handleChange, handleBlur to cover the cases listed in the topic.

crutchcorn:
@joaom00 I originally tried to implement this until I realized that we'd run into the same problem but for the React adapter. I'd rather not ship out different adapters for React / React Native.

I mean, we could theoretically do a `useForm` exported from `@tanstack/react-form/dom` but then there's still:

- Confusion about API differences
- More than one way to do a thing, making training harder

PR #442, opened 2023-09-05
feature(FieldAPI): Change from touched error message and  error message to error map and array of errors

### LINKED ISSUE #431: Change from single error message to object or array of errors
Right now, we only have a single error per field. This seems like it might limit the UX of any app using TanStack Form.

Instead, I'd like to change our validators to return an object or array of errors:

```ts
errorMap = {
    onChange: ValidationError,
    onMount: ValidationError,
    onBlur: ValidationError,
}
```

Or:

```ts
errors: [] // As many errors as show up in `onChange`, `onMount`, `onBlur`
```

If anything, we may want to provide both to the `FieldAPI` in our react adapter. Something like:

```tsx
<form.Field
  name="firstName"
  children={(field) => {
    const errArr = field.state.meta.errors // Array of errors
    const errMap = field.state.meta.errorMap; // Object of errors
    const onChangeErr = errMap?.onChange; // string | undefined | null
  }}
/>
```

Where `errors` is derived from `errorMap` as a derived value

CheRayLiu:
We'd like to change our validators to return an object and array of errors to avoid UX limitations for any app that uses TanStack Form. 

Quoted from @crutchcorn. Based on the changes, an example usage in our react adapter would be:
```
<form.Field
  name="firstName"
  children={(field) => {
    const errArr = field.state.meta.errors // Array of errors
    const errMap = field.state.meta.errorMap; // Object of errors
    const onChangeErr = errMap?.onChange; // string | undefined | null
  }}
/>
```

This change includes:
- Replacing `FieldMeta.error` with `FieldMeta.errors` and `FieldMeta.errorMap`
- Replacing `FieldMeta.touchedError` with `FieldMeta.touchedErrors` 
- Updating the validation methods to accommodate those changes 
- Update documentation to reflect the changes

Resolves #431


packages/form-core/src/FieldApi.ts
```diff
@@ -1,8 +1,13 @@
...
+  isNonEmptyArray,
+  type DeepKeys,
+  type DeepValue,
+  type Updater,
+} from './utils'
+import type { FormApi, ValidationError, ValidationErrorMap } from './FormApi'
 import { Store } from '@tanstack/store'
 
-export type ValidationCause = 'change' | 'blur' | 'submit'
+export type ValidationCause = 'change' | 'blur' | 'submit' | 'mount'
```
crutchcorn, line 10: Glad you made this change - I intended to refactor our mount validation to include this

packages/form-core/src/FormApi.ts
```diff
@@ -37,13 +37,19 @@ export type FieldInfo<TFormData> = {
...
-  validationResolve?: (error: ValidationError) => void
-  validationReject?: (error: unknown) => void
+  validationPromise?: Promise<ValidationError[]>
+  validationResolve?: (errors: ValidationError[]) => void
+  validationReject?: (errors: unknown) => void
 }
 
 export type ValidationError = undefined | false | null | string
 
+export type ValidationErrorMapKeys = 'onChange' | 'onSubmit' | 'onBlur'
```
crutchcorn, line 47: I'm in favor of changing this to:

```ts
export type ValidationErrorMapKeys = `on${Capitalize<ValidationCause>}`
```

But I could be convinced either way. What do you think?

packages/form-core/src/FieldApi.ts
```diff
@@ -410,3 +427,11 @@ function normalizeError(rawError?: ValidationError) {
 
   return undefined
 }
+
+function getErrorMapKey(cause: ValidationCause) {
+  return cause === 'submit'
```
crutchcorn, line 432: I'd rather this be a `switch`/`case` as it reads a lot nicer than nested turneries.

crutchcorn: Let's also add `onMount`

packages/form-core/src/tests/FieldApi.spec.ts
```diff
@@ -427,10 +457,47 @@ describe('field api', () => {
...
     field.validate('submit')
     await vi.runAllTimersAsync()
-    expect(field.getMeta().error).toBe('Please enter a different value')
+    expect(field.getMeta().errors).toContain('Please enter a different value')
+    expect(field.getMeta().errorMap).toMatchObject({
+      onSubmit: 'Please enter a different value',
+    })
+  })
+
+  it('should contain multiple errors when running validation onBlur and onChange', () => {
```
crutchcorn, line 470: Love this test!

crutchcorn (commented):
**Fantastic** start! There's some super minor nitpicks here, but you're on an amazing track!

packages/form-core/src/FieldApi.ts
CheRayLiu: Good call, switch/case sounds much nicer

packages/form-core/src/FormApi.ts
CheRayLiu: Oh interesting, I didn't about this in TS. Learned something new today! Thanks, I prefer your suggestion actually, will update!

packages/form-core/src/tests/FieldApi.spec.ts
CheRayLiu: Thanks!

PR #451, opened 2023-09-08
fix(vue): Component default scoped slot value types

wobsoriano:
Hi! This PR adds typings to the `Subscribe` and `Field` component default slot values.

Screenshot 2023-09-08 at 12 22 55 AM

Screenshot 2023-09-08 at 12 17 50 PM

Some reference: https://github.com/vuejs/core/pull/7982

Tests passed ✅

### DIFF
--- a/docs/framework/vue/quick-start.md
+++ b/docs/framework/vue/quick-start.md
@@ -28,7 +28,7 @@ form.provideFormContext()
   <div>
     <div>
       <form.Field name="fullName">
-        <template v-slot="field">
+        <template v-slot="{ field }">
           <input
             :name="field.name"
             :value="field.state.value"
--- a/examples/vue/simple/src/App.vue
+++ b/examples/vue/simple/src/App.vue
@@ -45,7 +45,7 @@ async function onChangeFirstName(value) {
         :onChangeAsyncDebounceMs="500"
         :onChangeAsync="onChangeFirstName"
       >
-        <template v-slot="field, state">
+        <template v-slot="{ field, state }">
           <label :htmlFor="field.name">First Name:</label>
           <input
             :name="field.name"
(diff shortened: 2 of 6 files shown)


wobsoriano:
We could probably add types to the `Field` component as well but the way values are currently passed to the default slot makes it difficult to type:

Screenshot 2023-09-08 at 12 34 25 AM

If you are interested I can update it to pass an object instead - in that way we can type it like this PR:

```ts
context.slots.default!({
  field: fieldApi.api,
  state: fieldApi.state.value
})
```
```ts
SetupContext<EmitsOptions, SlotsType<{ default: {
    field: FieldApi<FieldValue<TParentData, TField>, TFormData>,
    state: FieldState<any>
  } }>
```

crutchcorn:
Now's the best time to do breaking changes, let's do it! 😄 

Thanks for this change. 

BTW, have you seen the typings for `FieldComponent`? They're pretty gross and type out `children`:

https://github.com/TanStack/form/blob/f6bb80ef932cd196eff0af2cd9ee269234542a87/packages/vue-form/src/useField.tsx#L115-L134

I don't think this is right, but no matter what I tried to update, I couldn't get the TSX in our tests working properly and had to resort to manually typing `field`:

https://github.com/TanStack/form/blob/f6bb80ef932cd196eff0af2cd9ee269234542a87/packages/vue-form/src/tests/useForm.test.tsx#L72

This would be _**amazing**_ to get fixed, any chance you have an idea of how to do so?

wobsoriano:
Added an update for the `Field` component as well.

> BTW, have you seen the typings for FieldComponent

@crutchcorn Unfortunately no, I think it only supports SFC for now? Even tried to do a quick component with a default slot to check its type:

Screenshot 2023-09-08 at 11 35 22 AM

That's plain TS and it can't still pick up the slot value types.

crutchcorn:
> That's plain TS and it can't still pick up the slot value types.

That's what I ran into as well..

Do we wanna just remove the `children` typing then as part of this Vue adapter type cleanup?

wobsoriano:
@crutchcorn Done!

crutchcorn (approved):
Great work, thank you for your help and patience!

PR #462, opened 2023-09-09
Third party adapters

### LINKED ISSUE #428: [Feature Request] Support third-party validator libraries
It would be great to have some internal logic that allowed a `validatorAdapter` like so:

```ts
import { ValidationError } from "@tanstack/form-core";
import type { ZodTypeAny } from "zod";

function zodValidator() {
  return {
    validate<T>(value: T, fn: ZodAny): ValidationError {
      // Call Zod on the value here and return the error message
      return 0 as never
    },
    async validateAsync<T>(value: T, fn: ZodAny): Promise<ValidationError> {
      // Call Zod on the value here and return the error message
      return 0 as never
    },
  }
}
```

Such that we could handle `onChange` and other validation types a bit easier:

```tsx
import * as React from 'react'
import { useForm } from '@tanstack/form-react'
import * as z from 'zod'
import {zodValidator} from "@tanstack/zod-validator";

export const App = () => {
  const form = useForm({
    defaultValues: {
      firstName: 'FirstName',
    },
    onSubmit: () => {
      console.log('Submitted')
    },
  })

  return (
    <form.Provider>
      <form.Field
        name="firstName"
        validator={zodValidator}
        // onChange is passed through validator above to make the Zod error
        // available and handle TypeScript OOTB
        onChange={z.string().min(3, 'Too short')}
        children={(field) => (
          <div>
            <input
              data-testid="fieldinput"
              name={field.name}
              {...field.getInputProps()}
            />
            <p>{field.getMeta().error}</p>
          </div>
        )}
      />
    </form.Provider>
  )
}
```

# Goals

Despite using Zod as an example, we want this `adapter` system to work for any library. To do this, we'll start with a `testAdapter` that allows us to play with the API a bit.

The idea is that we'll want to use TypeScript in the `@tanstack/form-core` library that allows us to:

**1) Doesn't break typical `onChange` usage**

```tsx
<form.Field
  name="firstName"
  validator={zodValidator}
  onChange={(val, fieldApi) => {
    //       ^ Should still be typed from ValidateFn
  }}
  children={(field) => (
    /* ... */ null
  )}
/>
```

**2) Throw errors when `onChange` doesn't match the `validator` type**

```tsx
<form.Field
  name="firstName"
  validator={zodValidator}
  onChange={
      yup.string()
  // ^ Should error in TypeScript, as it's not Zod
  }
  children={(field) => (
    /* ... */ null
  )}
/>
```

**3) Throws when the wrongly typed validator is passed**

```tsx
<form.Field<string>
  name="firstName"
  validator={zodValidator}
  onChange={
      z.object({})
  // ^ Should error in TypeScript, as this is a string
  }
  children={(field) => (
    /* ... */ null
  )}
/>
```

We also want to make sure the PR that introduces this change has:

- [ ] Some integration tests to validate functionality
- [ ] Some type tests to validate TS typings

> This is a fairly complex ask, I am happy to mentor anyone through any of these steps - [just ask in our Discord and tag me](https://tlinz.com/discord) 😄

# Non-goals

While we want the adapter system to work for the following projects, we do not want to include these adapters in our v1 of this idea:

- [ ] [Zod](https://zod.dev/)
- [ ] [Valibot](https://github.com/fabian-hiller/valibot)
- [ ] [Yup](https://github.com/jquense/yup)

Instead, as mentioned in the `goals` section, we want to make a generic test adapter we're able to base others off of.

crutchcorn:
This PR implements an API akin to the following:

```jsx
<form.Field validator={zodValidator} onChange={z.string().min(1, "Test")}>
```

And makes it a reality.

Closes #428

# TODO

- [x] Split up `zod` adapter into dedicated package
     - [x] Write integration tests
     - [x] Write type tests
- [x] Write Yup adapter
- [x] Fix Vue adapter typings
- [x] Support async validation
    - [x] Write integration tests for Yup/Zod validators
    - [x] Write functionality itself
    - [x] Add types
- [x] Add support for Form-level third-party adapters
    - [x] Write integration test
    - [x] Write types
    - [x] Write functionality itself
- [ ] Investigate/Fix legacy DTS generation from TSUp causing Vue failures
- [x] Docs
- [x] Examples


crutchcorn:
I'm reasonably sure there's a bug in the `legacy` setup of TSUp that's renaming `this` to `this$1`, even for DTS generation:

image

crutchcorn:
The bug with TSUp is an underlying bug with `rollup-plugin-dts`: https://github.com/Swatinem/rollup-plugin-dts/issues/279

crutchcorn:
Temporarily fixed the `tsup`/`rollup-plugin-dts` issue using PNPM's `patch` functionality. Long-term tracking of fix here:

https://github.com/Swatinem/rollup-plugin-dts/pull/287

crutchcorn:
@wobsoriano I believe I broke Vue's `onChange` `value` inferencing in this PR. Any chance you'd want to help me review the changes and figure out why that's the case? I'm admittedly a little stumped.

wobsoriano:
@crutchcorn Doing something like `:onChange` makes it a prop instead of an event in a Vue SFC, hence losing the types. What we're looking for is `@change`. 

Changing the Vue example to `@change` fixes it:

Screenshot 2023-09-10 at 9 29 18 AM

SFC - `@eventName`
JS/JSX - `onEventName` (you can confirm this by looking at the test file)

wobsoriano:
@crutchcorn Doing something like `:onChange` makes it a prop instead of an event in a Vue SFC, hence losing the types. What we're looking for is `@change`. 

Changing the Vue example to `@change` fixes it:

Screenshot 2023-09-10 at 9 29 18 AM

SFC - `@eventName`
JS/JSX - `onEventName` (you can confirm this by looking at the test file)

PR #465, opened 2023-09-13
fix: remove _TData abstraction

tannerlinsley:


tannerlinsley:
I'm a big fan of this change for a few reasons:
- Less code
- Uniform generic slot names and order across the library
- Still approachable for constraints e.g.

PR #467, opened 2023-09-14
docs(quick-start): add <form> tag to quick start example

r-bakes:
Align the "quick-start" example and "simple" example. Fix misleading doc code example.


crutchcorn (approved):
Thanks for this!

PR #471, opened 2023-10-03
feat: solid.js form

aadito123:
The Solid.js port of the form/core

Made all the options parameters functions to match how they are done in TanStack/query v5, that way reactive parameters don't require getters.

For Field, I had to do some finagling to register changes in the Field.store into Solid's reactivity. It makes the reactivity far less fine-grained than it could be since any change in the store would cause any effect subscribing to any unrelated field would still run. I was thinking of using a Proxy but the Field.store has many methods on them and I am unsure how to work on those.


crutchcorn:
This is incredible! Thank you!

Are you up for a chat via Discord voice call by chance? Would love to walk through what potential next steps are and what edges of the code you'd like to look through! 😊

aadito123:
Sure! What's your username?

packages/solid-form/src/createField.tsx
```diff
@@ -0,0 +1,132 @@
...
+// ugly way to trick solid into triggering updates for changes on the fieldApi
+function makeFieldReactive<FieldApiT extends FieldApi<any, any>>(
+  fieldApi: FieldApiT,
+): () => FieldApiT {
+  const [flag, setFlag] = createSignal(false)
+  const fieldApiMemo = createMemo(() => [flag(), fieldApi] as const)
+  const unsubscribeStore = fieldApi.store.subscribe(() => setFlag((f) => !f))
+  onCleanup(unsubscribeStore)
+  return () => fieldApiMemo()[1]
+}
```
crutchcorn, line 39: Let's see if we can't replace this with our `@tanstack/store` implementation from @wobsoriano (speaking of, if you @wobsoriano could review this PR as another Solid pro, that would be rad):

https://github.com/TanStack/store/pull/25

packages/solid-form/src/createForm.tsx
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * formApi.update should not have any side effects. Think of it like a `useRef`
+   * that we need to keep updated every render with the most up-to-date information.
+   */
+  createComputed(
+    on(
+      () => [opts?.(), formApiStore.isSubmitting],
+      () => formApi.update(opts?.()),
+    ),
+  )
```
crutchcorn, line 64: Let's split this into two distinct items for code readability and maybe even remove the `isSubmitting` check

crutchcorn (changes requested):

crutchcorn:
I'm taking over this here PR 🏴‍☠️🦜 

Namely, I want to make sure we have:

- Updates to docs (even if they're just `todo`s for now)
- Examples
- Adding to our deployment strategy so the new package gets published

And I feel guilty with how long this PR has been in the oven (my fault 100%). Gonna try to take your awesome effort and push it past the finish line tonight

PR #475, opened 2023-10-18
fix: form's valid state should be work fine

### LINKED ISSUE #464: form.Subscribe is only re-rendered if input length > 1
### Describe the bug

As soon as the input validation fails the first time, the children of form.Subscribe are only re-rendered with the correct value of `canSubmit` if the input length is longer than 1 character.

https://codesandbox.io/p/sandbox/musing-microservice-gpzhn3?embed=1&file=/src/index.tsx:18,25

### Steps to reproduce

1. Go to code sandbox: https://codesandbox.io/p/sandbox/musing-microservice-gpzhn3?embed=1&file=/src/index.tsx:18,25
2. type something into the input field, then delete everything
3. type again something into the input field

### Expected behavior

The submit button should be activated as soon as the input field is validated correctly, however here it only happens after two characters were typed into the input field.

Every time

### Screenshots or Videos

https://github.com/TanStack/form/assets/6340397/76969c07-049e-4507-9fcd-baa7c002f03e

### Platform

- OS: macOS
- Browser (Firefox and Chrome)

### Tanstack Form adapter

react-form

### TanStack Form version

0.3.4

### TypeScript version

_No response_

### Additional context

_No response_

minosss:
Field just update meta `errorMap` when error change. The value of form is the previous value.

Close #464

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -134,8 +134,10 @@ export class FormApi<TFormData, ValidatorType> {
             (field) => field?.isValidating,
           )
 
-          const isFieldsValid = !fieldMetaValues.some((field) =>
-            isNonEmptyArray(field?.errors),
+          const isFieldsValid = !fieldMetaValues.some(
+            (field) =>
+              field?.errorMap &&
+              isNonEmptyArray(Object.values(field.errorMap).filter(Boolean)),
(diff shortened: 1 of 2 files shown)


crutchcorn (approved):
Great catch, thank you for implementing tests as well!

Issue #477, opened 2023-10-18
[Docs] Add more info to the guide page about validation


juliendelort:
Some topics that the page should cover:

- [Already partially covered] How do I validate fields and when (on change, on blur, on submit, on mount)? Especially addressing confusions like [this](https://github.com/TanStack/form/issues/469).
-  More details about showing validation errors: should I use `field.state.meta.touchedErrors` or `field.state.meta.errors`? Should I loop over them?
- Handling validation at the fields level (using each field onChange) vs at the form level (using the callbacks in `useForm`). Can I do a mix of both? Should I?
- How do I handle validation rules that include multiple fields (for example a “password” and a “repeat password” field that need to be identical, or a “start date” that needs to be before an “end date”).
- Show the possible ways to prevent submitting invalid forms.

crutchcorn:
This makes sense to me, especially with `Form.Subscribe` not being the most straightforward API design

juliendelort:
Working on this!

juliendelort:
@crutchcorn I'm trying to push my branch in order to open a PR but looks like I don't have permission:

> remote: Permission to TanStack/form.git denied to juliendelort.
fatal: unable to access 'https://github.com/TanStack/form.git/': The requested URL returned error: 403

crutchcorn:
Hey @juliendelort - you'll need to fork the repo:

https://docs.github.com/en/get-started/quickstart/fork-a-repo

Then make a pull request from your fork to this repo 😊

Let me know if there's anything I can do to help answer! Happy to walk through step-by-step

juliendelort:
Haha I'm such a noob 😅 
Thanks @crutchcorn I should be good now!

crutchcorn:
Closing, as @juliendelort has addressed most of these concerns in the now-merge #484.

This doesn't mean our validation docs can be added onto and improved, however! If anyone sees room to improve, feel free to make a PR or follow-up issue!

PR #484, opened 2023-10-21
docs(validation): Improved validation guide

juliendelort:
Issue: https://github.com/TanStack/form/issues/477

Add more info to the validation guide page.

While working on this, I've identified a few issues and inconsistencies. I'll probably create issues and discussion topics about these, but listing them here first in case I've missed some things:

## 1. formAPI's `onChange` not called
formAPI's `onChange` doesn't seem to be called at all and it expects the same return type as fieldAPI's `onChange`. If we want to allow doing validation at the form level in `onChange`, we probably need it to return a mapping of errors per field (same with onBlur and async methods).  [Here](https://codesandbox.io/p/sandbox/determined-mayer-yn4qfc?file=%2Fsrc%2Findex.tsx%3A43%2C29) is a repro of this.

## 2. Validate untouched forms
As I have mentioned in the docs, it seems that the way to prevent submission of untouched forms is to do some validation in the `onSubmitAsync` prop (at the `Field` level). However, once the form is in error due to an error raised in `onSubmitAsync`, it seems the form is stuck in error and the user can't submit, even if they've fixed the issue. [Here](https://codesandbox.io/p/sandbox/focused-lovelace-f2p8wz?file=%2Fsrc%2Findex.tsx%3A37%2C32) is a repro. Try submitting the form before making any change. Then it's impossible to submit again, even after providing correct values to the fields.

## 3. sync vs async validation
I don't know the motivation behind it but I wonder if we actually need to separate sync and async validation? For example, instead of having `onChange` and `onChangeAsync`, it could be simpler if we just had `onChange` that would accept a function that either returns a promise (async) or directly a validation error (sync)?

## 4. Per-event validation
This is a broader discussion topic but I'm seeing potential issues with separating out validation per event (onChange vs onBlur vs onMount etc...). I don't know if there is a use case where users would want to do different validations in `onChange` and `onBlur` and `onSubmitAsync` for example. From my experience, a field is either valid or invalid at a given time, regardless of *when* the validation is done, and users may often end up reusing the same validation rules in `onChange` and `onSubmitAsync` or in `onBlur` and `onSubmitAsync` for example.

Imo, it could make things simpler if we decoupled the validation rules (*what* makes a field valid) from *when* the validation should be done. 

Another challenge created by this approach is that errors are cleared by the same event that raised them. For example if `onBlur` returns a validation error, that error will only be cleared the next time the field is blurred. In some use cases, users could want to do the initial validation on blur but remove the errors as the user is typing to fix them. [Here](https://codesandbox.io/p/sandbox/determined-chaplygin-hlk7vz?file=%2Fsrc%2Findex.tsx%3A21%2C21) is an example:  (type 2 characters in the first name field then focus the other field: "First name must be at least 3 characters" is shown. It could be good to be able to remove the error as soon as the user types the third character).

I believe this is also what is causing point 2 above about validating untouched forms.

### DIFF
--- a/docs/guides/validation.md
+++ b/docs/guides/validation.md
@@ -16,29 +16,22 @@ It's up to you! The `<Field />` component accepts some callbacks as props such a
 Here is an example:
 
 ```tsx
- <form.Field
-  name="firstName"
-  onChange={(value) => {
-    return !value
-      ? "A first name is required"
-      : value.length < 3
-        ? "First name must be at least 3 characters"
-        : undefined;
-  }}
+<form.Field
+  name="age"
+  onChange={val => val < 13 ? "You must be 13 to make an account" : undefined}
 >
-  {(field) => {
-    return (
-      <>
-        <label htmlFor={field.name}>First Name:</label>
-        <input
-          name={field.name}
-          value={field.state.value}
-          onChange={(e) => field.handleChange(e.target.value)}
-        />
-        {field.state.meta.errors ? <em role="alert">{field.state.meta.errors.join(', ')}</em> : null}
-      </>
-    );
-  }}
+  {field => (
+    <>
+      <label htmlFor={field.name}>First Name:</label>
+      <input
+        name={field.name}
+        value={field.state.value}
+        type="number"
+        onChange={(e) => field.handleChange(e.target.valueAsNumber)}
+      />
+      {field.state.meta.errors ? <em role="alert">{field.state.meta.errors.join(', ')}</em> : null}
+    </>
+  )}
 </form.Field>
 ```
 
@@ -46,32 +39,24 @@ In the example above, the validation is done at each keystroke (`onChange`). If,
 
 ```tsx
 <form.Field
-  name="firstName"
-  // Implement onBlur instead of onChange here
-  onBlur={(value) => {
-    return !value
-      ? "A first name is required"
-      : value.length < 3
-        ? "First name must be at least 3 characters"
-        : undefined;
-  }}
+  name="age"
+  onBlur={val => val < 13 ? "You must be 13 to make an account" : undefined}
 >
-  {(field) => {
-    return (
-      <>
-        <label htmlFor={field.name}>First Name:</label>
-        <input
-          name={field.name}
-          value={field.state.value}
-          // Listen to the onBlur event on the field
-          onBlur={field.handleBlur}
-          // We always need to implement onChange, so that TanStack Form receives the changes
-          onChange={(e) => field.handleChange(e.target.value)}
-        />
-        {field.state.meta.errors ? <em role="alert">{field.state.meta.errors.join(', ')}</em> : null}
-      </>
-    );
-  }}
+  {field => (
+    <>
+      <label htmlFor={field.name}>First Name:</label>
+      <input
+        name={field.name}
+        value={field.state.value}
+        type="number"
+        // Listen to the onBlur event on the field
+        onBlur={field.handleBlur}
+        // We always need to implement onChange, so that TanStack Form receives the changes
+        onChange={(e) => field.handleChange(e.target.valueAsNumber)}
+      />
+      {field.state.meta.errors ? <em role="alert">{field.state.meta.errors.join(', ')}</em> : null}
+    </>
+  )}
 </form.Field>
 ```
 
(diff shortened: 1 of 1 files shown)


natedunn:
Thanks for bringing up the issues, especially with #2 being an absolute menace. @crutchcorn should these be broken out to their own issues for visibility or are they ok here?

juliendelort:
[This recent issue](https://github.com/TanStack/form/issues/486)  pretty much covers point 2.

I've opened issues and discussions for the other points:
- point 1: Issue https://github.com/TanStack/form/issues/489
- point 3: discussion https://github.com/TanStack/form/discussions/488
- point 4: discussion https://github.com/TanStack/form/discussions/487

crutchcorn:
@juliendelort love the efforts you've put into documenting this with us all and making things go from @tannerlinsley and my brain to more on-paper. This is huge, sincerely thank you.

Here's my thoughts, and let me know your thoughts;

- All 4 of the raise items are results of some flavor of bug in the codebase. We've since responded to the discussions, issues, and opened new issues to clarify
- As such, we shouldn't change our docs based off of bugs (ideally, we should just fix our bugs)

With this in mind, here's how I would expect TanStack Form to behave (just in one place to avoid confusion):

```tsx
  const form = useForm({
    defaultValues: {
      firstName: '',
      lastName: '',
    },
    validator: zodValidator,
    // This will be renamed in the future to avoid confusion with `change` DOM event
    // This validation will run on submit also
    onChange: z.object({
      firstName: z.string().min(3),
    }),
    onSubmit: async (values) => {
      console.log(values)
    },
  })

  return (
    <div>
      <form.Provider>
        <form
          onSubmit={(e) => {
            e.preventDefault()
            e.stopPropagation()
            void form.handleSubmit()
          }}
        >
          <div>
            <form.Field
              name="firstName"
              // This validator will run in addition to the form validator
              // In addition, this validation will run on submit also
              onChange={(value) =>
                value.includes('error') && 'No "error" allowed in first name'
              }
              // This validation will run on submit also
              onBlur={z.string().max(100)}
              onChangeAsyncDebounceMs={500}
              // This validation will run on submit also
              // And will run after the debounce time
              onChangeAsync={(value => checkDatabaseForName(value)}
              children={(field) => {
                // ...
              }}
            />
          </div>
          {/* ... */}
        </form>
      </form.Provider>
    </div>
  )
}
```

Does this help answer any questions that might arise from 1-4 as outlined here?

I'm happy to review the docs if this all sounds good to you @juliendelort - what do you think? :)

juliendelort:
Thanks @crutchcorn, sounds good to me and the code snippet helped a lot! I've made a few updates based on this and this is ready for a first round of reviews!

docs/guides/validation.md
```diff
@@ -205,8 +267,52 @@ This will debounce every async call with a 500ms delay. You can even override th
...
+        type="password"
+        onChange={(e) => field.handleChange(e.target.value)}
+      />
+      <FieldInfo field={field} />
+    </>
+  )}
+/>
+```
+
+
```
crutchcorn, line 314: I'd like to remove this temporarily and move this into an issue/discussion.

Right now, I'm not convinced this is a reliable mechanism for validating across fields. **If** this works, I'm almost certain it's due to a performance problem that we'll aim to fix using something akin to this:

https://houseform.dev/guides/performance-optimizations.html

docs/guides/validation.md
```diff
@@ -3,43 +3,113 @@ id: form-validation
...
+```tsx
+ <form.Field
+  name="firstName"
+  onChange={(value) => {
+    return !value
+      ? "A first name is required"
+      : value.length < 3
+        ? "First name must be at least 3 characters"
+        : undefined;
```
crutchcorn, line 26: This is admittedly our fault for having this example anywhere in our current docs; but people's names should not be validated for basically anything, especially length mins/maxes.

Let's change this validation example to something else - maybe back to our `age < 13` example? 😄

crutchcorn (changes requested):
Love the updates and tone of these new docs. Two thoughts we'll want to fix, then we can merge

docs/guides/validation.md
juliendelort: I agree. I've updated all the examples to show an "age" field instead of "first name".

docs/guides/validation.md
juliendelort: Makes sense! I've removed this part and created [a discussion](https://github.com/TanStack/form/discussions/494).

crutchcorn (approved):

juliendelort:
Hey @crutchcorn I don't know if it's related to this being merged but the validation page is not loading for me: https://tanstack.com/form/latest/docs/guides/validation

crutchcorn:
Fun!

CC @tannerlinsley

tannerlinsley:
I’ll debug this.

Savageman:
@crutchcorn Sorry for a late and unrelated question to this PR.

I saw this:
```
          onSubmit={(e) => {
            e.preventDefault()
            e.stopPropagation()
            void form.handleSubmit()
          }}
```

and I was wondering why the `void` was here, is it useful?

Issue #470, opened 2023-10-02
Nextjs useForm gives a type error


samyarkd:
### Describe the bug

When i use `useForm` i get an error saying use_isomorphic_layout_effect_cjs is not a function. and it's due to Nextjs server side rendering. 

Right now I'm getting over it using Nextjs dynamic import and disabling ssr but it's not a good approach.

when i want to use `useForm` hook i get the bellow error:

```bash
TypeError: use_isomorphic_layout_effect_cjs is not a function
    at useForm (/home/challen-quiz/dist/apps/challenquiz/.next/server/chunks/9063.js:828:3)
    at QuestionForm (/home/challen-quiz/dist/apps/challenquiz/.next/server/app/admin/add-question/page.js:442:44)
    at nP (/home/challen-quiz/node_modules/next/dist/compiled/next-server/app-page-experimental.runtime.prod.js:45:40134)
    at nP (/home/challen-quiz/node_modules/next/dist/compiled/next-server/app-page-experimental.runtime.prod.js:45:53256)
    at n$ (/home/challen-quiz/node_modules/next/dist/compiled/next-server/app-page-experimental.runtime.prod.js:45:56848)
    at nj (/home/challen-quiz/node_modules/next/dist/compiled/next-server/app-page-experimental.runtime.prod.js:45:60443)
    at nR (/home/challen-quiz/node_modules/next/dist/compiled/next-server/app-page-experimental.runtime.prod.js:45:59154)
    at n$ (/home/challen-quiz/node_modules/next/dist/compiled/next-server/app-page-experimental.runtime.prod.js:45:57094)
    at nj (/home/challen-quiz/node_modules/next/dist/compiled/next-server/app-page-experimental.runtime.prod.js:45:60443)
    at nP (/home/challen-quiz/node_modules/next/dist/compiled/next-server/app-page-experimental.runtime.prod.js:45:50666)

```

### Your minimal, reproducible example

https://stackblitz.com/edit/stackblitz-starters-bbkdbd?file=components%2Fform.tsx

### Steps to reproduce

1. create a new nextjs app project
2. create a client component
3. import useForm
4. use the client compoent
5. check the logs

### Expected behavior

i expected that the useForm would work in a client component

### How often does this bug happen?

Every time

### Platform

- OS Linux fedora
- Browzer: Edge
- Node version: 20
- Nextjs: 13.5.3
- @tanstack/react-form": "^0.3.6

### Tanstack Form adapter

react-form

### TanStack Form version

[...]

PR #485, opened 2023-10-24
fix: remove use-isomorphic-layout-effect package for nextjs compatibility

juliendelort:
Issue: https://github.com/TanStack/form/issues/470

The [use-isomorphic-layout-effect](https://github.com/Andarist/use-isomorphic-layout-effect) package seems to cause issues on NextJS.

See the detailed investigation [here](https://github.com/TanStack/form/issues/470#issuecomment-1777837403). 
TL;DR it may be a NextJS issue or an issue with `use-isomorphic-layout-effect` (it seems adding an "exports" section to its package.json file fixes the issue).

We only use the `useIsomorphicLayoutEffect` hook from that package and since its implementation is a one-liner, I've added it to `react-form`, which fixes the nextjs issue.

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "lib": ["ES2022"],
+    "lib": ["ES2022", "DOM"],
     "target": "ES2020",
     "module": "ES2020",
     "moduleResolution": "node",
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "lib": ["ES2022", "DOM"],
+    "lib": ["ES2022"],
     "target": "ES2020",
     "module": "ES2020",
     "moduleResolution": "node",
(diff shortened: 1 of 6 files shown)


tsconfig.base.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "lib": ["ES2022"],
+    "lib": ["ES2022", "DOM"],
```
juliendelort, line 3: Had to add "DOM" here to access `document` in `useIsomorphicLayoutEffect`.

crutchcorn: We can't have this in our codebase, as it makes it far too easy to add in a mistaken reference to the DOM in an otherwise runtime agnostic library.

Let's change this before merging

crutchcorn (changes requested):
Marking as request changes, as we need to update tsconfig to not include `DOM`

crutchcorn (approved):
I've fixed the outlined issues I had previously. This PR will be our `0.6.1` release

Issue #466, opened 2023-09-14
Re-Add Form-level Validation Logic


tannerlinsley:
Somewhere in our refactoring, the form-level validation logic was removed. We need to:

- Add it back
- Determine how it fits in with schema-based validation @crutchcorn is working on
- Potentially abstract shared logic to utils

hiddentao:
Yes please. Just missing this now.

crutchcorn:
Adding a bit of context, maybe @aadito123 can tackle while I'm working on schema based validation:

This code was removed here: https://github.com/TanStack/form/commit/884235f21197468c3d5d4eba024e58f826e4dfb2

And needs to be added back akin to this bit of code:

https://github.com/TanStack/form/blob/0159e61103e2d1fbceb5f9543d5d5ab2edb2ed4c/packages/form-core/src/FieldApi.ts#L290-L389

crutchcorn:
After talking with @aadito123, who's working on this issue (reassigning the GH issue), I think we need to change our `formError` to:

```ts
export type FormState<TData> = {
  formErrors?: ValidationError[]
  // Computed, just like `FieldAPI`
  formErrorMap?: ValidationErrorMap
  
  // Not in this PR, but in some follow-up
  fieldErrors?: Record<DeepKeys<TData>, ValidationError[]>
  fieldErrorMap?: Record<DeepKeys<TData>, ValidationErrorMap>

  // ...
}
```

So that our users can both:

- See the form errors on a per-validator method
- Match `FieldAPI`'s API
- See all fields' errors to render them somewhere outside of `Field` if required

PR #493, opened 2023-10-30
[WIP] Add back form validation

crutchcorn:
@aadito123 took a stab at adding back form validation - ala #466 - this is that work tracked in a PR so I don't lose track of it

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: Feature Requests & Questions
     url: https://github.com/TanStack/form/discussions
--- a/docs/config.json
+++ b/docs/config.json
@@ -157,6 +157,58 @@
           ]
         }
       ]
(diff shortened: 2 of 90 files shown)


aadito123:
Still WIP. The validation logic for onChange and onBlur run but the form states are not updating as expected. Also need to implement the onMount validation.

crutchcorn:
No worries @aadito123 - take your time 😄 

Anything I can do to help?

crutchcorn:
Per a convo with @aadito123 in DMs, we're deciding to add in a `mount` method to `FormAPI`. As a result, this will _technically_ be a breaking change, as our vanilla adapters will expect us to add `form.mount()` prior to `field.mount`

This allows us more flexibility in our future API development while also enabling `onMount` validators

Let's make sure we're updating our tests to reflect this change

aadito123:
Feeling pretty confident in the implementation so far. Only had one test case though, so I want to add all the applicable validation test cases from the `FieldApi` to the `FormApi` tests.

crutchcorn:
Migrated to PR #505 as I've somehow managed to break my ability to push to this PR

PR #496, opened 2023-10-31
feat: mount method on FormApi

aadito123:
As discussed in #493 

Edit:
Should mention that `onMountAsync` and `onMountAsyncDebounceMs` were removed since onMount callbacks should only run once.

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -24,8 +24,6 @@ export type FormOptions<TData, ValidatorType> = {
   asyncDebounceMs?: number
   validator?: ValidatorType
   onMount?: ValidateOrFn<TData, ValidatorType>
-  onMountAsync?: ValidateAsyncFn<TData, ValidatorType>
-  onMountAsyncDebounceMs?: number
   onChange?: ValidateOrFn<TData, ValidatorType>
   onChangeAsync?: ValidateAsyncFn<TData, ValidatorType>
   onChangeAsyncDebounceMs?: number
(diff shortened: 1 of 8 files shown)


packages/form-core/src/FormApi.ts
```diff
@@ -170,6 +168,18 @@ export class FormApi<TFormData, ValidatorType> {
...
+  mount = () => {
+    if (typeof this.options.onMount === 'function') {
+      return this.options.onMount(this.state.values, this)
+    }
+    if (this.options.validator) {
+      return (this.options.validator as Validator<TFormData>)().validate(
+        this.state.values,
+        this.options.onMount,
+      )
+    }
```
crutchcorn, line 180: Let's remove `validator`. It's mostly a holdover from the older API we had before `onMount`/etc.

crutchcorn (changes requested):
Small change I'd like us to make prior to merge.

Also, let's add `form.mount()` to all of the `form-core` tests. It'll help us avoid refactoring/debugging those tests in the future in case the `mount` ever becomes required.

packages/form-core/src/FormApi.ts
aadito123: Do you mean remove the `validator` option entirely from the form or do you mean we should not run the validator in the `mount` function? If we removed the validator option from the Form then you would have to manually specify the validator in every `Field`.

crutchcorn: OH! You're gonna laugh - I legit forgot what `validator` was 🤦‍♂️

Disregard!

crutchcorn (approved):

Issue #492, opened 2023-10-27
Add docs and tests for Field Arrays


crutchcorn:
Right now, we _do_ support Field Arrays, but their usage is not clear and I have low confidence as a maintainer that we won't break them in the future. Let's update docs and tests for field arrays so that we're able to ship this more confidently.

Christian24:
I just tested this in my prototype. How is the API meant to work? I just iterated over the array and built the field name with the index aka 
```js
{ name: `employees[${index}].color` }
``` 

This seems to work fine, but `DeepKeys` does not support it. Is this how you envision the API for FieldArrays though, @crutchcorn?

Christian24:
Upon further investigation it seems correct names look as follows:

```js
{ name: `employees.${index}.color` }
```

This works and types also work for this.

Christian24:
Interestingly, there seems to be some documentation on the website already: https://tanstack.com/form/latest/docs/guides/basic-concepts#array-fields

I did do a quick find usages on index though and Webstorm said it was used nowhere. So, I have no idea if these examples are correct.

Christian24:
There are also already tests present, so maybe this can be closed @crutchcorn

crutchcorn:
Yea this deffo can be closed, my bad

PR #500, opened 2023-11-01
Add basic tests for arrays

Christian24:
I added very basic tests for arrays. This is mostly trying to solve #492.


Christian24:
I noticed some tests already exist.

PR #510, opened 2023-11-05
Fix deleteField inside of an array

### LINKED ISSUE #509: deleteField does not work inside if field is inside an array
### Describe the bug

`FormApi.deleteField` does not work if said field is in an array. This is because we use `delete`, which does not work with the dot path notation. 

I will submit a PR soon

https://stackblitz.com/github/christian24/tanstack-form-lit-prototype?file=src%2Findex.ts

### Steps to reproduce

1. Check the employed checkbox
2. Write something in the job title field
3. uncheck the box
4. Values are still there

### Expected behavior

It works

Every time

### Screenshots or Videos

_No response_

### Platform

- latest Edge on Mac

### Tanstack Form adapter

None

### TanStack Form version

latest

### TypeScript version

_No response_

### Additional context

_No response_

Christian24:
This should fix deleting a field that is inside of an array. It should close #509


crutchcorn:
This now fixes #511 as well.

crutchcorn (approved):
Great work on this! Thank you!

PR #512, opened 2023-11-05
Add tests for getBy, setBy and deleteBy, small refactorings

### LINKED ISSUE #511: Add tests for getBy, setBy and deleteBy
We use the utils `getBy`, `setBy`and `deleteBy` (new in #510) extensively inside Tanstack Form. We should write tests for them. 

We should test easy and complex paths (and objects). We may also explore how to type them more properly and probably throw better errors if at all. 

CC @crutchcorn

Christian24:
This is based on #510. It needs to be merged first. This fixes #511.

@crutchcorn I tried doing something cool `DeepKeys` and `DeepValues`, but `DeepKeys` often seems to break when arrays are involved. Do you know why?


crutchcorn:
Closing in favor of #510, which now contain these changes

PR #517, opened 2023-11-09
docs(FormApi): Include link to TanStack Store docs beneath FormApi's Store instance

alexkatz:
The TanStack Store instance has some useful methods of its own. It may be helpful to provide a link to its docs from here.


alexkatz:
Result of [this discussion.](https://github.com/TanStack/form/discussions/515#discussioncomment-7525879)

crutchcorn (approved):
Thanks for this!

PR #518, opened 2023-11-10
docs and examples: Improve text and Valibot example

fabian-hiller:


crutchcorn (approved):
Thank you so much for helping with this, it means a lot ✨

PR #519, opened 2023-11-10
fix: async Valibot string validation

fabian-hiller:
Sorry, I added a bug in #518 and fixed it with this PR


crutchcorn (approved):
No worries! Thanks again :)

Issue #490, opened 2023-10-27
onChange and onBlur validation errors should prevent submission of form


crutchcorn:
Currently, when using a field with:

```tsx
<Field onChange={z.string().min(1)}>
```

When the user submits the form, it does not show a form error on submit. This causes headaches currently as it makes things harder to workflow properly.

vikaskumar89:
Hey @crutchcorn 
Can you assign this to me please

I-MirzaBaig:
Can you aasign this issue to me?

vikaskumar89:
hey @I-MirzaBaig 
I am working on this. Could you pick any other open issue.

I-MirzaBaig:
@vikaskumar89 Alright!

vikaskumar89:
Hey @crutchcorn 
The root cause of this error is because we have defined `undefined` for the `submit` cause that leads to validateSync returning early and not executing any `onBlur/onChange` validation.

One solution that comes to my mind is in case of `submit` cause, we can execute either `onChange` or `onBlur` validation(whichever is defined in the field) but this would lead to incorrect errorMap detail as the errorMapKey would be `submit` but the error would correspond to `onChange/onBlur`.

Do you have any alternate suggestion to fix this ?

vikaskumar89:
Due to this errorMap array, we are facing this issue in https://github.com/TanStack/form/issues/486, as the errorMap is not clear after a valid value is entered.

danteissaias:
Is there currently a workaround to this?

eastack:
I'm not sure when the maintainer will merge this, but is there any other way we can work around these issues temporarily?

Issue #486, opened 2023-10-26
If validation fails in `onSubmitAsync`, error does not resolve after input is changed


AkshayPathak:
### Describe the bug

To prevent the submission of a pristine form, the recommended way to validate is in the `onSubmitAsync`. In the code sandbox, the `onBlur` and `onSubmitAsync` functions both have the same `Zod` validator. 

When a pristine form is submitted, an error on the first name appears. However, if the input is modified, the error does not resolve and also does not allow the form to be submitted.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/confident-tereshkova-2xl2hj?embed=1&file=%2Fsrc%2Findex.tsx%3A1%2C1

### Steps to reproduce

1. Navigate to sandbox
2. Click `Submit` without touching any of the fields. You will see the validation error shown to the right of the `First Name` input.
3. Enter some text in the `First Name` input. Notice that the validation message does not disappear and even with a valid input, the `Submit` button is not enabled.

### Expected behavior

The error should clear after a validation error with a pristine form in the `onSubmitAsync` function and allow the form to be submitted. 

### How often does this bug happen?

Every time

### Platform

- OS: Linux
- Browsers: Chrome, Firefox

### Tanstack Form adapter

react-form

### TanStack Form version

0.4.1

juliendelort:
I've noticed this as well (I mentioned it in point 2 of [this PR description](https://github.com/TanStack/form/pull/484)). I believe this is a side effect of separating it out validation per event (it's possible to have different validation rules in `onChange`, `onBlur`, `onSubmitAsync` for example, so the concept of whether a field is valid or not at a given time is fuzzy).

I've opened a discussion [here](https://github.com/TanStack/form/discussions/487).

crutchcorn:
While this _appears_ unrelated at first glance to #490, I think that when we fix 490, we'll fix this one as well

PR #520, opened 2023-11-10
fix: should run validation on submit and clear error if a valid  value is entered

vikaskumar89:
This PR fixes issue https://github.com/TanStack/form/issues/490 and https://github.com/TanStack/form/issues/486.

## What is changed
- run onChange/onBlur validation on form submit . If the validation fails, form is disabled
- clear the onSubmit error If a valid value is entered.

PR-form-1


vikaskumar89:
> Is there anything currently preventing this from being merged?

Maintainer is very slow. cc. @tannerlinsley

eastack:
I'm not sure when the maintainer will merge this, but is there any other way we can work around these issues temporarily?

crutchcorn:
Hey all, maintainer here.

@vikaskumar89 I will need you to please start being patient with me. On top of TanStack Form not being 1.x yet (in which case bug fixes will be merged much faster), I also work on various other OSS projects, am trying to finish writing a book, am oftentimes busy with personal issues, and to top it off am not really getting paid for this work (minus a few generous donors albeit for relatively little amounts).

Please do not tag Tanner directly, and know that I will eventually come back to things here.

There is a reason I am "slow" :)

That all said, I will be reviewing this shortly.

packages/form-core/src/FieldApi.ts
```diff
@@ -332,10 +332,15 @@ export class FieldApi<
   validateSync = (value = this.state.value, cause: ValidationCause) => {
     const { onChange, onBlur } = this.options
     const validate =
-      cause === 'submit' ? undefined : cause === 'change' ? onChange : onBlur
+      cause === 'submit'
+        ? onChange
+          ? onChange
+          : onBlur
```
crutchcorn, line 338: This is not the correct backup to `cause === "submit" && !onChange`

Instead, we need to run both `onChange` **and** `onBlur`.

I'll take over this PR to make the appropriate changes.

crutchcorn, line 338: This is not the correct backup to `cause === "submit" && !onChange`

Instead, we need to run both `onChange` **and** `onBlur`.

I'll take over this PR to make the appropriate changes.

crutchcorn (changes requested):

crutchcorn (approved):
Issues with implementation have now been resolved.

In addition, I've added `<Field onSubmit={z.string().min(1)}/>` functionality, which was missing previously.

I also added more tests than before.

PR #524, opened 2023-11-26
docs(formApi.md): Fix a small typo by changing 'yonu' to 'your'

danisaza:
Fix a small typo by changing 'yonu' to 'your' in `formApi.md`.

No breaking changes.

### DIFF
--- a/docs/reference/formApi.md
+++ b/docs/reference/formApi.md
@@ -20,7 +20,7 @@ An object representing the options for a form.
 - ```tsx
     defaultValues?: TData
   ```
-  - Set initial values for yonu form.
+  - Set initial values for your form.
 - ```tsx
     defaultState?: Partial<FormState<TData>>
   ```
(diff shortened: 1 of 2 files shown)


danisaza:
Shoot - looks like CI is breaking, possibly because I didn't install dependencies, etc.

(I was betting on not having to do this since this is just a typo fix in the docs.)

I'll make sure that I can get the app / test suite running locally to see if that resolves the issue.

danisaza:
Okay, looks like CI is passing now, so this PR should be all set for review - but please let me know if I'm missing any context.

As you can probably tell, this is my first rodeo. 😅

crutchcorn:
Thanks for this! Typically I would suggest not modifying the lockfile or package.json unless needed (even if CI was failing) because it can be an attack vector for many OSS projects, but I manually reviewed and the changes look good.

Thanks again! 😊

PR #529, opened 2023-12-14
test: Add typescript tests for versions 4.8-5.2

aadito123:
These tests ensure that the repo can still be built with older versions of TypeScript, down to 4.8, but the final build still occurs with 5.2. This should ensure that users don't have breaking builds when using older TS versions.


crutchcorn (approved):
You're a legend, thank you for this!

PR #528, opened 2023-12-13
feat: SSR/Next Server Actions Support

crutchcorn:
This PR improves on our SSR story by implementing many of the ideas outlined in #480

IE; we now support Next/React Server Actions like so:

```tsx
// action.ts
'use server'

import { formFactory } from './shared-code'

export default async function someAction(prev: unknown, formData: FormData) {
  return await formFactory.validateFormData(formData)
}
```

```tsx
// shared-code.ts
import { createFormFactory } from '@tanstack/react-form'

export const formFactory = createFormFactory({
  defaultValues: {
    firstName: '',
    age: 0,
  },
  onServerValidate({ value }) {
    if (value.age < 12) {
      return 'Server validation: You must be at least 12 to sign up'
    }
  },
})
```

```tsx
/// <reference types="../../node_modules/@types/react-dom/experimental" />
'use client'

import { useFormState } from 'react-dom'
import someAction from './action'
import { formFactory } from './shared-code'
import { useTransform, mergeForm, FormApi } from '@tanstack/react-form'

export const ClientComp = () => {
  const [state, action] = useFormState(someAction, formFactory.initialFormState)

  const form = formFactory.useForm({
    transform: useTransform(
      (baseForm: FormApi<any, any>) => mergeForm(baseForm, state),
      [state],
    ),
  })

  const formErrors = form.useStore((state) => state.errors)

  return (
    <form.Provider>
      <form action={action as never} onSubmit={() => form.handleSubmit()}>
        {formErrors.map((error) => (
          <p key={error as string}>{error}</p>
        ))}

        <form.Field
          name="age"
          validators={{
            onChange: ({ value }) =>
              value < 8
                ? 'Client validation: You must be at least 8'
                : undefined,
          }}
        >
          {(form) => {
            return (
              <div>
                <input
                  name="age"
                  type="number"
                  value={form.state.value}
                  onChange={(e) => form.handleChange(e.target.valueAsNumber)}
                />
                {form.state.meta.errors.map((error) => (
                  <p key={error as string}>{error}</p>
                ))}
              </div>
            )
          }}
        </form.Field>
        <form.Subscribe
          selector={(state) => [state.canSubmit, state.isSubmitting]}
          children={([canSubmit, isSubmitting]) => (
            <button type="submit" disabled={!canSubmit}>
              {isSubmitting ? '...' : 'Submit'}
            </button>
          )}
        />
      </form>
    </form.Provider>
  )
}
```

To do this, we've implemented:

[...]

PR #533, opened 2023-12-28
chore: Workspace fixes

lachlancollins:
These are all changes taken from my work in the `tanstack/query` repo. Sorry they're kind of all over the place - the commit descriptions are a fairly good changelog.

There's a few more things that could be tightened up and might help identify the bug in #528 , but I thought I'd be incremental and submit these changes first.


package.json
```diff
@@ -113,17 +102,7 @@
...
-      "tsup@7.2.0": "patches/tsup@7.2.0.patch"
-    },
-    "overrides": {
-      "@tanstack/form-core": "workspace:*",
-      "@tanstack/react-form": "workspace:*",
-      "@tanstack/vue-form": "workspace:*",
-      "@tanstack/solid-form": "workspace:*",
-      "@tanstack/yup-form-adapter": "workspace:*",
-      "@tanstack/zod-form-adapter": "workspace:*",
-      "@tanstack/valibot-form-adapter": "workspace:*"
```
crutchcorn, line 126: These overrides are required for the examples to use the local version of the packages to avoid breakage

examples/vue/valibot/package.json
```diff
@@ -9,34 +9,15 @@
...
   "dependencies": {
-    "@tanstack/form-core": "0.11.0",
     "@tanstack/vue-form": "0.11.0",
     "@tanstack/valibot-form-adapter": "0.11.0",
     "vue": "^3.3.4",
-    "valibot": "^0.20.1",
-    "@tanstack/zod-form-adapter": "0.11.0",
-    "@tanstack/react-form": "0.11.0",
-    "@tanstack/yup-form-adapter": "0.11.0",
-    "@tanstack/solid-form": "0.11.0"
```
crutchcorn, line 20: This is theoretically nice to remove these packages, but they are added back automatically on every release via a CI/CD commit. We'll need to address this root cause first before merging

crutchcorn (changes requested):
Good changes overall! Let's solve these two problems before moving forward

package.json
lachlancollins: The `prefer-workspace-packages=true` option prevents this :)

examples/vue/valibot/package.json
lachlancollins: How strange! I'll have a look soon.

package.json
crutchcorn: Oh! Neat, TIL

examples/vue/valibot/package.json
lachlancollins: Fixed! It was the `Updating all example dependencies...` block in `scripts/publish.ts`. Shouldn't be necessary since pnpm will resolve to the internal workspace and users will resolve the latest semantic version. We don't have this in tanstack-query anymore.

crutchcorn (approved):

PR #534, opened 2023-12-29
Demo: Build with `tsc`

lachlancollins:
Part of the effort to address `tsup` issues.

Pros:
- The output is simply the TS stripped types, no bundling occurs
- Easy to transpile to target older ES spec

Cons:
- Only generates `.js` and `.d.ts` for ESM _or_ for CJS, not both at once
- To generate both, it would require a lot more config (is CJS still necessary?)


crutchcorn:
Superseded by #528 and migrating to Vite

Thanks so much for your help!

PR #535, opened 2023-12-29
Demo: Build with `vite`

lachlancollins:
Part of the effort to address `tsup` issues.

Pros:
- Very popular and maintained build tool
- Simple configuration

Cons:
- Bundles _everything_ by default, so need to tell it which dependencies to keep external (or add a plugin)
- Doesn't generate `.d.ts` by default, so need a separate tsc task or a separate plugin (which only generates for ESM _or_ CJS - is CJS still necessary?)
- Feels like it's designed for bundling apps, not libraries


crutchcorn:
Superseded by #528 and migrating to Vite

Thanks so much for your help!

PR #548, opened 2024-01-03
chore: Avoid running Vitest twice

lachlancollins:
`vitest --typecheck` runs all type tests _and_ all normal tests. It makes more sense to run these as a single task under `test:lib`. The Vite config has been updated to enable coverage and type tests without needing CLI flags.


crutchcorn (approved):
LGTM!

PR #550, opened 2024-01-04
fix: delete non-existent fields without errors

chrisirhc:
Hit this issue when I had a nested field on a field that didn't have a default value set up.


crutchcorn (approved):
Gonna merge this once it passes CI/CD - thanks for this!

PR #551, opened 2024-01-04
feat: support escaped paths in field name

chrisirhc:
Attempts to address https://github.com/TanStack/form/discussions/532 . 

This is a sketch implementation for feedback, using lodash's internals for get. 
Happy to roll a custom implementation but I thought of using an existing battle-tested implementation. 

This implementation handles some of the cases in https://github.com/lodash/lodash/blob/main/test/toPath.spec.js .

### DIFF
--- a/packages/form-core/src/utils.ts
+++ b/packages/form-core/src/utils.ts
@@ -114,35 +114,49 @@ export function deleteBy(obj: any, _path: any) {
   return doDelete(obj)
 }
 
-const reFindNumbers0 = /^(\d*)$/gm
-const reFindNumbers1 = /\.(\d*)\./gm
-const reFindNumbers2 = /^(\d*)\./gm
-const reFindNumbers3 = /\.(\d*$)/gm
-const reFindMultiplePeriods = /\.{2,}/gm
-
-const intPrefix = '__int__'
-const intReplace = `${intPrefix}$1`
+// This utilty is from stringToPath in lodash with minor modifications to extract numbers.
+// https://github.com/lodash/lodash/blob/aa18212085c52fc106d075319637b8729e0f179f/src/.internal/stringToPath.ts
+const charCodeOfDot = '.'.charCodeAt(0)
+const reEscapeChar = /\\(\\)?/g
+const rePropName = RegExp(
+  // Match anything that isn't a dot or bracket.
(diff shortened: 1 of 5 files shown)


packages/form-core/src/utils.ts
```diff
@@ -110,35 +110,49 @@ export function deleteBy(obj: any, _path: any) {
...
+    if (quote) {
+      key = subString.replace(reEscapeChar, '$1')
+    }
+    else if (expression) {
+      key = expression.trim()
+    }
+    // key difference from lodash string to path algoritm
+    if ((/^\d+$/).test(key)) {
+      key = parseInt(key, 10);
+    }
```
chrisirhc, line 151: Main modification I made based on the previous logic in `makePathArray`.

crutchcorn:
This looks good overall! The only concern I have now is typings. We'll want a way to merge the types from `string[]` to `string`. This can be done like so:

https://github.com/crutchcorn/ts-util-helpers/blob/main/src/concat-string.ts#L1-L7

But then the question becomes "how do we get `TName` inference working for this?

chrisirhc:
> This looks good overall! The only concern I have now is typings. We'll want a way to merge the types from `string[]` to `string`. 
Not sure I get this part of your comment. Where are you referring to the `string[]` type? Is it the `makePathArray`?

> But then the question becomes "how do we get `TName` inference working for this?
Just to make sure I get what you're referring to. You're referring to making sure that the `DeepKeys`  https://github.com/TanStack/form/blob/77f119b0a6e6caccb8fdc3d3cdeda97b3560d1e5/packages/form-core/src/utils.ts#L299 can work for the `['']` syntax, right? I'll look in this.

crutchcorn:
I wouldn't modify `DeepKeys`. I would modify the instance where `TName` is being passed and introduce a new utility type of some kind.

chrisirhc:
Added spec updates and also utilities as discussed. Ready for comment/review.

crutchcorn:
Hey, I am super sorry - but we've made **drastic** changes to the internals of these files. It's honestly better to close out this PR and revisit them. I _think_ our types support this now, but our logic certainly doesn't.

Let's track the progress of this in: #639

Issue #503, opened 2023-11-03
Caching Primitives


crutchcorn:
Ideally, it would be awesome to have a form that allows you to persist values to a form and easily load them back into memory when the user navigates away.

More details and a potential API to come soon.

CC @tannerlinsley per: https://twitter.com/tannerlinsley/status/1720575334909964612

crutchcorn:
Prior art: https://formkit.com/plugins/local-storage

j0rgedev:
Hello @crutchcorn ! Any update on this? 👀

zainthedev:
Looking forward to this, may I throw in a request to make sure that it's super easy to clear the saved values?

PR #561, opened 2024-01-08
Form persistence

aadito123:
Attempt at addresssing #503

This PR includes, the form-persist-core package with tests and the changes to the form-core which make it work. The form-persist-core exports two main things: `class PersisterAPI` and the helper `function createFormPersister`.

The persister is an optional field to the FormApi class, which when passed, triggers restoring and persistence on FormState changes. For persisting to localStorage, it is:
 ```typescript
const persisterAPI = new PersisterAPI({ storage: window.localStorage })
new FormApi({ persister: createFormPersister(persisterAPI, 'myFormAPI') })
new FormApi({ persister: createFormPersister(persisterAPI, 'myFormAPI2') })
```

The storage field can also be async, as inspired by the persister in TanStack/query.

Opinions needed for:
 - API for framework adapters. For React and Solid adapters, I am thinking there should be a global provider for the PersisterAPI, and on then something like `useForm({ persistFormKey: 'myFormKey' })` or `useForm({ formPersister: useFormPersister('myFormKey') })` would be all that is needed. This way the dev can also opt out of persistence for a specific form. Very open to opinions and suggestions. Don't have anything in mind for the Vue version but I don't see why it cant just be the same thing.
 
 - Persistence state i.e. `isRestored` and `isRestoring` are part of the FormState, but I don't really like having that there. Would it be better to have it in the PersisterAPI itself? If it is moved into the persister, we will need some way to pass that state back to the user, which might necessitate something like 'useIsRestoring` or something similar. Again, open to opinions and suggestions.

- Fine-grained persistence. This might be more of a future thing, but perhaps, there should be a way to opt-out of having certain fields in the persistence. This could certainly just be done by ignoring the form-persistence-core package and passing in a completely custom Persister to the FormApi that can process the FormState as wanted. Just a thought. 

- Custom serializer. Right now I am just using JSON.parse and JSON.stringify to serial the form state. It should be more than enough considering that state is only strings and numbers. TanStack/query allows their persister to have custom serializers. I don't see the usecases for forms though, although open to hear them, if any. It is very easy to add it in.

To be done:
- DOCS! Haven't been pulling my weight with the docs so will try to include some in follow up commits.
- Framework adapters. I feel confident in making the React and Solid adapters but might need help with the Vue one. 

Am I missing anything else here?

### CODE REVIEW

**Comment by crutchcorn** [comment]
This is incredible work! I'll have to think more about this issue when I have a bit more time, but at first glance(s) I think you're more than on the right track.

I'll follow-up this weekend sometime to add thoughts on what we have opinions on. In the meantime, let's see if the community has any thoughts on this API

**Comment by crutchcorn** [comment]
> Buckle in, folks, this is a long one!

# Framework Adapter API

1) On the note of Framework adapters, I'd like us to do a modification of the second syntax you suggested:

```tsx
useForm({
    formPersister: useFormPersister({
        persistKey: ['myFormKey']
    })
})
```

This would do a few things for us:

1a) Match TanStack Query's APIs for a more consistent cross-ecosystem feel and consistency in codebases (able to follow @tkdodo's caching key suggestions by keeping related keys in an object/functions)
1b) By using a hook here, we might be able to add additional memoization like `useMemo` or whatnot if we need to rely on that internally. It overall gives _us_ as maintainers more flexibility

1c) `useFormPersister` allows us to return a memoized `FormPersisterAPI.ts` instance with an `update` function ran during renders, which aligns with how all of our other hooks work (more or less), so there's consistency in implementation

1d) We'd then be able to pass additional keys to modify the persistence functionality. IE:

```tsx
useForm({
    formPersister: useFormPersister({
        persistKey: ['myFormKey'],
        omitKeys: ['password']
    })
})
```

# Persisted State APIs

2) I'd like to propose one of two APIs to solve this:

2a) Mutate `form` using `useFormPersister`:

```tsx
const form = useForm({
    defaultValues: {/*...*/}
})

const {isRestoring, isRestored} = useFormPersister({
    form,
    persistKey: ['myFormKey'],
    omitKeys: ['password']
})
```

Which would allow us to reuse the `mergeForm` function (currently only used for SSR) to do something like:

```tsx
const useFormPersister = ({form, persistKey, omitKeys}) => {
	const {state, persister} = getFormPersister({persistKey, omitKeys})
	form.persister = persister;
    // Mutates `form` state, doesn't return a new instance
    mergeForm(form, state)
}
```

2b) Use the API outlined in #1 to extract from the `formPersister` API directly:

```tsx
const form = useForm({
    formPersister: useFormPersister({
        persistKey: ['myFormKey'],
        omitKeys: ['password']
    })
})

// Under-the-hood, these fields are part of the FormPersisterAPI.ts file, not FormAPI.ts
const {isRestoring, isRestored} = form.formPersister;
```

# Fine-Grained Persistence

3) I think we solved this using the API outlined in #1, which enables us to exclude or include keys either like:

3a)

```tsx
useForm({
    formPersister: useFormPersister({
        persistKey: ['myFormKey'],
        omitKeys: ['password'], // Or maybe: (fieldName, fieldValue) => key !== password (?)
        // Or
		includeKeys: ['name']
    })
})
```

3b)

```tsx
useForm({
    formPersister: useFormPersister({
        persistKey: ['myFormKey'],
        shouldPersist(fieldName, fieldValue) {
			if (fieldName === "password") return false;
            return true;
		}
    })
})
```

# Custom Serializer

4) I vote that we allow support for custom serializers. This does a few things for us as well:

4a) Consistency with TanStack Query

4b) Persisting built-in objects like `Date`s using `JSON5` without us having to roll it as a dependency

aadito123:
I think just about everything is implemented. Next great effort will come in writing the docs and examples. Would love help on that. 

Notes:
1. Framework Adapters
Think we are basically on the same page. For the `formKey`, I left the type as a string, since I am not sure what the advantage of having an array would be. My understanding is that for Query, the array is used for cache invalidation at different levels. I am not sure if this would be applicable to the form persister. As such, I made the API similar to the Query persister.

2. Persistence APIs
I found 2b to be the cleanest, but made a few modifications. I attached the restore states to the `store` on the `FormApi` instance itself. This gives us the `watch` and `Subscribe` functionality for free. My main concern is that it pushes quite a bit of persistence logic into the `form-core` but I am not sure how else we could make it work, since the `restore` function needs to be called in the `onUpdate` for the store. Definitely would like some more suggestions for this.

3 & 4. Totally agreed, so went ahead and added those in, too.

ricardopacheco:
@aadito123 This feature is amazing, and I'd like to see it implemented as soon as possible. What exactly do you need? If it's documentation, I can help.

aadito123:
@ricardopacheco feel bad for forgetting about this. Yes, certainly docs but also there are a few failing tests. Considering how long it has been, I'm sure it will need to be reworked again to fit any other changes in the repo. I would say this PR is honestly 75% done. If you want to take it past the finish line, you are welcome to do so. Reach out on the TanStack discord server if you need any help with that. Username is aadit0 there.

TkDodo:
@crutchcorn sent me here in 2026 so let me note down my thoughts on persistence APIs:

> Consistency with TanStack Query

TanStack Query uses its `hydration` API for both SSR and persistence, and this has served us well:

```
const state = dehydrate(queryClient, options)
```

basically takes the `queryClient` and creates a serializable object out of it that you can then send over the wire (SSR) or to a persister (localstorage). The reverse is:

```
hydrate(queryClient, state, options)
```

which takes the passed `state` and puts it _into_ the `queryClient`.

---

With that in mind, I think a `form` instance is similar to a `queryClient` here: Something that you’d want to serialize in full to a different place. If the `formState` is already fully serializable, that’s great! I would still add this layer so that you can make the state non-serializable (have function as options etc) somewhen.

The Framework integration doesn’t happen with a hook, but with a `<Provider>`. The reason is that we don’t do persistence per-query, but per-client. By the time you call `useQuery`, restore has already happened. I think this is different for `form` because you’d want to store each `formState` into its own place in the storage.

some more unstructured thoughts:
- `new PersisterAPI` is pretty much `createPersister` in Query.
- omitKeys: ['password'], // Or maybe: (fieldName, fieldValue) => key !== password (?)
I think the callback function API is better. We have this as part of `options` on `dehydrate`, where `options.shouldDehydrateQuery` is `(query: Query) => boolean`. This allows for tagging certain queries for exclusion via things other than its key, and those APIs generally evolve better than a fixed set of keys.
- have you thought about race conditions with `initialValues` ? Like what happens if restore for storage is async and `initialValues` comes from a Query. What should overwrite what?
- Similarly, since restore can be `async`, which state would the form be in while restore is taking place? What if users start to type while restore is still going on etc..

PR #562, opened 2024-01-08
Add IDs to input elements in examples and remove empty .prettierrc files

michaelbrusegard:
1. Input elements were missing IDs, which are necessary for labels to reference them correctly using the 'for' attribute. This is better for accessibility and now when you click the labels it should automatically focus the input.
2. Empty `.prettierrc` files were overwriting the global Prettier configuration in the react-examples causing prettier to replace all singlequotes with doublequotes and add semicolons.


crutchcorn:
Can we split out the prettier changes from the ID changes into two PRs? :)

michaelbrusegard:
I will do that

PR #564, opened 2024-01-09
fix: adds IDs to input elements with labels

michaelbrusegard:
This PR adds IDs to the input elements with labels so they are referenced correctly using the "for" attribute.

The formatting changes come from the empty .prettierrc files in the examples which override the global config. There will be a secondary PR to fix this, but that is the reason for the semicolons and singlequotes being changed here.


michaelbrusegard:
Since this messes up the stylings because of that the empty .prettierrc files have not been removed yet I will add another commit to fix that as if they were already removed

crutchcorn:
@michaelbrusegard let's merge `main` into this & re-run prettier to fix the formatting issues

crutchcorn (approved):

Issue #571, opened 2024-01-12
'Cannot convert undefined or null to object' when rendering field conditionally with React.StrictMode 


Hatzl:
### Describe the bug

If you conditionally render a field with React.StrictMode activated, the library crashes.
When disabling React.StrictMode, everything works fine.

### Your minimal, reproducible example

https://codesandbox.io/p/devbox/tanstack-form-conditional-field-kfljkr

### Steps to reproduce

1. Click "Submit" - You will get a error "A first name is required"
2. Click "Hide field" - This will remove First Name field
3. Click "Submit" again - The app will crash

### Expected behavior

The form should submit without crashing. 

### How often does this bug happen?

Every time

### Platform

OS: MacOS
Browser: Chrome

### Tanstack Form adapter

react-form

### TanStack Form version

v0.13.3

### TypeScript version

v5.2.2

crutchcorn:
Wow! Great catch - we'll tackle this ASAP

Hatzl:
Yes, it took some time to find the problem 😀
Thank you for your support!

Hatzl:
I think it has something to do with issue #495 . The crash point is also
`state.meta.errors = Object.values(state.meta.errorMap).filter(
            (val) => val !== void 0
          );` because errorMap is not defined.

fulopkovacs:
I don't know if I add anything valuable here, but to me, it looks like a new field instance (in `form.getFieldInfo('firstName').instances`) is created every time the `firstName` field is rendered, but deleted only every second time it's destroyed. I suspect that this is because the 1st runs are always triggered by `<StrictMode>` and there's [a piece of logic in the `useIsomorphicEffectOnce()` hook](https://github.com/TanStack/form/blob/08ad74d07d1dcaf117a9ef0390e937cf20edbcd6/packages/react-form/src/useIsomorphicEffectOnce.ts#L28-L32), that prevents the cleanup function from being executed for _"dummy React cycles"_ (this hook is [used in `useField()`](https://github.com/TanStack/form/blob/08ad74d07d1dcaf117a9ef0390e937cf20edbcd6/packages/react-form/src/useField.tsx#L109-L113)):

I think this is what leads to the number of `Error messages` growing  every time we toggle the visibility of the `firstName` field in the [CodeSandbox from the description](https://codesandbox.io/p/devbox/tanstack-form-conditional-field-kfljkr):



[...]

PR #576, opened 2024-01-22
fix: Conditionally rendering fields in <StrictMode>

fulopkovacs:
Addresses https://github.com/TanStack/form/issues/571.

# TL;DR

It seems like [the bug]( https://github.com/TanStack/form/issues/571) can be tracked to `useField` creating duplicate instances of the field (`FieldApi`) in strict mode and not cleaning them up properly. This PR at this point is just a proof of concept for a solution.

# Details of the bug

Basically, the `useField` hook runs twice in strict mode, and creates a new instance of `FieldApi` every time it runs ([source](https://github.com/TanStack/form/blob/08ad74d07d1dcaf117a9ef0390e937cf20edbcd6/packages/react-form/src/useField.tsx#L70-L89)):
```tsx
  const [fieldApi] = useState(() => {
    [...]
    
    const api = new FieldApi({
      ...opts,
      form: formApi as never,
      // TODO: Fix typings to include `index` and `parentFieldName`, if present
      name: name as typeof opts.name as never,
    })


    api.Field = Field as never


    return api
  })
```

I think that problem is that the cleanup function only runs once, because it's wrapped in a `useIsomorphicEffectOnce` hook [source](https://github.com/TanStack/form/blob/08ad74d07d1dcaf117a9ef0390e937cf20edbcd6/packages/react-form/src/useField.tsx#L109-L119):
```tsx
  useIsomorphicEffectOnce(() => {
    return () => {
      unmountFn.current?.()
    }
  })


  // We have to mount it right as soon as it renders, otherwise we get:
  // https://github.com/TanStack/form/issues/523
  if (!unmountFn.current) {
    unmountFn.current = fieldApi.mount()
  }
```

Honestly, there's a lot of things happening here, so I'm not a 100% percent sure I read the situation correctly. 😆 

# My solution and questions

I thought the best way to approach this issue is preventing a duplicate `FieldInstance` from being created, by checking if there's an instance already in the initialization function of the `useState` hook in `useField`. My question is: how dangerous is it? 🤣  In what cases would the field have multiple instances (in general, not just at this point)? Also, I'd love to understand a bit more about `tanstack/store`, especially if it's important here (a quick overview would be enough). I'm happy to get guidance on Discord, if a (or rather _the_) maintainer has time for it.

### DIFF
--- a/packages/react-form/src/useField.tsx
+++ b/packages/react-form/src/useField.tsx
@@ -76,6 +76,28 @@ export function useField<
       .filter((d) => d !== undefined)
       .join('.')
 
+    // NOTE: I know, it's a type error, but this is not the final code
+    const previousInstance:
+      | FieldApi<
+          TParentData,
+          TName,
+          TFieldValidator,
+          TFormValidator,
+          DeepValue<TParentData, TName>
+        >
+      | undefined = Object.values(formApi.getFieldInfo(name).instances)[0]
+
+    if (previousInstance) {
+      /* NOTE: In <StrictMode>, an instance of `FieldApi` is already created
+      the first time this hook ran, but it will not get cleaned up on its own,
+      when the component unmounts, because of the  useIsomorphicEffectOnce() hook,
+      which doesn't execute the cleanup function on "dummy React cycles".
+
+      This code change seems to solve https://github.com/TanStack/form/issues/571
+      */
+      return previousInstance
+    }
+
     const api = new FieldApi({
       ...opts,
       form: formApi as never,


crutchcorn:
> creates a new instance of FieldApi every time it runs

I'm not sure this is right. I thought, and I could be wrong, that a function passed to `useState` only ever executes once, even across duplicate effect runs.

Let me check and confirm this is the case. Either way, I wouldn't want to go down this path as I think there are other options available to us.

fulopkovacs:
Hmmm, I think the docs do say that the initializer function of a `useState` will be executed twice in Strict Mode:

> In Strict Mode, React will call your initializer function twice in order to [help you find accidental impurities.](https://react.dev/reference/react/useState#my-initializer-or-updater-function-runs-twice) This is development-only behavior and does not affect production. If your initializer function is pure (as it should be), this should not affect the behavior. The result from one of the calls will be ignored. ([docs](https://react.dev/reference/react/useState#caveats))

And if we modify an external state in the initializer function, then we won't have a clean start for the second render. Check this (horrendous and client-only) modification of your example ([see on StackBlitz](https://stackblitz.com/edit/vitejs-vite-qtdxrw?file=src%2FApp.tsx)):
```tsx
import { useState, useReducer } from 'react';
import reactLogo from './assets/react.svg';
import viteLogo from '/vite.svg';
import './App.css';

function App() {
  const [count, setCount] = useState(() => {
    if (!window.executeCount) window.executeCount = 0;
    window.executeCount++;
    return {
      count: window.executeCount,
    };
  });
  const [_, rerender] = useReducer(() => ({}), {});

  return (
    <div>
      <button onClick={() => rerender()}>{count.count}</button>
      <button onClick={() => setCount((v) => ({ count: v.count++ }))}>
        {count.count}
      </button>
    </div>
  );
}

export default App;
```

For me, the counter shows `2` instead of `1` the first time I see it.

Is it really not possible that we have an impure initializer function for a `useState` hook in `useField()`? 🤔 

> Either way, I wouldn't want to go down this path as I think there are other options available to us.
🙌 Totally understandable, I was very new to the repo when I made this PR (still am), so I just tried the solution that was the most accessible for me, which is probably not the most optimal one 😅.

crutchcorn:
OMG you're right. I never knew about this 🫠🫠

Let me see if I can't figure out a different solution to this problem knowing this now

crutchcorn:
OK so things that don't seem to work include:

- `useRef` outside of `useState`, early return if `useRef` is defined
- Replacing `useState` with `useMemo`

I like your solution the best, but let's add it into:

#606 instead of this PR

PR #598, opened 2024-02-11
feat(form-core): isDirty flag

fulopkovacs:
Adding the `isDirty` flag to the form and fields inspired by [`react-hook-form`](https://www.react-hook-form.com/api/useform/formstate/).

# Description

| | When is it `true`? |
| --- | --- |
| `field.state.meta.isDirty` | the value of the field does not match the default value of the field |
| `form.state.isDirty` | there's at least one _"dirty"_ field in the form |

# TODO
- [x] Implement the flag for fields with primitive values (e.g.: string, number)
- [ ] Implement the flag for fields with non-primitive values (e.g.: array fields)
- [ ] document the feature

# Motivation

This feature was request on Discord multiple times:
- https://discord.com/channels/719702312431386674/1100437019857014895/1191884556341620816
- https://discord.com/channels/719702312431386674/1199336913862135900


examples/react/simple/src/index.tsx
fulopkovacs, line 1: I only updated this file to make the PR more easy to test, we don't have to keep these changes!

packages/form-core/src/FieldApi.ts
```diff
@@ -326,6 +328,13 @@ export class FieldApi<
             ? state.meta.errors
             : []
 
+          const fieldDefaultValue = getBy(
+            this.form.options.defaultValues,
+            this.name,
```
fulopkovacs, line 333: Is it OK to assume that passing `this.form.options.defaultValues` as the first argument to `getBy()` will never result in an error? `defaultValues` cannot be `null`,  or any other value that could cause us trouble here, right? 😅

packages/form-core/src/FieldApi.ts
```diff
@@ -696,3 +705,46 @@ function getErrorMapKey(cause: ValidationCause) {
       return 'onChange'
   }
 }
+
+export function deepEqual<T>(objA: T, objB: T): boolean {
```
fulopkovacs, line 709: This function is a modified version of `shallow` from `tanstack/store`: https://github.com/TanStack/store/blob/be32993f5224c8e60ce6c3fe6674ec5907713f80/packages/react-store/src/index.ts#L27-L55

examples/react/simple/src/index.tsx
```diff
@@ -94,11 +95,20 @@ export default function App() {
...
+            ]}
+            children={([canSubmit, isSubmitting, isDirty]) => (
+              <>
+                <button type="submit" disabled={!canSubmit}>
+                  {isSubmitting ? '...' : 'Submit'}
+                </button>
+                <p>
+                  <code>isDirty:{isDirty ? 'true' : 'false'}</code>
+                </p>
+              </>
```
crutchcorn, line 111: Let's revert this from the example :)

crutchcorn:
Hey @fulopkovacs sorry for not clarifying this much sooner - as I was busy wrapping up my book. Now that I've done so, let my sample some writing from a (currently unpublished) blog post I wrote about forms validation states:


> One feature that's added with reactive forms is the concept of an input's state. An input can have many different states:
> 
> - "Touched" - When the user has interacted with a given field, even if they haven't input anything
>   - Clicking on the input
>   - Tabbing through an input
>   - Typing data into input
> - "Pristine" - The user has not yet input data into the field
>   - Comes before "touching" said field if the user has not interacted with it in any way
>   - Comes between "touched" and "dirty" when the user has "touched" the field but has not put data in
> - "Dirty"  - When the user has input data into the field
>   - Comes after "touching" said field
>   - Opposite of "pristine"
> - "Disabled" - Inputs that the user should not be able to add values into
> 
> While some of these states are mutually exclusive, an input may have more than one of these states active at a time. For example, a field that the user has typed into has both "dirty" and "touched" states applied at the same time.
>
> These states can then be used to apply different styling or logic to each of the input's associated elements. For example, a field that is `required && touched && pristine`, meaning that the user has clicked on the field, not input data into the field, but the field requires a user's input. In this instance, an implementation might show a `"This field is required"` error message.

As we can see here, `dirty` isn't _quite_ defined as `Value is not the same as defaultValue`, since the user could have touched a field, typed a different value, then changed it back to the original.

The way I did this in HouseForm was by:

- Setting `isDirty` in `setValue` (https://github.com/houseform/houseform/blob/main/lib/field/use-field-like.ts#L268)
- Setting `isTouched` in `onBlur` (https://github.com/houseform/houseform/blob/main/lib/field/field.tsx#L60)

I think we should set up our structure to be very similar and even provide `pristine` as a helper property.

What do you think?

crutchcorn:
The `isDirty` flag functions the way you'd outlined in RHF, that is true. However, I think that implementation of `isDirty` is flawed and would lead to bug reports in my user testing at work.

Consider the following:

- The user has typed a value into the form
- The user has deletes the value in the form
- The user tries to navigate backwards

Should they be prevented because they did, technically, interact with the form. I'd say "yes" and vote that to be our default functionality.

Instead of changing that, let's provide the `isDirty`, `isPristine` and other options as I outlined, but also provide the `defaultValue` to the `form.Field` component function so that users can write their own `isNotDefault` state if they wanted to.

fulopkovacs:
>  Instead of changing that, let's provide the isDirty, isPristine and other options as I outlined, but also provide the defaultValue to the form.Field component function so that users can write their own isNotDefault state if they wanted to.
> 
> We can even make a documentation note for this if users are coming from RHF

Nice! I think this is the best solution. It allows us to follow _Cruthchley's Four Input States Model™️_ , while providing an API for our users to get the data they were asking for. ☺️ 

I'll update the PR!

crutchcorn:
Closing as I know @fulopkovacs has a more up-to-date PR coming soon 🎉🎉

PR #599, opened 2024-02-15
fix: Partial form types

SagnikPradhan:


packages/form-core/src/tests/FieldApi.test-d.ts
aadito123, line 1: Can you add 2 more test cases?
1. Nothing is passed to FormApi constructor
2. FormApi constructor gets partial default values for the required fields in the generic argument. Just wanna make sure that if people set required fields in the generic, then they provide default values for them.

SagnikPradhan:
LMK if that is better? @aadito123

aadito123:
Thanks for contributing!

PR #601, opened 2024-02-15
Expose useFormContext

UrbanChrisy:
Exposes the form context via the already made hook.

Stop prop drilling the `formAPI` down the tree.


UrbanChrisy:
@crutchcorn How would you nest a field in another component within the form?

Or is the idea to force you to pass it down so it's all typesafe?

xmsz-stu:
+1

PR #606, opened 2024-02-25
Remove UID and Field Instances

### LINKED ISSUE #571: 'Cannot convert undefined or null to object' when rendering field conditionally with React.StrictMode

crutchcorn:
This PR removes the concept of multiple `instances` of a Field for a given `name`, which in turn removes UID and, in happenstance, fixes #571.

This should be good-to-go, but breaks the ability to have multiple `Field` items with the same `name`. The question is if any of the @TanStack/form-maintainers can think of a reason why we might need that.


crutchcorn:
Updating some of this code based on: #576

packages/react-form/src/tests/useField.test.tsx
```diff
@@ -462,4 +502,115 @@ describe('useField', () => {
...
+                />
+              </div>
+              <form.Subscribe
+                selector={(state) => [state.canSubmit, state.isSubmitting]}
+                children={([canSubmit, isSubmitting]) => (
+                  <button type="submit" disabled={!canSubmit}>
+                    {isSubmitting ? '...' : 'Submit'}
+                  </button>
+                )}
+              />
```
fulopkovacs, line 591: ## TL;DR
Currently, this part [makes the CI fail](https://cloud.nx.app/runs/jp3SAmy7kl), because of [a bug in TypeScript](https://github.com/microsoft/TypeScript/issues/52786) `5.0.4` (`4.9.1` has this error too, but [we use a different tsconfig file for it](https://github.com/TanStack/form/blob/ab591c93afe7e536755e4be1abe1e8a1bd938f02/packages/react-form/package.json#L17), that [excludes the `useField.test.tsx` file from the compilation](https://github.com/TanStack/form/blob/ab591c93afe7e536755e4be1abe1e8a1bd938f02/packages/react-form/tsconfig.legacy.json#L8), so we don't see any errors for `4.9.1` in the CI).

## 🐛 The bug in TS `5.0.4` 

> [!NOTE]
> ✨  I made two TS Playground projects to explain the problem in a practical way:
> - [❌ `5.0.4`](https://www.typescriptlang.org/play?ts=5.0.4#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wG4AocpAD0ljhgE8wk4A5CASQDtMkoAPABUAfHAC8cANpCAujLg0YSbgBMAznBTdGcAPxwADHABccbkgBu-WRXJMWcAGLQQAZRgplwl1BAARLxQxSQBvcgBIVSCzIV8AoPIAX0oHVgAFfnUIbgk4cIjuFBAkM3UYKGBuAHMAGkiUatLzAFcQACN+ZMoq5ShMdFY3Fva0Ss70nDB1YTckABskDCRVMQL1BaWYaD0zAApyr2aOHj5BeI8jgUyobO4REQBKCTEhOcXl1Ui0AAtgedURG4+wOnmUZhOvH4s02nyeL0QxBgADpkOgYBxVEhngAfRHo1FIzFIbrkTAtbgYYA5ODDdrqMbATqzd5bFZ5C5gpDXLI5B57MBTdRmOmMiZClmw5SrZ4FIgwFpQXICVTASwidQjOBgeYzAD0qvVpL1erggFByOBCP6aYCaCAAazgVTgABZkQBOZEAVlqcC9yMMyJdWjUfuRAEZkQA2cgCOkM8YkiIbD7bKDiUKgo7PcRiQ7KFIRX7-QEqDP51i5uB7SIRFVqkQAd1YKCIDB+rGLAKBAgNDcijxSepEdhNcEAMuRuz0+sOBl3kMcW-2RmNxkYJplJlNbaAZrPKHNiKQDXVIX0njayQtd0vcPdSdoQCDzcO+x-PgBMskP1dr9fVzZaG2MAdnAN49n2RoRIOC4jpQi6uh63q+v6c4hqoYYrrG8ZikmoTIgRBTJlK0D7BWP7Higp7nlRl71BERZ-N2Kj7A+T4vm+7Ffj+NYMXWhpNi2wGgeBKi9gJtaPJESRDiOQA)
> - [✅  `5.1.6`](https://www.typescriptlang.org/play?ts=5.1.6#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wG4AocpAD0ljhgE8wk4A5CASQDtMkoAPABUAfHAC8cANpCAujLg0YSbgBMAznBTdGcAPxwADHABccbkgBu-WRXJMWcAGLQQAZRgplwl1BAARLxQxSQBvcgBIVSCzIV8AoPIAX0oHVgAFfnUIbgk4cIjuFBAkM3UYKGBuAHMAGkiUatLzAFcQACN+ZMoq5ShMdFY3Fva0Ss70nDB1YTckABskDCRVMQL1BaWYaD0zAApyr2aOHj5BeI8jgUyobO4REQBKCTEhOcXl1Ui0AAtgedURG4+wOnmUZhOvH4s02nyeL0QxBgADpkOgYBxVEhngAfRHo1FIzFIbrkTAtbgYYA5ODDdrqMbATqzd5bFZ5C5gpDXLI5B57MBTdRmOmMiZClmw5SrZ4FIgwFpQXICVTASwidQjOBgeYzAD0qvVpL1erggFByOBCP6aYCaCAAazgVTgABZkQBOZEAVlqcC9yMMyJdWjUfuRAEZkQA2cgCOkM8YkiIbD7bKDiUKgo7PcRiQ7KFIRX7-QEqDP51i5uB7SIRFVqkQAd1YKCIDB+rGLAKBAgNDcijxSepEdhNcEAMuRuz0+sOBl3kMcW-2RmNxkYJplJlNbaAZrPKHNiKQDXVIX0njayQtd0vcPdSdoQCDzcO+x-PgBMskP1dr9fVzZaG2MAdnAN49n2RoRIOC4jpQi6uh63q+v6c4hqoYYrrG8ZikmoTIgRBTJlK0D7BWP7Higp7nlRl71BERZ-N2Kj7A+T4vm+7Ffj+NYMXWhpNi2wGgeBKi9gJtaPJESRDiOQA)


This is the type of `form.Subscribe` ([source](https://github.com/TanStack/form/blob/ab591c93afe7e536755e4be1abe1e8a1bd938f02/packages/react-form/src/useForm.tsx#L24-L27)):
```ts
    Subscribe: <TSelected = NoInfer<FormState<TFormData>>>(props: {
      selector?: (state: NoInfer<FormState<TFormData>>) => TSelected
      children: ((state: NoInfer<TSelected>) => ReactNode) | ReactNode
    }) => JSX.Element
  ```
  
In versions more recent than `5.0.4`, the type of the`TSelected` generic would be inferred from the return type of the `selector` method. If the returned value is `[state.canSubmit, state.isSubmitting]`, then `TSelected` will be `boolean[]`. However, in `5.0.4` that kind of inference is broken, and `TSelected` will be `unknown`, falling back to the default value, which is `NoInfer<FormState<TFormData>>`.

In the CI, the error we get for the highlighted code snippet complains about `boolean[]` not being compatible with `FormState<{ firstName: string; lastName: string; }`.
```
. test:types:versions50: src/tests/useField.test.tsx(585,17): error TS2322: Type '(state: FormState<{ firstName: string; lastName: string; }>) => boolean[]' is not assignable to type '(state: FormState<{ firstName: string; lastName: string; }>) => FormState<{ firstName: string; lastName: string; }>'.
. test:types:versions50:   Type 'boolean[]' is missing the following properties from type 'FormState<{ firstName: string; lastName: string; }>': isFormValidating, isFormValid, errors, errorMap, and 11 more.
```

## How should we solve this?

Fun fact, we can make `form.Subscribe` work in TS `5.0.4` like this:
```tsx
              <form.Subscribe
                 {...{
                  selector: (state) => [state.canSubmit, state.isSubmitting],
                  children: ([canSubmit, isSubmitting]) => (
                    <button type="submit" disabled={canSubmit}>
                      {isSubmitting ? '...' : 'Submit'}
                    </button>
                  ),
                }}
              />
```

It looks ugly, unconventional and doesn't solve our bit issue, so I wouldn't recommend it.

We could also use `//ts-ignore` here and explain in a comment why we do it, but might not prevent someone in the future running into this very hard-to-debug problem when they contribute to this repo, if they use `form.Subscribe`. Of course our users might run into this too.

My favorite solution is following what we did for `4.9.5`: ignore the test files for `5.0.4` in the CI (and obviously we should still mention this bug in the docs). I would also insert a comment at the [type definition of `form.Subscribe`](https://github.com/TanStack/form/blob/ab591c93afe7e536755e4be1abe1e8a1bd938f02/packages/react-form/src/useForm.tsx#L24-L27) to warn future collaborators (and our future selves) about this problem.

Wdyt?

crutchcorn: I like your outlined solution of ignoring the test files in 5.0.4, writing a comment on the type definition, and mentioning in the TypeScript docs page.

Any chance you want to tackle this?

fulopkovacs: Yeah, sure! I'll make a PR for it.

packages/form-core/src/FormApi.ts
```diff
@@ -340,17 +337,16 @@ export class FormApi<
...
-          if (!instance.state.meta.isTouched) {
-            // Mark them as touched
-            instance.setMeta((prev) => ({ ...prev, isTouched: true }))
-          }
-        })
+        if (!field.instance) return
+        // Validate the field
+        fieldValidationPromises.push(
+          Promise.resolve().then(() => field.instance!.validate(cause)),
+        )
```
fulopkovacs, line 344: We can get around the non-null assertion in `field.instance!` like this:

```suggestion
        if (!field.instance) return
        const fieldInstance = field.instance
        // Validate the field
        fieldValidationPromises.push(
          Promise.resolve().then(() => fieldInstance.validate(cause)),
        )
```

fulopkovacs (changes requested):
I left one change suggestion that eliminates a non-null assertion, otherwise LGTM 🚀!

This PR probably fixed many unborn bugs by removing `useIsomorphicEffectOnce.ts`. Great move before the `v1` release! 🙌

PR #588, opened 2024-01-28
Array and provider refactor

crutchcorn:
This PR aims to fix issues with array usage typings and removes the Provider component.

With this change, an array-based form in React might look like:

```tsx
import * as React from "react";
import { createRoot } from "react-dom/client";
import { useForm } from "@tanstack/react-form";

export default function App() {
  const form = useForm({
    defaultValues: {
      people: [] as Array<{ age: number; name: string }>,
    },
    onSubmit: async ({ value }) => {
      // Do something with form data
      alert(JSON.stringify(value, null, 2));
    },
  });

  return (
    <div>
      <h1>Simple Form Example</h1>
      <form
        onSubmit={(e) => {
          e.preventDefault();
          e.stopPropagation();
          void form.handleSubmit();
        }}
      >
        <form.Field name="people">
          {(field) => {
            return (
              <div>
                {field.state.value.map((_, i) => {
                  return (
                    <form.Field name={`people[${i}].name`}>
                      {(subField) => {
                        return (
                          <div>
                            <label>
                              <div>Name for person {i}</div>
                              <input
                                value={subField.state.value}
                                onChange={(e) =>
                                  subField.handleChange(e.target.value)
                                }
                              />
                            </label>
                          </div>
                        );
                      }}
                    </form.Field>
                  );
                })}
                <button
                  onClick={() => field.pushValue({ name: "", age: 0 })}
                  type="button"
                >
                  Add person
                </button>
              </div>
            );
          }}
        </form.Field>
        <form.Subscribe
          selector={(state) => [state.canSubmit, state.isSubmitting]}
          children={([canSubmit, isSubmitting]) => (
            <button type="submit" disabled={!canSubmit}>
              {isSubmitting ? "..." : "Submit"}
            </button>
          )}
        />
      </form>
    </div>
  );
}

const rootElement = document.getElementById("root")!;

createRoot(rootElement).render(<App />);
```

# TODO List

- [x] Fix tests for core path helpers
- [x] Fix tests for array usage in Core
- [x] Fix types in React
- [x] Fix types in Solid
- [x] Fix types in Vue
- [x] Write array tests for React
- [x] Write array tests for Solid
- [x] Write array tests for Vue
- [x] Write array example for React
    - [x] Update docs example as well
- [x] Write array example for Solid
    - [x] Mention `For` vs `Index` looping API
- [x] Write array example for Vue
- [x] Remove provider for React
    - [x] Remove from docs
    - [x] Remove from examples
- [x] Remove provider for Solid
[...]

PR #607, opened 2024-02-25
Remove SubField API

crutchcorn:
This PR removes the subfield API from `FieldAPI`, as:

1) We're removing the implicit ability to create a new FieldAPI from a parent's name in `react-core` (#588)
2) This API adds a ton of complexity (especially around typings)
3) It's just as easy to `new FieldAPI` with a new name


fulopkovacs (approved):
🚢 lgtm!

PR #615, opened 2024-03-01
docs: `touchedErrors` is array

melonges:
Array in condition will always evaluate to true
have to check the length and render the error separately


crutchcorn:
There's a few too many changes required to this code to approve. We'll fix the issue with the docs soon enough :)

melonges:
Ok, then add just `lenght`

PR #619, opened 2024-03-06
Pristine and dirty field states

fulopkovacs:
# Changes

This PR introduces 2 new flags for `form.Field`:

| flag | meaning |
| -- | -- |
| `isPristine` | The input field's value has not been modified yet |
| `isDirty` | The input field's value has been modified |

field-states

# TODO

- [x] `isPristine` flag
- [x] `isDirty` flag
- [x] tests
- [x] update the docs with info about the new flags

# Prior discussions about the `isDirty` flag
- #598 
- [check this discussion to see why we have a different concept of `dirty` from React Hook Form](https://github.com/TanStack/form/pull/598#issuecomment-1965519587)


docs/guides/basic-concepts.md
```diff
@@ -66,6 +66,19 @@ Example:
 const { value, error, touched, isValidating } = field.state
 ```
 
+There are three field states can be very useful to see how the user interacts with a field. A field is _"touched"_ when the user clicks/tabs into it, _"pristine"_ until the user changes value in it, and _"dirty"_ after the value has been changed. You can check these states via the `isTouched`, `isPristine` and `isDirty` flags, as seen below.
+
+```tsx
+const { isTouched, isPristine, isDirty } = field.state.meta
+```
+
+![Field states](https://github.com/TanStack/form/assets/43729152/513c29d4-23bc-4934-9658-9a5856ad2ef6)
```
crutchcorn, line 75: Let's upload this image inside of `assets` on the repo or something. That way, GH can't accidentally GC it and we can keep track of it locally

packages/form-core/src/FormApi.ts
```diff
@@ -622,6 +631,8 @@ export class FormApi<
         this.setFieldMeta(field, (prev) => ({
           ...prev,
           isTouched: true,
+          isPristine: false,
```
crutchcorn, line 634: We don't need this, do we? This is a derived field, right?

packages/form-core/src/FormApi.ts
```diff
@@ -207,6 +211,9 @@ export class FormApi<
 
           const isTouched = fieldMetaValues.some((field) => field?.isTouched)
 
+          const isDirty = fieldMetaValues.some((field) => field?.isDirty)
+          const isPristine = !isDirty
```
crutchcorn, line 215: Can we do this same trick inside of `FieldApi.ts` to make `isPristine` a non-actual meta field? (actually asking, IDK off the top)

crutchcorn (changes requested):

docs/guides/basic-concepts.md
fulopkovacs: Ugh, that'd be annoying! 😆  I uploaded it to `/docs/assets/field-states.png`, and updated the link.

packages/form-core/src/FormApi.ts
fulopkovacs: Ok, I think now we should be using it: see [`packages/form-core/src/FieldApi` line 325](https://github.com/TanStack/form/blob/314db2335ddda223d48e81a7865363cb652b8473/packages/form-core/src/FieldApi.ts#L325).

crutchcorn (approved):
Amazing work!

PR #626, opened 2024-03-16
Fix nested types

### LINKED ISSUE #625: Wrong type in nested array
### Describe the bug

In the arrays example, we have a form where the array is at the root of the object.
If I add a nesting level, the handleChange function cannot infer the correct type.

Example:
```

type FormDefinition = {
  nested: {
    people: {
      name: string;
      age: number;
    }[];
  };
};
```

On `subField.handleChange(e.target.value)` I get the following Typescript error:

```
Argument of type 'string' is not assignable to parameter of type 'UpdaterFn<never, never>'.(2345)
(property) React.ChangeEvent<HTMLInputElement>.target: EventTarget & HTMLInputElement
```

(see stackblitz)

https://stackblitz.com/edit/tanstack-form-kar7dz?file=src%2Findex.tsx

### Steps to reproduce

1. Have a form with an array inside an object

### Expected behavior

Typescript gets the right type on `handleChange`, in this case `(updater: Updater<string>) => void`

None

### Screenshots or Videos

_No response_

### Platform

macOS but also on stackblitz

### TanStack Form adapter

react-form

### TanStack Form version

v0.14.0

### TypeScript version

v5.4.2

### Additional context

An extra note, in an example I'm testing locally the component doesn't render at all and I get this error:
image

However I was not able to reproduce this on stackblitz so I'm only adding this as additional context.

crutchcorn:
This PR closes #625 by:
- Adding in edgecases for arrays in the 1`DeepValues` type
- Add a `TDepth` to `DeepValues` to prevent TypeScript from overflowing memory
- Fixing issues with FW adapters not passing `TData` properly
- Fixing an issue where `defaultValue` would infer the data type incorrectly
- Fixing weird `plugin` typing issues with Vite by sidestepping with `as never` ( CC @lachlancollins )
- Adding tests to validate typing fixes
- Adds TS 5.3 and 5.4 to the TSC options, update `tsc` to run in 5.4 to solve a bug in TS 5.2

Apologies for the big PR, this isn't easily broken up 😓


packages/form-core/src/util-types.ts
```diff
@@ -83,17 +85,40 @@ type PrefixFromDepth<
...
-    ? DeepValue<TValue[TBefore][TBrackets], TAfter>
-    : TAccessor extends `[${infer TBrackets}]`
-      ? DeepValue<TValue, TBrackets>
-      : TAccessor extends `${infer TBefore}[${infer TBrackets}]`
-        ? DeepValue<TValue[TBefore], TBrackets>
-        : TAccessor extends `${infer TBefore}.${infer TAfter}`
-          ? DeepValue<TValue[TBefore], TAfter>
-          : TValue[TAccessor & string]
-  : never
+export type DeepValue<
```
fulopkovacs, line 88: I don't know if it's an issue we can solve, but this type is pretty complex and hard to understand (for me)... 😅 

TS in libs can get pretty crazy, I get that, but I think `DeepValue` has the potential to make it very hard to add new features or fix bugs in the future. And `DeepValue` is a pretty foundational part of the code base.

_How could we address this issue?_ It's not entirely fixable, but maybe we could add comments to make the intent clearer in some places. E.g.:

```ts
// Limit the depth of recursion to 3 levels
TDepth['length'] extends 3
  ? never
  : TValue extends ReadonlyArray<any> ? ...
```

It'd be also good to document the type in some way, like:
```ts
/**
   Infer the type of a deeply nested property within an object or array.
   See the tests in `util-types.test-d.ts` for usage.
*/
export type DeepValue<TValue, TAccessor, TDepth extends any[] = []>
```

In general, I'm not a fan of commenting code, because comments tend to get outdated rather quickly, but in this case, I think they would help us immensely, and we do proper code reviews anyways.

Here's my proposal: if you're okay with adding comments, I would be happy to do this work in a suggestion or commit to this PR, and then we could discuss them.

There's also the possibility of making smaller types for the different logical groups in the type (e.g.: one branch of a conditional infers the property from a `Record` (could be `DeepValueInRecord`??), the other one from an `Array` (`DeepValueInArray`??), and things like that). I feel like this approach could overcomplicate things, so I'd avoid it (this is just a gut feeling, so maybe I'm wrong).

Wdyt?

fulopkovacs (commented):
I concluded the _Part 1_ of my review (will continue it tomorrow). I'm done with 22/30 files.

packages/form-core/src/util-types.ts
crutchcorn: Mentioned this privately in our maintainers channel but want to confirm publicly; **Comments are always welcome in the codebase!**

In fact, I'm glad you're advocating for this; it'll a great change for our users and future collaborators.

packages/form-core/src/util-types.ts
```diff
@@ -83,17 +85,40 @@ type PrefixFromDepth<
...
+      ? DeepValue<
+          DeepValue<TValue, TBrackets, [...TDepth, any]>,
+          TAfter,
+          [...TDepth, any]
+        >
+      : TAccessor extends `[${infer TBrackets}]`
+        ? DeepValue<TValue, TBrackets, [...TDepth, any]>
+        : TAccessor extends keyof TValue
+          ? TValue[TAccessor]
+          : TValue[TAccessor & number]
```
fulopkovacs, line 105: Could you explain to me why we need `TAccessor &` here:
```suggestion
          : TValue[number]
```

I tried to come up with an answer by myself, but I couldn't. The inferred type here should be the type of the elements in the array, no? 🤔


I ran the tests, with my change, and some of them failed ([source](https://cloud.nx.app/runs/FYlyIB1Kap)):
image

I'm a bit lost 😅

crutchcorn: Funny enough it's more the `& number` that we don't (theoretically) need. TAccessor here is referring to the exact number being passed, while `& number` is type casting it (sorta) to a number so that TS doesn't complain about non numerical access of an array

packages/form-core/src/util-types.ts
```diff
@@ -83,17 +85,40 @@ type PrefixFromDepth<
...
+          : TAccessor extends `${infer TBefore}.${infer TAfter}`
+            ? DeepValue<
+                DeepValue<TValue, TBefore, [...TDepth, any]>,
+                TAfter,
+                [...TDepth, any]
+              >
+            : TAccessor extends string
+              ? TValue[TAccessor]
+              : never
+      : never
```
fulopkovacs, line 124: Okay, these are the comments I'd add to this type. 

```suggestion
/**
   Infer the type of a deeply nested property within an object or an array.
   See the tests in `util-types.test-d.ts` for usage.
*/
export type DeepValue<
  // The object or array in which we have the property whose type we're trying to infer
  TValue,
  // A string representing the path of the property we're trying to access
  TAccessor,
  // (Optional) An array representing the current depth of recursion
  TDepth extends any[] = [],
> =
  // Limit the depth of recursion to 3 levels
  TDepth['length'] extends 3
    ? never
    : // Check if we're looking for the property in an array
      TValue extends ReadonlyArray<any>
      ? TAccessor extends `[${infer TBrackets}].${infer TAfter}`
        ? /*
          Extract the first element from the accessor path (`TBrackets`)
          and recursively call `DeepValue` with it
          */
          DeepValue<
            DeepValue<TValue, TBrackets, [...TDepth, any]>,
            TAfter,
            [...TDepth, any]
          >
        : TAccessor extends `[${infer TBrackets}]`
          ? DeepValue<TValue, TBrackets, [...TDepth, any]>
          : TAccessor extends keyof TValue
            ? TValue[TAccessor]
            : TValue[TAccessor & number]
      : // Check if we're looking for the property in an object
        TValue extends Record<string | number, any>
        ? TAccessor extends `${infer TBefore}[${infer TEverythingElse}`
          ? DeepValue<
              DeepValue<TValue, TBefore, [...TDepth, any]>,
              `[${TEverythingElse}`,
              [...TDepth, any]
            >
          : TAccessor extends `[${infer TBrackets}]`
            ? DeepValue<TValue, TBrackets, [...TDepth, any]>
            : TAccessor extends `${infer TBefore}.${infer TAfter}`
              ? DeepValue<
                  DeepValue<TValue, TBefore, [...TDepth, any]>,
                  TAfter,
                  [...TDepth, any]
                >
              : TAccessor extends string
                ? TValue[TAccessor]
                : never
        : // Do not allow `TValue` to be anything else
          never

```

crutchcorn: Love what I'm seeing. Let's move this to a nested PR (a PR onto this PR) so we can discuss more thoroughly

fulopkovacs: Awesome! Here you go: https://github.com/TanStack/form/pull/634

packages/react-form/src/tests/useField.test-d.tsx
```diff
@@ -65,3 +65,44 @@ it('should type onChange properly', () => {
...
+    }
+  }
+
+  function App() {
+    const form = useForm({
+      defaultValues: {
+        nested: {
+          people: [],
+        },
+      } as FormDefinition,
```
fulopkovacs, line 85: Just a suggestion, and maybe not a very important one, but we could use `satisfied`  if we added an element to the `people` array in `defaultValues`.  This way we'd get rid of the type casting.

```suggestion
      defaultValues: {
        nested: {
          people: [{ name: 'Tom', age: 50 }],
        },
      } satisfies FormDefinition,
```

fulopkovacs (approved):
Added 1 small suggestion, but this PR could already be merged in its current form. Massive effort! 🙌 🚀

PR #631, opened 2024-03-19
fix(solid-form): Add `solid` export condition

Brendonovich:
Solid relies on the solid export condition to perform a custom JSX transform depending on whether the code is being built for client or server, but `@tanstack/solid-form` only has `dist/cjs` and `dist/esm`, which are client-only since they import stuff like `memo` from `solid-js/web` (which is only available in client builds).

To fix this I've added `tsc` to the build of process of `@tanstack/solid-form` so that `.js/.jsx` files are emitted which solid can apply its custom JSX transform on. There's probably nicer ways of doing this that can integrate with vite but this is what I did to fix [a similar problem](https://github.com/solidjs/solid-start/issues/1228) for SolidStart.


crutchcorn:
@lachlancollins can you investigate this? We'd want this to integrate with Vite

Also CC @ryansolid if you have any guidance on how to do this (sorry for the ping Ryan)

ryansolid:
This is pretty much correct. We want the source, but not necessarily the TypeScript so we tend to tsc it into dist folder. My only recommendation is build it into a `.jsx` entry file if there is any JSX anywhere in the solid implementation of the library. That way it will get processed properly.

Brendonovich:
> My only recommendation is build it into a .jsx entry file if there is any JSX anywhere in the solid implementation of the library. That way it will get processed properly.

This is probably true but having the entrypoint be `.js` seems to be working fine for us

packages/solid-form/package.json
```diff
@@ -22,7 +22,7 @@
     "test:lib": "vitest",
     "test:lib:dev": "pnpm run test:lib --watch",
     "test:build": "publint --strict",
-    "build": "vite build"
+    "build": "vite build && tsc -p tsconfig.build.json"
```
crutchcorn, line 25: Leaving a comment here to remind @lachlancollins to investigate moving this into our Vite build if at all possible

crutchcorn (changes requested):

packages/solid-form/package.json
lachlancollins: @crutchcorn I've tweaked the config slightly so that it generates a flatter structure. AFAIK, there is no need for the `dist/source` folder to have its own types, since they are identical to those in `dist/esm` (but if not, it can be re-enabled by removing `"declaration": false`).

I agree it might be nice to have a direct TS->JS copy of the source built by `@tanstack/config`, I'll have a look!

TheKnightCoder:
Is there anything blocking this from being merged? tanstack forms is unusable on solid without this

crutchcorn:
@TheKnightCoder I'd like to see our build tooling unified to solve this problem before we merge. @lachlancollins any chance you could take another look into merging into Vite?

Alternatively, if you think this fine to publish as-is for now and refine later, we can

crutchcorn:
Merging, per some research from @lachlancollins showing this _should_ work now. Let us know if you run into additional problems around ESM/CJS support

PR #632, opened 2024-03-19
Updated Form Comparison Chart w/ Formik Info

anthonylevine:
Source:

Nested object/array fields: https://formik.org/docs/examples/field-arrays & https://formik.org/docs/guides/arrays

Async validation: https://formik.org/docs/api/formik#validate-values-values--formikerrorsvalues--promiseany

Schema-based Validation: https://formik.org/docs/guides/validation


crutchcorn:
Nice, this is awesome! Thank you!

We could actually use a fair bit of help with this page if you'd like to keep investigating :)

PR #646, opened 2024-03-24
Update RHF isDirty note

kennethvdberghe:
I think the author meant it the other way around.


fulopkovacs (approved):
Nice catch! Thanks for the contribution! 🚀

Issue #645, opened 2024-03-24
Angular form does not allow standard html attributes for boolean and number inputs


jyjor:
### Describe the bug

Boolean and number attributes in Angular do not follow typical HTML conventions. Here are some examples:
`<div myInput>` - In Angular, `myInput` is `''`, but in standard HTML it is the boolean value `true`.
`<div myInput="true">` - In Angular, `myInput` is the string `'true'`, but in standard HTML it is the boolean value `true`.

Angular has transform functions to address these discrepancies and it looks like the number and boolean inputs are missing these. This is a simple problem with a simple fix, and I would love put up a PR to fix it.

Side note - I spoke with Corbin at ng-conf during a lunch break, was very impressed with him from our short conversation, and he inspired me to take a look. He also said he won't respond with " :frowning_face: " and close the issue, even though it's small.

### Your minimal, reproducible example

https://stackblitz.com/edit/stackblitz-starters-pszun1?file=src%2Fmain.ts

### Steps to reproduce

1. Open the linked example
2. Note the error on line 19 of main.ts
> NG2: Type '""' is not assignable to type 'boolean | undefined'.
3. Change the line to `asyncAlways="true"` to get the same error
4. Change the line to `[asyncAlways]="true"` and it will be fixed.
5. In addition to `asyncAlways`, this also applies to`asyncDebounceMs` and `preserveValue` attributes.

### Expected behavior

`asyncAlways` and `asyncAlways="true"` should be equivalent to `[asyncAlways]="true"` and `asyncAlways="false"` should be equivalent to `[asyncAlways]="false"`.

### How often does this bug happen?

Every time

### Platform

- OS: Linux
- Browser: Chrome
- Version 0.17.0

### TanStack Form adapter

None

### TanStack Form version

0.17.0

### TypeScript version

5.4.2

### Additional context

In the "TanStack Form adapter" drop down above, there is no option for "angular-form", but that is the version with the error.

crutchcorn:
If you're up for it, I'd accept a PR adding the transforms 😊 Let's just make sure it doesn't break the TS type inferencing

PR #648, opened 2024-03-24
fix: transform valid values for number and boolean HTML attributes

jyjor:
Fix: #645

This is the quickest, closest-to-how-Angular-does-it fix. It does allow _any_ value as an input, however:
| Input value | Previous result | New result |
|---|---|---|
| **Boolean transform** | | |
| `[asyncAlways]="true"` | `true` | `true` |
| `[asyncAlways]="false"` | `false` | `false` |
| `asyncAlways` | Error (should be `true`) | `true` |
| `asyncAlways="true"` | Error (should be `true`) | `true` |
| `asyncAlways="false"` | Error (should be `false`) | `false` |
| `asyncAlways="fasle"` | Error | `true` (should be an error) |
| **Number transform** |||
| `[asyncDebounceMs]="500"` | `500` | `500` |
| `asyncDebounceMs="500"` | Error (should be `500`) | `500` |
| `asyncDebounceMs` | Error | `NaN` (should be an error) |
| `asyncDebounceMs="123abc"` | Error | `NaN` (should be an error) |

Alternatively, we could wrap the transform functions and have them only accept valid values. This would fix everything.
```ts
function strictBooleanAttribute(
  value: 'true' | 'false' | '' | boolean | null | undefined,
): boolean {
  return booleanAttribute(value)
}

function strictNumberAttribute(value: `${number}` | number | null | undefined) {
  return numberAttribute(value)
}
```

This would fix all of the above problems
| Input value | Previous result | New result |
|---|---|---|
| **Boolean transform** | | |
| `[asyncAlways]="true"` | `true` | `true` |
| `[asyncAlways]="false"` | `false` | `false` |
| `asyncAlways` | Error (should be `true`) | `true` |
| `asyncAlways="true"` | Error (should be `true`) | `true` |
| `asyncAlways="false"` | Error (should be `false`) | `false` |
| `asyncAlways="fasle"` | Error | Error |
| **Number transform** |||
| `[asyncDebounceMs]="500"` | `500` | `500` |
| `asyncDebounceMs="500"` | Error (should be `500`) | `500` |
| `asyncDebounceMs` | Error | Error |
| `asyncDebounceMs="123abc"` | Error | Error |

### DIFF
--- a/packages/angular-form/src/tanstack-field.directive.ts
+++ b/packages/angular-form/src/tanstack-field.directive.ts
@@ -4,6 +4,8 @@ import {
   type OnChanges,
   type OnDestroy,
   type OnInit,
+  booleanAttribute,
+  numberAttribute,
 } from '@angular/core'
 import {
   type DeepKeys,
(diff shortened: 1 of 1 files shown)


crutchcorn (approved):
LGTM! Thanks for this!

Issue #653, opened 2024-03-25
onChangeAsyncDebounceMs shouldn't be called when calling form.submit


francois-pasquier:
### Describe the bug

With the current behaviour, `onChangeAsyncDebounceMs` is waited when calling `form.submit`

### Your minimal, reproducible example

See PR with the test

### Steps to reproduce

See PR with the test

### Expected behavior

It shouldn't wait for this onChange debounce.

### How often does this bug happen?

None

### Platform

React

### TanStack Form adapter

None

### TanStack Form version

latest

PR #654, opened 2024-03-25
fix: remove onChangeAsyncDebounceMs waiting for submit (#653)

francois-pasquier:
Fixes https://github.com/TanStack/form/issues/653

### DIFF
--- a/packages/form-core/src/utils.ts
+++ b/packages/form-core/src/utils.ts
@@ -197,6 +197,12 @@ export function getAsyncValidatorArray<T>(
     debounceMs: onChangeAsyncDebounceMs ?? defaultDebounceMs,
   } as const
 
+  const changeValidatorForSubmit = {
+    cause: 'change',
+    validate: onChangeAsync,
+    debounceMs: defaultDebounceMs,
+  } as const
+
   const blurValidator = {
     cause: 'blur',
     validate: onBlurAsync,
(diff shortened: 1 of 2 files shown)


crutchcorn (changes requested):
This change would need to be made for all of the validators:

- onChange
- onBlur
- onSubmit

But change `debounceMs` to `0`

francois-pasquier:
Ah your are right! I removed the debounce time as the submit validator will still have the user defined or default one, did the same for onBlur and updated the test. I kept the submit validator as it is because we still need the debounce on one validator

crutchcorn:
>  I kept the submit validator as it is because we still need the debounce on one validator

I'm unclear on this. Why do we still need to debounce one validator? What breaks if we don't? This feels unexpected to me

francois-pasquier:
> > I kept the submit validator as it is because we still need the debounce on one validator
> 
> I'm unclear on this. Why do we still need to debounce one validator? What breaks if we don't? This feels unexpected to me

I am not sure I follow.

I kept the submit validator as it is because this prop `onSubmitAsyncDebounceMs` is available for the user.
So if we get rid of the debounce in the submit validator there is no point in having `onSubmitAsyncDebounceMs`

Maybe I am not understanding what you mean though.

crutchcorn:
Oh... Hmmm.. That's a really good point. TBH I'm actually thinking that we remove the `onSubmitAsyncDebounceMs` and ability to debounce submit async validators entirely at the moment.

What do you think?

crutchcorn:
Tests seem to be failing here FWIW

crutchcorn:
Replaced by https://github.com/TanStack/form/pull/687 using your implementation. Thank you for your help! :D

PR #656, opened 2024-03-27
feat: set field errors from the form validators

fulopkovacs:
This PR adds the ability to set field errors from the form validators.

The API would look like this (this is a value returned by a validator):
```ts
{
  form: "General form error", // optional,
  fields: {
    username: "Username taken",
    password: "Password is too short"
  }
}
```

We'll add support for Zod, Yup, and Valibot in a separate PR.

# TODO

- [x] Initial implementation of the feature
- [x] Add tests for
  - [x] basic fields
  - [x] array fields
  - [x] nested fields in array fields
  - [x] linked fields
- [x] Clean up the code 
    - [x] `form-core`
    - [x] tests
- [x] Docs
- [x] Rebase to `main`
- [x] Review the code after the rebase
- [x] Update the docs after the rebase


packages/form-core/src/FormApi.ts
```diff
@@ -41,7 +58,15 @@ export type FormValidateAsyncFn<
   value: TFormData
   formApi: FormApi<TFormData, TFormValidator>
   signal: AbortSignal
-}) => ValidationError | Promise<ValidationError>
+}) => FormValidationError<TFormData> | Promise<FormValidationError<TFormData>>
+
+export type FormValidator<TFormData, Type, Fn = unknown> = {
```
crutchcorn, line 63: Small tip that @tannerlinsley shared with me: Prefix all generics with `T` so you don't lose track of what's a generic and what's not.

`Type` would become `TType` and `Fn` -> `TFn`

packages/form-core/src/FormApi.ts
```diff
@@ -731,16 +790,25 @@ export class FormApi<
...
+function normalizeError<TFormData>(rawError?: FormValidationError<TFormData>): {
+  formError: ValidationError
+  fieldErrors?: Partial<Partial<Record<DeepKeys<TFormData>, ValidationError>>>
+} {
   if (rawError) {
+    if (typeof rawError === 'object' && rawError.form) {
+      const formError = normalizeError(rawError.form).formError
+      const fieldErrors = rawError.fields
+      return { formError, fieldErrors }
+    }
```
crutchcorn, line 802: This is a really clever way of doing this! I like it

denisorehovsky:
Hi. When are you planning to release it? Thanks.

fulopkovacs:
> Hi. When are you planning to release it? Thanks.

I'm planning to pick this up during the weekend. 😅 Unfortunately there are some urgent things that I have to take care of in the `tanstack/query` repo (a lot of broken links in the docs), but I'll jump on this PR after that.

denisorehovsky:
@fulopkovacs Perfect. Thank you. This is indeed necessary. I migrated from react-hook-form, which had a setError function that allowed setting errors when the backend responded with errors. It was surprising that this library doesn't yet support something similar. I think it's a very common use case to have backend validation and then set field-level validation errors.

fulopkovacs:
Sad status update: I ran into some issues as I started adding more complex tests. This is a bummer because I wanted to get this one ready during the weekend. 😢

denisorehovsky:
@fulopkovacs It's ok :)
Will be waiting for the update

denisorehovsky:
@fulopkovacs Hey :) You're probably busy, so I'm sorry to bother you, but do you know when you will be able to get back to this PR? Without this solution, I will likely need to find a workaround for properly handling backend errors, or switch back to react-hook-form

crutchcorn:
Hey @denisorehovsky, just as a heads up - we don't follow any kind of release schedule or cadence. As we're all volunteer based, we're only able to take on tasks as we have time. Maybe you could consider sponsoring the project financially or making a PR if you're needing priority support? 😄

Moreover, though, I'm not sure I understand how this is a blocker for anything. This PR is a nicety, not a required feature to merge form and field errors. You can always do that manually by using a helper component.

fulopkovacs:
> @fulopkovacs Hey :) You're probably busy, so I'm sorry to bother you, but do you know when you will be able to get back to this PR? Without this solution, I will likely need to find a workaround for properly handling backend errors, or switch back to react-hook-form

Hey ☺️,

I'm actively working on this PR! The process is slow, because there are many scenarios to handle. All the `form-core` tests are passing for the sync validators, and locally I am getting closer to solving the problem of the async validators too, but unexpected things keep popping up (right now I'm solving some issues for the linked fields).

After all the `form-core` tests pass, I'm planning to add some library tests (React), to make sure that everything is OK.

Then, the PR will go to review, but unfortunately, I don't expect that to be very quick (it's gonna be a beefy PR), and there's always the chance that the approach I chose doesn't fit the project well enough. 😬

In short, I'm sorry, but this will likely take some time. (Honestly, it's more complex than I expected it to be, I thought I'd be ready by the end of last week.) 

**Finding a workaround is a more realistic solution** than waiting for this feature to be merged. For example, you can handle the form's validation on the server with the form's `onSubmitAsync` validator, and save the field-related errors to your component's local state before you return an error (that you don't have to display btw) for the form. There are also other ways to solve this, I hope you find one that fits your use case.

In situations like this, I'm really sad that it's not my full-time job, it'd be awesome to respond to the users' needs more quickly! 😢 Thanks for your understanding though.

packages/form-core/src/FormApi.ts
```diff
@@ -455,7 +517,13 @@ export class FormApi<
...
 
       promises.push(
-        new Promise<ValidationError | undefined>(async (resolve) => {
+        new Promise<
+          | {
+              fieldErrors: Partial<Record<DeepKeys<TFormData>, ValidationError>>
+              errorMapKey: ValidationErrorMapKeys
+            }
+          | undefined
+        >(async (resolve) => {
```
meszarosdezso, line 526: Could use `ValidationPromiseResult` here.

fulopkovacs: Nice catch, thanks! ☺️ Fixed it!

MikeBurkeMed:
Hi all, this feature would be amazing and is very critical, I'll be happy to help but I cannot see what's blocking its merge.

fulopkovacs:
> Hi all, this feature would be amazing and is very critical, I'll be happy to help but I cannot see what's blocking its merge. 

Hey 👋! It's waiting for a rebase (by me) before others in the team can review it. Unfortunately too many things have been changed, so it doesn't make sense to change the order of these things.

I was away from the keyboard for 2 weeks, and I won't be able to work on this during the workweek, but I'll look into it on the weekend! 

(Can't promise anything though. A lot of PRs have been merged since I've finished this one. I hope it'll go smooth! 🤞)

MikeBurkeMed:
> I was away from the keyboard for 2 weeks, and I won't be able to work on this during the workweek, but I'll look into it on the weekend!
> 
> (Can't promise anything though. A lot of PRs have been merged since I've finished this one. I hope it'll go smooth! 🤞)

Hey! That's excellent, thank you for all your work! Please let me know if I can be of any help.

fulopkovacs:
Here's quick report on the progress I made during the weekend:

- there were a lot of conflicts during the rebase (basically each commit had at least 1 conflict)
- I decided to create a [new branch](https://github.com/fulopkovacs/form/tree/field-errors-from-form-validators-rebase) for myself where all the commits of this branch is squashed into a single one -> I'll work on this new branch for now, and will probably force push the commits from the new branch to this one once the issues caused by the rebase are resolved
- I'm done resolving the merge conflicts on the new branch, but some tests are failing, so I'll have to fix them (I'll do it next week)
  - only 3 tests failed in `form/core`, which is a good sign
  - here are the error logs: https://cloud.nx.app/runs/GzTGGOR3Md
  
image

packages/form-core/tests/FieldApi.spec.ts
```diff
@@ -1105,11 +1106,10 @@ describe('field api', () => {
     field.mount()
 
     field.setValue('one')
-    // Allow for a micro-tick to allow the promise to resolve
-    await sleep(1)
+    await vi.runAllTimersAsync()
```
fulopkovacs, line 1109: After the changes in the validators, the number of micro-ticks needed also increased in some places. Instead of calculating the exact number of micro-ticks we need at each place, I decided to use fake timers and `vi.runAllTimersAsync`.

examples/react/field-errors-from-form-validators/src/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+      age: 0,
+    },
+    validators: {
+      onSubmitAsync: async ({ value }) => {
+        console.log({ value })
+        // Verify the age on the server
+        const isNotOlderThan13 = !(await verifyAgeOnServer(value.age))
+
+        // Verify the availability of the username on the server
+        const isUsernameTaken = await checkIfUsernameIsTaken(value.username)
```
Balastrong, line 27: Personal preference: I'd keep both these positive here: `verifyAgeOnServer` is returning true if the value is ok, `checkIfUsernameIsTaken` is returning true if the value is *not* ok.

Let's make them consistent, an example:
```ts
async function verifyAgeOnServer(age: number) {
  return age < 13
}
...
const isTooYoung = await verifyAgeOnServer(value.age)
```

examples/react/field-errors-from-form-validators/src/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+          )}
+        />
+
+        {form.state.errorMap.onSubmit ? (
+          <div>
+            <em>
+              There was an error on the form: {form.state.errorMap.onSubmit}
+            </em>
+          </div>
+        ) : null}
```
Balastrong, line 108: This has to be in `form.Subscribe` or read from `useStore` to get the latest values

docs/framework/react/guides/validation.md
```diff
@@ -194,6 +194,75 @@ export default function App() {
...
+            </>
+          )}
+        </form.Field>
+        {form.state.errorMap.onSubmit ? (
+          <div>
+            <em>
+              There was an error on the form: {form.state.errorMap.onSubmit}
+            </em>
+          </div>
+        ) : null}
```
Balastrong, line 257: Same as in the example, needs form.Subscribe or useStore

Balastrong (commented):
Just a couple of comments in the docs, looks great!

examples/react/field-errors-from-form-validators/src/index.tsx
fulopkovacs: Great points, I fixed it!

examples/react/field-errors-from-form-validators/src/index.tsx
fulopkovacs: 😅

Balastrong (approved):
LGTM!

fulopkovacs:
Current status of this PR: we're gonna have one more review by @crutchcorn. (I introduced some serious changes in the validation, so the more people see it before the merge the better.)

crutchcorn:
Just thinking out loud here:

image

This introduces a new behavior that many might consider a "bug". This behavior would occur when the following pattern emerges in a codebase:

```jsx
  const form = useForm({
    defaultValues: {
      age: 0,
    },
    validators: {
      onChange: ({ value }) => {
        return {
          fields: {
            age: value.age < 12 ? 'Too young!' : undefined,
          },
        }
      },
    },
  })

  return (
    <div>
      <h1>Field Errors From The Form's validators Example</h1>
      <form
        onSubmit={(e) => {
          e.preventDefault()
          e.stopPropagation()
          void form.handleSubmit()
        }}
      >
        <form.Field
          name="age"
          validators={{
            onChange: ({ value }) => value % 2 === 0 ? 'Must be odd!' : undefined,
          }}
          children={(field) => (
            <div>
              <label htmlFor={field.name}>Age:</label>
              <input
                id={field.name}
                name={field.name}
                value={field.state.value}
                type="number"
                onChange={(e) => field.handleChange(e.target.valueAsNumber)}
              />
              {field.state.meta.errors.length > 0 ? (
                <em role="alert">{field.state.meta.errors.join(', ')}</em>
              ) : null}
            </div>
          )}
        />
        <form.Subscribe
          selector={(state) => [state.canSubmit, state.isSubmitting]}
          children={([canSubmit, isSubmitting]) => (
            <button type="submit" disabled={!canSubmit}>
              {isSubmitting ? '...' : 'Submit'}
            </button>
          )}
        />
      </form>
    </div>
  )
}
```

Where the developer might want to get two different errors at the same time, but is only reported one of the errors due to how we have this edgecase handled in our code.

Now, to be clear, I don't think this is actually a huge problem - but:

1) It does play into the idea that for 2.x (depending on how things shake out) we might want to migrate things to an array for `errorMap.onChange`). [I'm still not fully convinced of this yet]

2) We should 100% document this behavior before it goes live to avoid confusion

----


Continuing my review otherwise ignoring this

crutchcorn (approved):
The code LGTM and has sane defaults/behaviors - you've done an **amazing** job with this PR. Prior to merging, let's update the docs to reflect what I was talking about with the edgecase.

I can tackle that tho

PR #658, opened 2024-03-27
fix: extract path from nested arrays

### LINKED ISSUE #657: Values in nested arrays do not get updated properly
### Describe the bug

When you have an array inside an array, values are not updated properly since the field is not treated as an array.

In the MRE the structure is as follows:
```
      people: [
        {
          names: ['Leonardo'],
        },
      ],
```

When editing the string "Leonardo" in the field, I'm expecting `setFieldValue` to properly compute the field as `people[0].names[0]` but instead it gets `people[0]["names[0]"]`

As a result, if I type the char `5` in the text field I do not see the value updated, and if I submit the form I see that the state is now at
```
      people: [
        {
          names: ['Leonardo'],
          names[0]: 'Leonardo5'
        },
      ],
```

https://stackblitz.com/edit/tanstack-form-9al5hb?file=package.json,src%2Findex.tsx

### Steps to reproduce

1. Try to edit the name in the text field, type a char
2. Hit submit
3. Notice you have both `names` array and a field called `names[0]`

### Expected behavior

In the example I would expect to see the value updated in the textfield and the state to be
```
      people: [
        {
          names: ['Leonardo5'],
        },
      ],
```

Every time

### Screenshots or Videos

_No response_

### Platform

MacOS

### TanStack Form adapter

None

### TanStack Form version

v0.18.0

### TypeScript version

_No response_

### Additional context

I tried to investigate and I think the root cause is on `makePathArray` not being able to handle multiple square brackets here
https://github.com/TanStack/form/blob/39848e3654e6b47acbf91e30858fb183493e5041/packages/form-core/src/utils.ts#L129-L136

Probably a replaceAll should be enough, I'll try to draft a PR and add a test

Balastrong:
Fixes #657 

I added a few tests and used it on a local project to check if it was working

### DIFF
--- a/packages/form-core/src/tests/utils.spec.ts
+++ b/packages/form-core/src/tests/utils.spec.ts
@@ -1,12 +1,12 @@
 import { describe, expect, it } from 'vitest'
-import { deleteBy, getBy, setBy } from '../utils'
+import { deleteBy, getBy, makePathArray, setBy } from '../utils'
 
 describe('getBy', () => {
   const structure = {
     name: 'Marc',
     kids: [
-      { name: 'Stephen', age: 10 },
-      { name: 'Taylor', age: 15 },
+      { name: 'Stephen', age: 10, hobbies: ['soccer', 'reading'] },
+      { name: 'Taylor', age: 15, hobbies: ['swimming', 'gaming'] },
     ],
     mother: {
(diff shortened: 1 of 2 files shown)


crutchcorn:
Thanks for this!

PR #673, opened 2024-04-13
update `react-hook-form` column in comparision table

AhmedBaset:
It's better to review this in [unified mode](https://github.com/TanStack/form/pull/673/files?diff=unified) with wide screen or scale down


tannerlinsley:
Do we actually know that hook form is fully inferred all the way down to nested and deep fields?

docs/comparison.md
```diff
@@ -17,15 +17,15 @@ Feature/Capability Key:
...
+| Framework agnostic                                | ✅                                            | ❓                             | ❓                                     | 🛑                                               | ❓                                     |
 | Granular reactivity                               | ✅                                            | ❓                             | ❓                                     | ❓                                               | ❓                                     |
-| Nested object/array fields                        | ✅                                            | ✅                             | ❓                                     | ❓                                               | ❓                                     |
+| Nested object/array fields                        | ✅                                            | ✅                             | ❓                                     | ✅                                               | ❓                                     |
```
A7med3bdulBaset, line 25: If I could put a tooltip here, I'd say while it suppors nested arrays but `useFeildArray` supports one level nesting

A7med3bdulBaset:
> Do we actually know that hook form is fully inferred all the way down to nested and deep fields?

Yeah. When you pass a generic it infer everything from it:
```tsx
const form = useForm<FormType>({
  // all props here infer FormType
});
//       ^? UseFormReturn<FormType>
```

all method and values of `form` are type safe

tannerlinsley:
Generic casting isn’t ideal. How does it do with *actual* inference?

docs/comparison.md
tannerlinsley: You can use a *(1) callout with a corresponding footnote

A7med3bdulBaset:
one trade of is with arrays:

let's say you have 
```ts
interface FormType {
  array: { key: string }[]
}
```

The way you access the `key` is by `"array.0.key"`. It won't suggest you this but it will error if you didn't follow the shape. In other words `array.whatever.key` will case a TS error. But <kbd>Ctrl</kbd>+<kbd>space</kbd> will not suggest `array.0.key`.

A7med3bdulBaset:
> Generic casting isn’t ideal. How does it do with _actual_ inference?

image

`form.trigger("c")` is TS error

docs/comparison.md
```diff
@@ -17,15 +17,15 @@ Feature/Capability Key:
...
-| Nested object/array fields                        | ✅                                            | ✅                             | ❓                                     | ❓                                               | ❓                                     |
-| Async validation                                  | ✅                                            | ✅                             | ❓                                     | ❓                                               | ❓                                     |
+| Nested object/array fields                        | ✅                                            | ✅                             | ❓                                     | ✅                                               | ❓                                     |
+| Async validation                                  | ✅                                            | ✅                             | ❓                                     | ✅                                               | ❓                                     |
```
A7med3bdulBaset, line 26:
```ts
const { name, onChange, onBlur, ref } = form.register("a", {
    validate: async (value, _formValues) => {
      await new Promise((resolve) => setTimeout(resolve, 1000));
      return value === "error" ? "error message" : undefined /* or true */;
    },
  });
```

docs/comparison.md
A7med3bdulBaset: I initially thought it was unsupported, but after rechecking the documentation, it appears to be a TypeScript issue.

> for nested field array, you will have to cast the field array by its name.
>```ts
>const { fields } = useFieldArray({ name: `test.${index}.keyValue` as 'test.0.keyValue' });
>```
https://react-hook-form.com/docs/usefieldarray

crutchcorn: That's the `*` I'm hoping for - type casting is an obtuse way of supporting this that we don't require

A7med3bdulBaset: See [`99a07ff` (#673)](https://github.com/TanStack/form/pull/673/commits/99a07fff483cfdfcb8e5d223faf4a8e5393a6999)

crutchcorn:
Thanks for this!

Issue #671, opened 2024-04-12
TanStack/Angular - Error when resetting the form


luishcastroc:
### Describe the bug

When i add the functionality to "reset" the form it will break the implementation.

### Your minimal, reproducible example

https://stackblitz.com/github/tanstack/form/tree/main/examples/angular/simple?embed=1&theme=light&preset=node

### Steps to reproduce

1. Open the Stackblitz
2. add a button for clear using this code:
```html
<button type="reset" (click)="form.reset()">
        Clear
      </button>
```
3. click on any input and get out to get the validation error.
4. click the clear button, validation should be gone
5. click on any input and get out of it
6. you should see an error.
```shell
chunk-3AFBTSQG.js?v=a3ca62aa:6796 ERROR TypeError: Cannot convert undefined or null to object
    at Function.values (<anonymous>)
    at Object.onUpdate (@tanstack_angular-form.js?v=7c012f42:961:38)
    at Store.setState (@tanstack_angular-form.js?v=7c012f42:43:87)
    at @tanstack_angular-form.js?v=7c012f42:681:24
    at Store.batch (@tanstack_angular-form.js?v=7c012f42:60:7)
    at @tanstack_angular-form.js?v=7c012f42:674:20
    at @tanstack_angular-form.js?v=7c012f42:53:9
    at Set.forEach (<anonymous>)
    at Store._flush (@tanstack_angular-form.js?v=7c012f42:50:22)
    at Store.setState (@tanstack_angular-form.js?v=7c012f42:44:12)
    ```

### Expected behavior

The form should be cleared and validators should reset with no errors.

### How often does this bug happen?

Every time

### Platform

- OS: Mac OS
- Browser: Chrome Version 123.0.6312.107 (Official Build) (arm64)

### TanStack Form adapter

None

### TanStack Form version

@tanstack/angular-form": "^0.19.0"

### TypeScript version

5.4.2

### Additional context

There's no option to add the proper platform to the issue but the platform is **Angular**

crutchcorn:
This should be fixed in the latest version

PR #674, opened 2024-04-13
fix(formApi): prevent reset from wiping validators in fieldMeta

luishcastroc:
PR should fix issue #671 

The new behavior now respects whatever is inside the fieldMeta but resetting any validation that the field might have, as well added the reset button to React and Angular examples (it might need to be added or not to more examples) and i added a single test.


crutchcorn (approved):
This code and sample LGTM, let's see if CI likes it and if it does we'll merge :)

PR #675, opened 2024-04-14
docs: debugging guidance for controlled => uncontrolled cases

bripkens:
# Why
Field unmount behavior can be rather confusing to understand and debug. We should provide guidance on the TanStack Form behavior and how to modify it.

# What
Extend TanStack Form React debugging instructions similar to how it has already been done for the uncontrolled => controlled case.


crutchcorn:
This is a confusing PR to me. The document you're updating already provides guidance on how to fix this issue, and further the suggested fix has behavior shifts from expected defaults.

Can you provide an example where your code runs this error in a new GH issue?

PR #713, opened 2024-05-20
feat: add support for nullable objects.

MAST1999:

Issue #691, opened 2024-04-27
Cannot infer types on arrays of union of objects


Balastrong:
### Describe the bug

The library cannot infer the right type of fields in a (probably weird) edge case, when an array is an union type of objects.

Take this as example:
```ts
type Text = {
  type: 'text';
  answer: string;
  textAnswer: string;
};

type Number = {
  type: 'number';
  answer: number;
};

type FormType = {
  questions: Array<Text | Number>
}
```

When accessing to `questions[0].answer` (field in both types) the type is `string | number`.
When accessing to `questions[0].textAnswer` (field on one type only) the type is `unkown`

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-form-ydorgn?file=src%2Findex.tsx%3AL53-L53

### Steps to reproduce

1. Have a structure with an array of union of objects
2. Use field.state in jsx

### Expected behavior

Not sure if it's a bug or a missing feature, but I'd like a way to know which is the correct type of the array entry and have all the fields typed accordingly.

The form is handling the values right, typescript isn't.

See the MRE for more comments.

### How often does this bug happen?

Every time

### Platform

macOS

### TanStack Form adapter

react-form

### TanStack Form version

v0.19.4

irwinarruda:
I found this problem interesting so I investigated a little. From what I've gathered, the issue revolves around the `DeepValue` type and happens because of a default typescript behavior with union types:
```ts
type A = { a: string; };
type B = { b: number; };
type C = boolean;
// Getting a value from a union causes some problems and the DeepValue uses this approach
type result1 = (A | B)["a"]; // any
type result2 = (B | C)["b"]; // any
```
My idea to solve it boils down to creating a new `Get` type that behaves differently from the default typescript one:
```ts
type Get<T, K extends string /* or keyof T*/> = T extends {
  [Key in K]: infer V;
} ? V : never;
type result1 = Get<A | B, "a">; // string
type result2 = Get<B | C, "b">; // number
```
Though I created it, I don't fully know how the `Get` type works because it sometimes behaves strangely, so this solution feels a bit hacky. Look at what I found changing the `else` clause. It might be a `typescript` bug.
```ts
type Get<...> = T extends { ... } ? V : unknown;
type result1 = Get<A | B, "a">; // unknown
type result2 = Get<B | C, "b">; // unknown

type Get<...> = T extends { ... } ? V : undefined;
type result1 = Get<A | B, "a">; // string | undefined
[...]

PR #724, opened 2024-05-30
fix(core): add support to union types for `DeepValue`

irwinarruda:
Solves #691

### Explanation

@Balastrong provided two issues in the example. Only one of those is a bug. Consider the following code:

```ts
type Text = { answer: string; textAnswer: string }
type Number = { answer: number }
type FormType = { questions: Array<Text | Number> }
```

- When calling `DeepValue<FormType, 'questions[0].textAnswer'>` the result is `unknown`. This is a bug!
- When calling `DeepValue<FormType, 'questions[0].answer'>` the result is `number | string`. This is not a bug because Typescript will never know the current state of the if checks.

### Solution

As I explained on the issue page, the problem was solved by swapping the default typescript get function with a custom one.

### Problems found along the way

Recently, the #713 and #717 were merged. From what I gathered, if there is a nullable union, all the properties from the object part will also be nullable. So for example:

```ts
type Obj = { a: { b: string } | null }
// DeepValue<Obj, 'a.b'> -> string | null
```

In order to create a standardized behavior, I decided to ask you some questions.

- Why is it only `null` values? What about `optional` and `undefined`?
- Shouldn't it work with deep nested values?

```ts
DeepValue<{ a: { b: { c: string } } | null }, 'a.b.c'>
// should be -> string | null but is only string
```

- Should the object union work the same way? So for the first example I gave:

```ts
DeepValue<FormType, 'questions[0].textAnswer'>
// should be -> string | undefined instead of just string (which is the way I implemented it)
```

Please let me know if you would like to chat more about it. I recently joined the Tanstack Discord.


crutchcorn (approved):
This is awesome, thank you!

irwinarruda:
The latest commit attempts to fix some of the issues discussed in the previous message. 

It was clear that solving the Deep Nested Nullable problem would help with almost all other issues, so that's where I started. 

After reading the code a bit more, the current implementation made sense and should've worked. I tried a lot of other approaches but none of them seemed to work at all. Then, I did the following test manually executing the DeepValue recursion:

```ts
type Obj2 = { user: { name: { first: string } } | null }
type normal1 = DeepValue<Obj2, 'user.name.first'> // string **Wrong! Should be string | null**
// The following is what happens recursively. 
type manual1 = DeepValue<Obj2, 'user'> // { name: { first: string; } } | null **Correct**
type manual2 = DeepValue<manual1, 'name'> // { first: string; } | null **Correct**
type manual3 = DeepValue<manual2, 'first'> // string | null **The correct result we wanted**
```
I've had similar experiences, and I think it's something with `typescript` itself. The workaround was changing one of the if statements and adding my Get function (I don't actually know why it worked and that's why I think it's a typescript bug):

```ts
export type DeepValue<...> =
      TValue extends ReadonlyArray<any>
      ? ...
      : // Before, it was TValue extends Record<string | number, any> 
        TValue extends any
        ...
                ? TNullable extends true
                  // Get is needed here because it can handle anything whether it is an object or not.
                  ? Nullable<Get<TValue, TAccessor>>
                  : Get<TValue, TAccessor>
                : never
```
Though this implementation worked, it introduced a strange typescript error `TS2589: Type instantiation is excessively deep and possibly infinite.` - I say that because any editor I use does not pick up on that error consistently.

After trying a lot of things, I created a `TDepth` arg to `DeepValue` passing all tests. 

I'm writing this to explain why I introduced a bit more complexity to the codebase. Also if it actually is a `typescript` bug maybe we should not do workarounds and instead open an issue since the code change seems very dangerous.

The only thing that did not work exactly as I wanted is described in the following test case:
```ts
type DoubleNestedNullableObjectCase = {
  mixed?: { mainUser: { name: 'name' } } | null | undefined
}
type DoubleNestedNullableObjectA = DeepValue<
  DoubleNestedNullableObjectCase,
  'mixed.mainUser'
>
// Notice how the `name` property is not nullable or undefined.
// It would be much harder to do that, but I wanted something like:
// assertType<{ name: 'name' | null | undefined } | null | undefined>(
assertType<{ name: 'name' } | null | undefined>(
  0 as never as DoubleNestedNullableObjectA,
)
```

crutchcorn:
We'd previously had issues with `TDepth` that we removed thanks to @chorobin. When we had `TDepth`, it caused major headaches that I'd like us to try to avoid merging back into the codebase if at all possible - as it leads to the `possibly infinite` issues.

@chorobin could you take a look at this issue and try to help us figure out what's going on and maybe open a GH issue with upstream TS if needed?

irwinarruda:
I dreamt about a way to remove the `TDepth` argument. It actually worked! It's still a dangerous change that's worth looking more into.

crutchcorn:
Closing in favor of #1180 based no the modern codebase. @irwinarruda I owe you a huge apology for not merging this **so much sooner**. This was a massive boon to our codebase that should have been merged ages ago.

We're adding back `TDepth` after a **lot** of testing to see if we could remove it after all, but the code is your second to last commit on this PR.

Again, from the deepest part of my heart: **thank you**.

irwinarruda:
It was also my bad for not participating more in the discussion. I'm happy you solved the problem!

Issue #723, opened 2024-05-29
`canSubmit` is always `true` on first render of form


engelkes-finstreet:
### Describe the bug

I am using a form with the `zodValidator` and have several field-level validations. On the first render the `canSubmit` property is true even though several fields are invalid. After changing the first input the validations run and `canSubmit` is invalid.

### Your minimal, reproducible example

https://tanstack.com/form/latest/docs/framework/react/examples/zod

### Steps to reproduce

1. Go to the [official zod example](https://tanstack.com/form/latest/docs/framework/react/examples/zod)
2. The `Submit` button is enabled even though the firstName has a minLength validation of 3
3. Type something in any of the fields
4. The `Submit` button is disabled until all the validations are correct

### Expected behavior

I would expect that the `canSubmit` property considers all validations on the first render and shows a disabled `Submit` button from the start instead of changing from enabled to disabled after typing one letter and going back to enabled after all validations are correct.

### How often does this bug happen?

Every time

### Platform

OS: MacOS
Browser: Arc

### TanStack Form adapter

react-form

### TanStack Form version

0.20.2

Balastrong:
Hey! The current behavior in the example seems correct to me, the validator is triggered `onChange` hence I'd expect no validation to be run as long as the value doesn't change.

What you're trying to achieve might be validating `onMount`.

---

With that said, if we agree that `onMount` is the answer, I noticed that there might be a bug there preventing `canSubmit` to go false but it's another topic we might want to tackle (and I already have an idea why we have this bug).

afriddev:
** I figured out the solution for `canSubmit` return `true` when your component renders,you need to validate all your form.Filed when it mounts **

`function runOnMount({ value }: any) {
    form.validateAllFields("change");
    return value;
  }`


`validators={{
                onChange: (value) => {
                  return handleMobileNumberChange(value?.value);
                },
                onMount: runOnMount,
              }}`

`<form.Subscribe
          selector={(state) => [state.canSubmit, state.isSubmitting]}
        >
          {([canSubmit, isSubmitting]) => {
            return (
              <Button disabled={!canSubmit} type="submit">
                {!isSubmitting ? (!edit ? SUBMIT : MODIFY) : "Submiting ..."}
              </Button>
            );
          }}
        </form.Subscribe>`

Glazy:
Not sure what the answer is here but I also encountered this same behaviour and find it confusing.

[...]

PR #726, opened 2024-05-30
fix(core): remove mount validation on touch

### LINKED ISSUE #689: onMount Error is not revalidated when the user changes the input
### Describe the bug

If you have an `onMount` that returns an error on the first render, changing the value after that does not remove the error.

https://stackblitz.com/edit/vitejs-vite-yjklcs?file=src%2FApp.tsx

### Steps to reproduce

Open the reproduction and see that the `onMount` returns error, but after changing the value the error still remains.

### Expected behavior

The `onMount` error should be revalidated (removed) after user changes the value.

Every time

### Screenshots or Videos

_No response_

### Platform

OS: Linux
Browser: Chrome
Version: 124.0.63

### TanStack Form adapter

None

### TanStack Form version

^0.19.4

### TypeScript version

^5.4.5

### Additional context

_No response_

Balastrong:
This PR closes #689 

A few considerations:

1. When should we remove the errors?
My proposal would be to remove each field individually when gets touched and the form as soon as a field is touched.
As long as a field is not touched, it keeps the mount errors.

2. Should we be able to submit a form with onMount errors?
I'd say no, but currently if `state.submissionAttempts === 0 && !isTouched` we have `canSubmit: true`
Adding `&& !hasOnMountError` also aims to fix the unwanted behavior described in #723 

3. We might want to explicitly explain in the docs how we intend onMount to behave
I'm happy to improve the docs in this PR if we decide to go through with the proposed behavior

4. Bonus
The amount of eslint disable on form.errorsMap makes me wonder if we should consider it as optional in first place, but that's maybe another topic.


brandonryan:
This really tripped our team up. We expected that if we specify onMount + onChange validators then when the field is "changed" it would clear the onMount error. I think clearing the mount error on change is probably a reasonable way to address it.

Issue #667, opened 2024-04-05
Add a way to display only one error for a field


francois-pasquier:
### Describe the bug

Currently we return `field.state.meta.touchedErrors` `ValidationError[]` as a string by concataining the different errors of a field with commas.

It would be great to be able to display only one error.

e.g:

```
         validators={{
            onChange: v.string([
              v.minLength(1, t(i18n)`This field is mandatory`),
              v.custom(isDateStringValid, t(i18n)`The birth date format is incorrect`),
              v.custom(isDateStringOver18, t(i18n)`You have to be at least 18 years`),
            ]),
          }}
```

These valibot validators will end up on showing all three errors just by typing a letter and deleting it.
Imo, this is not a good UX.

I would expect some kind of priority in the error messages, thus only displaying the first validation to return an error.

### Your minimal, reproducible example

https://github.com/TanStack/form/

### Steps to reproduce

         validators={{
            onChange: v.string([
              v.minLength(1, t(i18n)`This field is mandatory`),
              v.custom(isDateStringValid, t(i18n)`The birth date format is incorrect`),
              v.custom(isDateStringOver18, t(i18n)`You have to be at least 18 years`),
            ]),
          }}

### Expected behavior

More control over the errors display

### How often does this bug happen?

None

### Platform

macOS

### TanStack Form adapter

react-form

### TanStack Form version

0.19.0

crutchcorn:
This should be done at the validator level where you can pass `transformErrors` to `transformErrors: errors => errors[0]` rather than   `transformErrors: errors => errors.join(', ')`

Will mark this as "Good first issue" and a feature request

seobbang:
Hello, can I work on this issue? If possible, please assign it to me.

seobbang:
@crutchcorn Hello, can I work on this issue? If possible, please assign it to me. 😊

jessekelly881:
Not sure if this is relevant to this issue or if I should create a separate one but being able to return an array from 

```ts
validators={{
    onChange: ["err1", "err2"] // string | string[]
}}
```

would be a very nice addition.

crutchcorn:
@jessekelly881 let's open a different issue on it :) 

@seobbang PRs welcome! No need to assign it out, but do feel free to tackle the issue :)

TheKnightCoder:
Would love this fixed @seobbang did you manage to work on this?

seobbang:
@TheKnightCoder I'm going to try again 🤗

[...]

PR #727, opened 2024-05-30
For valibot validator, Add a way to display only one error for a field

seobbang:
**Related issue :** #667

Referring to the [comments](https://github.com/TanStack/form/issues/667#issuecomment-2043786027) of @crutchcorn I solved it by changing the code as below.

```js
return {
    validate({ value }, fn) {
      if (fn.async) return
      const result = safeParse(fn, value)
      if (result.success) return
      return result.issues[0].message // here
    },
    async validateAsync({ value }, fn) {
      const result = await safeParseAsync(fn, value)
      if (result.success) return
      return result.issues[0].message // here
    },
```
<!--
### Additional Question
Now the error shown on the UI became only the first error,
however, at the same time valibot validator's `field.getMeta().errors` and `field.state.meta.touchedErrors` are returning Array containing only the first error.

I have not changed the code anymore because I thought the current test code has no problem validating `valibot validator` logic. Also, `touchedErrors` is used by various validators.

I was wondering if there would be any additional test modifications or code modifications needed for this. --!>

### DIFF
--- a/packages/valibot-form-adapter/src/validator.ts
+++ b/packages/valibot-form-adapter/src/validator.ts
@@ -8,12 +8,12 @@ export const valibotValidator = (() => {
       if (fn.async) return
       const result = safeParse(fn, value)
       if (result.success) return
-      return result.issues.map((i) => i.message).join(', ')
+      return result.issues[0].message
     },
     async validateAsync({ value }, fn) {
       const result = await safeParseAsync(fn, value)
       if (result.success) return
-      return result.issues.map((i) => i.message).join(', ')
+      return result.issues[0].message
     },
   }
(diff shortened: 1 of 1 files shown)


crutchcorn (changes requested):
Hey there! This isn't _quite_ what I meant with this API change 😅

See, I was thinking more like:

```
valibotAdapter({errorMap: errors => errors[0]})
```

Where `errorMap`'s default is `errors => errors.join(", ")`

What do you think?
```

crutchcorn:
Temporarily closing, as I'm sorting through PRs and trying to figure out which ones to merge 😇 

Feel free to re-ping me if you've made the requested changes.

PR #733, opened 2024-06-03
chore: fix React TSC usage for 4.9

crutchcorn:
This is a first: React's types conditionally export a different version for 4.9 and 5.0:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/package.json#L10-L15

Among the changes between 4.9 and 5.0 appear to be a mismatch between React.JSX.Element and React.ReactNode, which was causing a failure in our build/TypeScript 4.9 support.

This PR replaces usages of `ReactNode` and `React.JSX.Element` with a custom type of:

```typescript
export type NodeType = ReturnType<FunctionComponent>
```

Which solves this incompatibility, while retaining our strict types.

### DIFF
--- a/packages/react-form/src/tests/useField.test-d.tsx
+++ b/packages/react-form/src/tests/useField.test-d.tsx
@@ -17,12 +17,14 @@ it('should type state.value properly', () => {
           name="firstName"
           children={(field) => {
             assertType<'test'>(field.state.value)
+            return null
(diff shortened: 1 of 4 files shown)


crutchcorn:
Nx is still having issues. `test:ci` runs locally and caught a few small things. Merging

Issue #738, opened 2024-06-06
DeepKeys type not working with all partial object


fkworld:
### Describe the bug

`DeepKeys` not working with all partial object, expect returns object keys, but returns `string`.

### Your minimal, reproducible example

see Steps to reproduce

### Steps to reproduce

```ts
import type { DeepKeys } from "@tanstack/react-form";

type AllPartialObject = {
  a?: string;
  b?: number;
  c?: {
    c1: string;
    c2: string;
  };
};

// it returns string
type AllPartialObjectKeys = DeepKeys<AllPartialObject>;

type PartPartialObject = {
  a?: string;
  b: string;
  c?: {
    c1: string;
    c2: string;
  };
};

// it returns "a" | "b" | "c" | "c.c1" | "c.c2" | undefined
type PartPartialObjectKeys = DeepKeys<PartPartialObject>;
```

### Expected behavior

DeepKeys<AllPartialObject> returns all object keys.

### How often does this bug happen?

None

### Platform

No need.

### TanStack Form adapter

None

### TanStack Form version

v0.20.3

### TypeScript version

v5.2.2

Balastrong:
Thanks! Adding one more case I found with one partial nested field:

```ts
type NestedPartialObject = {
  c: {
    c1?: string
    c2: string
  }
}

// "c" | "cundefined" | "c.c1" | "c.c2"
type PartPartialObjectKeys = DeepKeys<NestedPartialObject>
```

PR #739, opened 2024-06-06
fix(core): fix DeepKeys not support partial nested object

fkworld:
fix https://github.com/TanStack/form/issues/738

## About `object extends T`

Typescript will see `{a?:number}` as `{ a: number | undefined } | {}`, so `object extends T` will return `true`.

Which cause `DeepKeys<{a?:number}>` go to `PrefixFromDepth<string, TDepth>`, returns `string`.

**Danger**: Delete it will cause `DeepKeys` handle other types: like `class`. I don't know if this is expected.

```ts
class SomeClass {
  someAttribute = "hello world";
  someFunction() {
    return this.someAttribute;
  }
}
type A = DeepKeys<{ a: SomeClass; b: string }>; // "a" | "b" | "a.someAttribute" | "a.someFunction"
```

## About `-?:`

In `PrefixObjectAccessor` type, if `K` is partial, `T[K]` will union with `undefined`. So `PrefixObjectAccessor` result will get `undefined`.

### DIFF
--- a/packages/form-core/src/tests/util-types.test-d.ts
+++ b/packages/form-core/src/tests/util-types.test-d.ts
@@ -42,6 +42,18 @@ assertType<
 >(0 as never as NestedSupport)
 
 /**
+ * Properly handles deep partial object nesting like so:
+ */
+type NestedPartialSupport = DeepKeys<{ meta?: { mainUser?: User } }>
+assertType<
+  | 'meta'
+  | 'meta.mainUser'
+  | 'meta.mainUser.name'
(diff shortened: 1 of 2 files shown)


Balastrong (approved):
LGTM!

Let's also wait the TS wizards to have a look :D

crutchcorn (approved):
This broadly looks good to me. I appreciate your highlighting of other concerning edgecases, but I think our tests and examples _should_ be enough to catch issues.

Thanks for this! These are _hard_ types to work with!

PR #761, opened 2024-06-20
chore: Temporarily disable type checking for Next.js example

lachlancollins:
The Next server actions example is having issues due to pnpm resolution of the global React 18 types and its local React 19 RC types:

`Type 'import("/home/workflows/workspace/node_modules/.pnpm/@types+react@18.3.3/node_modules/@types/react/index").ReactNode' is not assignable to type 'React.ReactNode'`

This PR temporarily disables its type checking until the React 19 release

### DIFF
--- a/examples/react/next-server-actions/package.json
+++ b/examples/react/next-server-actions/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "next dev",
     "build": "next build",
-    "test:types": "tsc",
+    "_test:types": "tsc",
     "test:eslint": "next lint"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


crutchcorn:
Is this resolution actually failing in prod usage or just our PNPM local env?

lachlancollins:
> Is this resolution actually failing in prod usage or just our PNPM local env?

`test:types` actually works locally, it only fails on CI for some reason. I'm pretty confident it's a pnpm issue, I've seen something similar before when there were duplicated versions of the package (hence introducing sherif)

crutchcorn:
Thanks for helping investigate this. Was a real headache and lost my cool trying to debug, your help is immensely appreciated

Issue #785, opened 2024-06-27
allow to reset form with new values


crutchcorn:
### Discussed in https://github.com/TanStack/form/discussions/596

<div type='discussions-op-text'>

<sup>Originally posted by **schiller-manuel** February  8, 2024</sup>
After form submission I want to keep the form open and make the form pristine again, but don't want to lose the current values.

Inspiration: In RHF this is possible via calling `reset` with new values: https://www.react-hook-form.com/api/useform/reset/</div>

vara855:
@crutchcorn Hi. I'm wanna to contribute. Can you assign it to me?

Balastrong:
Done in #791

PR #791, opened 2024-06-28
feat: added ability to specify values for `form.reset` method

vara855:
#785


thomasvt1:
This would be nice!

Any ideas when this would be merged?

abel-altex (approved):

Balastrong (commented):
Thanks for the PR, looks like a great starting point!
Let me add a thought, this seems to me more like a `setAll` than a proper `reset`, which I would expect also setting the new default values.

Doing this (in two separate moments):
```ts
reset({ firstName: "foo" }) // Passing an object - set the new default
reset() // No parameters - reset to default
```
Should end up with the form this the state `{ firstName: "foo" }`.
At least as a default behaviour, then nothing stops us from adding an extra parameter with a flags, similar to RHF, to control that.

What do you think?

zwenger:
a functionality like this would be pretty helpful.

PR #796, opened 2024-06-29
fix(core): abort async validation if sync validation fails in the meantime

Balastrong:
The async validator (arriving late by its nature) overrides the result of the sync validator, ending up in a valid component when async is valid but sync is not.

With debounce:
1. Input a valid value for the sync validator
2. Async validator is _enqueued_
3. Input an invalid value for the sync validator
4. Async validator _is called_ and when resolves if valid removes the errors

Without debounce, but async takes some time:
1. Input a valid value for the sync validator
2. Async validator _is called_
3. Input an invalid value for the sync validator
4. Async validator _resolves_ and if valid removes the errors

It's easy to reproduce in the [simple example](https://tanstack.com/form/latest/docs/framework/react/examples/simple), by writing `aaaa` and waiting the async validator, then rapidly deleting two chars -> field is considered as valid.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -875,6 +875,7 @@ export class FieldApi<
           } catch (e: unknown) {
             rawError = e as ValidationError
           }
+          if (controller.signal.aborted) return resolve(undefined)
           const error = normalizeError(rawError)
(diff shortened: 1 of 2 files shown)


crutchcorn (approved):
Great catch and good fix - thanks!

PR #801, opened 2024-06-30
feat(core): add field listeners

ha1fstack:
from discussion #709 

### Background

Sometimes we may want to attach our own side effects to the field when some event happends. For example, resetting another field if some field value has changed.

Attaching these 'listeners' inside the validator does not work as intended because validator does not guarantee if it was triggered by the exact event. (eg. all validators run when form submits)

It's of course possible to implement but it's not very clean and the concerns become scattered.

### Proposal

This PR adds `listeners` api to the field for easy handling of forementioned use cases.

`onChange` and `onHandleChange` are differentiated because `onHandleChange` captures the 'manual' changes to the field. (eg. user input) while `onChange` also captures programmatic changes. (eg. `setValue`)

### Thoughts

I'm not certain about the whole idea or implementation, open to suggestions.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -129,6 +129,24 @@ export type FieldAsyncValidateOrFn<
         TData
       >
 
+/**
+ * @private
+ */
+export type FieldListenerFn<
+  TParentData,
+  TName extends DeepKeys<TParentData>,
+  TFieldValidator extends
+    | Validator<DeepValue<TParentData, TName>, unknown>
+    | undefined = undefined,
+  TFormValidator extends
+    | Validator<TParentData, unknown>
+    | undefined = undefined,
+  TData extends DeepValue<TParentData, TName> = DeepValue<TParentData, TName>,
+> = (props: {
+  value: TData
+  fieldApi: FieldApi<TParentData, TName, TFieldValidator, TFormValidator, TData>
+}) => void
+
 export interface FieldValidators<
   TParentData,
   TName extends DeepKeys<TParentData>,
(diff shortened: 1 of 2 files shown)


zaosoula:
This feature will be useful +1

crutchcorn:
@zaosoula this would be helpful to understand a practical use case :)

zaosoula:
@crutchcorn 

>Screenshot 2024-07-04 at 09 48 09
>
> Here we are using onChange to dynamically set a checkbox to true if the selected value match a list

>Screenshot 2024-07-04 at 09 49 11
>
>Here we are using onChangeAsync to open a modal to create a new entry for a sub-relation

Each time we have to return null or an empty string as we do not want to show errors; using the validators property is also confusing for new developper reading our code as the functions above are clearly not validators

crutchcorn:
This alone makes a lot of sense to me. Let's move forward with the idea. Reviewing the code now

packages/form-core/src/FieldApi.ts
```diff
@@ -251,6 +269,40 @@ export interface FieldValidators<
...
+    TFormValidator,
+    TData
+  >
+  onHandleChange?: FieldListenerFn<
+    TParentData,
+    TName,
+    TFieldValidator,
+    TFormValidator,
+    TData
+  >
```
crutchcorn, line 296: I'm not convinced we need a different function for programmatic vs user input. If the user needs this distinction, they can do so outside of our `listeners`

It also opens a naming can of worms and adds another one when we need an `onBlurChange` kinda deal

packages/form-core/src/FieldApi.ts
```diff
@@ -251,6 +269,40 @@ export interface FieldValidators<
...
+  onBlur?: FieldListenerFn<
+    TParentData,
+    TName,
+    TFieldValidator,
+    TFormValidator,
+    TData
+  >
```
crutchcorn, line 303: Do we want an `onSubmit` and `onMount` here as well?

zaosoula: Yes it would give more versatility to the feature, and maybe onSubmit should allow for transformation before sending out the data ?

Balastrong:
Completed in #1032

PR #806, opened 2024-07-01
refactor: remove unnecessary `async` keywords

### LINKED ISSUE #805: CI has a lot (~20) of warnings about unnecessary `await` expressions
This clutters log output and adds unnecessary noise.

hamirmahal:
fixes #805.

### DIFF
--- a/packages/form-core/tests/FieldApi.spec.ts
+++ b/packages/form-core/tests/FieldApi.spec.ts
@@ -105,7 +105,7 @@ describe('field api', () => {
     expect(field.getValue()).toStrictEqual(['one', 'other'])
   })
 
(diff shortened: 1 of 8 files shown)


packages/form-core/tests/FieldApi.spec.ts
```diff
@@ -642,7 +642,7 @@ describe('field api', () => {
       return
     })
 
-    const mockOnChangeAsync = vi.fn().mockImplementation(async ({ value }) => {
```
hamirmahal, line 645: If `async` is necessary in places like these it might make sense to add `eslint-disable` to reduce noise in the output.

PR #809, opened 2024-07-02
test: remove first level sleep in tests and use fake timers

Balastrong:


lachlancollins (approved):
Works locally with `pnpm test:ci --skip-nx-cache`, fingers crossed it works in CI!

lachlancollins:
@Balastrong I re-ran the failed CI workflow and it passed (I think the tests were just flaky). Hopefully this PR is a more stable solution!

Balastrong:
Thank you!

PR #820, opened 2024-07-07
Fixed value type coming from `TextInputFieldProps`

Andarist:


crutchcorn:
Thank you for your help and discussion throughout.

That said, I'm closing this PR, as this isn't exactly the API we're wanting for this.

Issue #786, opened 2024-06-27
Add an API to set errors


crutchcorn:
```
field.setErrorMap();
```

> We cannot provide a `setErrors` API as it derives that state from the `errorMap`

And:

```
form.resetErrors();
```

> I'm less certain about this API. Why do we need it? CC @samuelthng

Toavina23:
hello @crutchcorn 

I would like to work on this issue. 
Is there any advice or guide that you can give the correctly implement this feature ?

crutchcorn:
Hey @Toavina23 the idea is that you'll add it to:

https://github.com/TanStack/form/blob/main/packages/form-core/src/FormApi.ts

https://github.com/TanStack/form/blob/main/packages/form-core/src/FieldApi.ts

`field.setErrorMap()` should update the field's meta `errorMap` property

Balastrong:
Done in #821

xmlking:
I am trying this, but with recent TanStack From release,  I am getting `Uncaught TypeError: field.setErrorMap is not a function`  error.

 
```tsx
 

import { Textarea } from "@repo/ui/components/shadcn/textarea";
import type React from "react";
import { useEffect, useState } from "react";
import { BaseField, type FieldControlProps } from "./base-field";
import { useFieldContext } from "./context";

type JSONFieldProps = React.ComponentProps<typeof Textarea> & FieldControlProps;

export function JSONField({
  label,
  description,
  tooltip,
  tooltipSide,
  ...props
}: JSONFieldProps) {
  const field = useFieldContext<Record<string, unknown> | undefined>();
  // const form = field.form; // Access the form instance from the field
  const isInvalid = field.state.meta.isTouched && !field.state.meta.isValid;
  const [text, setText] = useState(() =>
    field.state.value ? JSON.stringify(field.state.value, null, 2) : ""
  );

  useEffect(() => {
    const value = field.state.value;
    const next = value ? JSON.stringify(value, null, 2) : "";
    setText(next);
  }, [field.state.value]);

  const applyChange = () => {
    const trimmed = text.trim();
    if (!trimmed) {
      field.handleChange(undefined);
      field.setErrorMap({ onBlur: undefined });
      return;
    }

    try {
      const parsed = JSON.parse(trimmed);
      if (!parsed || typeof parsed !== "object") {
        throw new Error("Invalid JSON");
      }
      field.handleChange(parsed as Record<string, unknown>);
      field.setErrorMap({ onBlur: undefined });
    } catch (error) {
[...]

PR #821, opened 2024-07-07
Add 'setErrorMap' api to 'formApi' and 'fieldApi'

Toavina23:
This PR adds an ability to set errorMap on a fieldApi on formApi as requested in #786.

work done:
- [x] adding the setErrorMap to both of fieldApi and formApi
- [x] write test for
    - [x] checking if the values are added
    - [x] checking if the other values on the errorMap are not overriden by incomming errorMap
    - [x] checking that only the concerned value is overriden when setting erroMap
- [x] add the necessary documentation on FormApi and FieldApi

Feel free to reach out if any of the implementation made needs to be corrected.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -963,6 +963,19 @@ export class FieldApi<
     }
     this.validate('blur')
   }
+
+  /**
+   * Updates the field's errorMap  
+   */
+  setErrorMap(errorMap: ValidationErrorMap) {
+    this.setMeta((prev) => ({
+      ...prev,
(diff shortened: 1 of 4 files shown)


Balastrong:
Thanks for the PR, it looks really good!

For the two conflicts could you please just revert the changes? Reference docs will be autogenerated (this is something rather new, we'll probably have to write it down somewhere).

Toavina23:
Hello @Balastrong , 
Thanks for your feedback.
I think it should be good now.

Balastrong (approved):
Looks good now, thanks again!

Issue #737, opened 2024-06-05
Type Error onServerValidate async


CoachNooB:
### Describe the bug

I got red squigly line when putting async in onServerValidate. But it run normally in runtime.
Type Promise <string | undefined> is not assignable to type ValidationError

### Your minimal, reproducible example

https://www.typescriptlang.org/play/?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydaiKKEEeBkGgAb1A7EEkAAYlTEKAAL44thsgBEAAFUGZBILYPFgJNYKhSslWTyANzBBCcYU41mgAC89MZLLIiAAFDTFCUStxIMkGFFUAA1BjQdiQfKgQ3Gl39BjIR08nlG43s3jBF2NYjfPRkG3QCjcOoOsTHcygA2gHS2+0cgCUGoAfE6fa7QCoACIYCpx2CgYNkUOgcORuoNTi5jmKdlp4IqHDNDxkLyIKnOYHsGhfNxmZjsBjMSD8WAR0ViYF6EReHkAUVW7oW+h5SWxPPADlQAHJhAA5SASHlBFQAeS+ZHejP4k6OgupkFc0jwoCcvZ0wyw+goAgiDQsIlzDhIVLRCIBgcNiCxMKwHA8EEihAA

### Steps to reproduce

1. Put async in the onServerValidate
2. Got Type error

### Expected behavior

No Type Error.

### How often does this bug happen?

Every time

### Screenshots or Videos

![image](https://github.com/TanStack/form/assets/55129369/05c5c8d2-285b-40d9-a840-ca06fce01d6a)

### Platform

- OS: Windows
- Browser: OperaGX

### TanStack Form adapter

[...]

PR #822, opened 2024-07-08
refactor(react-form/nextjs): update `onServerValidate` to enable async behavior

my-slab:
It's not possible to run async validations in the `onServerValidate` function. This PR enables that.

- Update `OnServerValidateFn` return type
- Add `async`/`await` keywords to `runValidator`
- Add tests

Fixes https://github.com/TanStack/form/issues/737

### DIFF
--- a/packages/react-form/src/nextjs/createServerValidate.ts
+++ b/packages/react-form/src/nextjs/createServerValidate.ts
@@ -9,7 +9,7 @@ import type { ServerFormState } from './types'
 
 type OnServerValidateFn<TFormData> = (props: {
   value: TFormData
-}) => ValidationError
+}) => ValidationError | Promise<ValidationError>
 
 interface CreateServerValidateOptions<
   TFormData,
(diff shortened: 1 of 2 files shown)


crutchcorn:
This is a great start!

Have you been able to test this with `zod` or any of the other adapters? I don't think this will work with those async values.

my-slab:
Thanks for reviewing. I've tested it with zod and it works as usual. I can give the other validators a go too.

Do you mean async actions like this?

```
const asyncZodOnServerValidate = createServerValidate({
    defaultValues: {
      name: "",
    },
    onServerValidate: z.object({
      name: z.string().refine(
        async (value) => {
          let error = await Promise.resolve("error");
          return value.includes(error);
        },
        { message: "Name can't contain 'error'" },
      ),
    }),
  });

```

You're right, it wont work because it needs to be called with the async version of the validator, like `safeParseAsync` in zod. Maybe it would make more sense to update the function to have its async counterpart like the event handlers (`onChange/onChangeAsync`) instead of what I've done?

PR #825, opened 2024-07-08
[WIP] Shared Component Library Docs

crutchcorn:
This PR: 

- [x] Adds an example of moving TanStack Form into your own shared component library
- [ ] Adds the docs to explain what the example is doing


Balastrong (approved):

izakfilmalter:
One thing I am running into with this branch, `field.state.value` seems to always be string. If you have something else as the value you run into a world of pain. What is the best course of action for field values that are not string?

Balastrong:
> One thing I am running into with this branch, `field.state.value` seems to always be string. If you have something else as the value you run into a world of pain. What is the best course of action for field values that are not string?

How are you defining your component?
```
function TextInputField<
  TFormData extends unknown,
  TName extends DeepKeyValueName<TFormData, string>,
>
```
Here you can change string to your data type

izakfilmalter:
CleanShot 2024-11-26 at 17 31 09

izakfilmalter:
Ended up doing this, which is stupid, but works.

```ts
type ExpirationSelectFieldProps<
  Key extends string,
  TFormData extends { [K in Key]: ExpirationItem | null },
  TName extends DeepKeyValueName<TFormData, ExpirationItem>,
> = Omit<FieldOptions<TFormData, TName>, 'name'> & {
  name: Key
  form: ReactFormApi<TFormData, any> & {
    state: FormState<TFormData>
  }
  selectProps?: Omit<ComponentProps<typeof Select>, 'onChange' | 'value'> & {
    label?: string
    placeholder?: string
    inputWrapperClassName?: string
  }
  symbol: string
}

export function ExpirationSelectField<
  Key extends string,
  TFormData extends { [K in Key]: ExpirationItem | null },
  TName extends DeepKeyValueName<TFormData, ExpirationItem>,
>(props: ExpirationSelectFieldProps<Key, TFormData, TName>) { ... } 
```

Meess:
I was also working on integrating it more smoothly, i.e. abstracting the boiler plate and allowing to make custom components of it. This is my first POC, works for my side project now. Any feedback / suggestions / improvements are always welcome.

Usage:
``` typescript
// app/SomeForm.tsx
import { FieldInput, Form } from "../lib/form/Form"
export const SomeForm = () => {

    const form = useForm({
        defaultValues: {
            name: '',
            id: '',
        }
    })

    return (
        <Form onSubmit={form.handleSubmit}>
            <div className="flex flex-col space-y-2">
                <FieldInput formT={form} name="name" placeholder="Name" />
                <FieldInput formT={form} name="id" placeholder="ID" />
                <button type="submit">Add</button>
            </div>
        </Form>
    )
}
```

Abstract form components created to reduce boiler plate and allow for custom custom form components  with @tanstack/form:
``` typescript
// lib/form/Form.tsx
import { DeepKeys, FieldApi, useForm } from '@tanstack/react-form'

export const Form: React.FC<React.FormHTMLAttributes<HTMLFormElement>> = ({
    children,
    ...props
}) => {
    return (
        <form
            {...props}
            onSubmit={(e) => {
                e.preventDefault()
                e.stopPropagation()
                if (props.onSubmit) {
                    props.onSubmit(e)
                }
            }}
        >
            {children}
        </form>
    )
}

export const FieldInfo = ({ field }: { field: FieldApi<any, any, any, any> }) => {
    return (
        <>
            {field.state.meta.isTouched && field.state.meta.errors.length ? (
                <em className=" text-red-500 text-xs">{field.state.meta.errors.join(',')}</em>
            ) : null}
            {field.state.meta.isValidating ? 'Validating...' : null}
        </>
    )
}

export const FieldInput = <TFormData, TNames extends DeepKeys<TFormData>>({
    formT,
    name,
    ...props
}: {
    formT: ReturnType<typeof useForm<TFormData, any>>
    name: TNames
} & React.InputHTMLAttributes<HTMLInputElement>) => (
    <formT.Field<TNames, any, any> name={name}>
        {(field) => (
            <>
                <input
                    value={field.state.value}
                    onChange={(e) => field.handleChange(e.target.value)}
                    onBlur={field.handleBlur}
                    className="w-full rounded-md border border-gray-300 px-4 py-2 transition duration-200 ease-in-out focus:border-transparent focus:outline-none focus:ring-2 focus:ring-blue-500"
                    {...props}
                />
                <FieldInfo field={field} />
            </>
        )}
    </formT.Field>
)
```

This works so far, but I might want to look into your solution @crutchcorn, with DeepKeyValueName, what's still going wrong (as it seems it's not merged)?

examples/react/custom-component-wrapper/src/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+  TFormData extends unknown,
+  TName extends DeepKeyValueName<TFormData, string>,
+> extends FieldOptions<TFormData, TName> {
+  form: ReactFormApi<TFormData, any>
+  // Your custom props
+  label: string
+}
+
+function TextInputField<
+  TFormData extends unknown,
```
RazerM, line 23: Does anyone know why `extends unknown` constrains `field.name` to `string` instead of `string | number`?

[`@typescript-eslint/no-unnecessary-type-constraint`](https://typescript-eslint.io/rules/no-unnecessary-type-constraint/) and other sources I've found claim that an unconstrained generic type parameter already defaults to `unknown`, but this must not be true. Maybe some TS wizard will say "oh that's the extends unknown trick it fixes ..." 😅

Before I found this issue I had some of the implementation done myself but couldn't work out why `DeepKeys` doesn't exclude `number`, among other TS issues.

It's also a shame that `form.Field<TName, any, string>` doesn't work, because the types of `field.state.value` and `field.handleChange` are exactly where it would be useful to have TypeScript help.

RazerM:
There's an issue with `DeepKeyValueName` when a nested object is nullable:

```typescript
  interface Fields {
    nested: {
      name: string
      labels: string[]
    } | null
  }

  const form = useForm<Fields>({
    defaultValues: {
      nested: { name: '', labels: [] },
    },
  })

  return <TextInputField form={form} name="nested.name" />
```

```
Type '"nested.name"' is not assignable to type '`nested.labels[${number}]`'.
```

crutchcorn:
@RazerM can you open a GH issue for the failing type error you're running into there?

RazerM:
@crutchcorn it's unique to `DeepKeyValueName` added in this PR, `<form.Field name={...} />` doesn't have the issue.

beeirl:
i've been adapting this branch into my project, and it's been working really well so far. one thing i've noticed: adding an array field to my form triggers a typescript error:
```
Type instantiation is excessively deep and possibly infinite
```

epenflow:
> i've been adapting this branch into my project, and it's been working really well so far. one thing i've noticed: adding an array field to my form triggers a typescript error:
> 
> ```
> Type instantiation is excessively deep and possibly infinite
> ```

Did you use vs code? If so, I think vs code fails to infer the related types. Just pass a generic 'any' type to resolve that issue.
image

beeirl:
@epenflow yeah, i'm on cursor. had the same thought that my ide could be causing this, so i need to test with a different one to verify. i meant to delete my comment here since i'm using [this branch](https://github.com/TanStack/form/blob/reusable-app-form-hook/examples/react/simple/src/app-form.tsx), which i originally thought was related, but it's not. i don't think your fix will work in my case.

examples/react/custom-component-wrapper/src/index.tsx
LeCarbonator: The issue comes from the way object keys are treated in distributive types.

See this issue: https://github.com/microsoft/TypeScript/issues/61242

PR #844, opened 2024-07-10
fix(solid-form): Build ESM, CJS, and source

Brendonovich:
This implements the changes discussed on Corbin's stream with Ryan.
- `esm` and `cjs` builds are now generated again for use in no-build environments
- The JSX-preserving build is now generated in `source`, keeping with Solid ecosystem conventions
- The `source` entrypoint is still `.jsx` to make sure that all JSX gets compiled properly


lachlancollins (approved):
Thanks for fixing this!

PR #855, opened 2024-07-14
Supporting async validation in `createServerValidate`

ogawa0071:
# Changes

Currently, server-side validation cannot be performed asynchronously. This is particularly inconvenient when using external validation adapters like Zod to check database values against form inputs.

Client-side validation can already be performed asynchronously, and since server-side operations are generally expected to be asynchronous, the impact of this change should be minimal.

- [x] Change `validatorAdapter().validate()` to `validatorAdapter().validateAsync()`
- [x] Fixed `onServerValidate` property is not supported for `validatorAdapter` (Ex. Zod) TypeScirpt types

# Sample Code

```ts:action.ts
"use server";

import {
  ServerValidateError,
  createServerValidate,
} from "@tanstack/react-form/nextjs";
import { zodValidator } from "@tanstack/zod-form-adapter";
import { redirect } from "next/navigation";
import { z } from "zod";

const serverSchema = schema.extend({
  firstName: z
    .string()
    .min(8, {
      message: "Server: First Name must be at least 8 characters.",
    })
    .refine(async () => {
      // In the actual implementation, queries to the database will be performed here.
      console.log("Server: Validate First Name...");
      await new Promise((resolve) => setTimeout(resolve, 1000));
      return true;
    }),
});

const serverValidate = createServerValidate({
  // TypeScript error on following line
  onServerValidate: serverSchema,
});

export default async function someAction(prev: unknown, formData: FormData) {
  try {
    await serverValidate(formData);
  } catch (e) {
    if (e instanceof ServerValidateError) {
      return e.formState;
    }

    // Some other error occurred while validating your form
    throw e;
  }

  // Your form has successfully validated!
  console.log(formData.get("firstName"));

  redirect("/");
}
```


ogawa0071:
TypeScript type implementation was referenced from the following code:
https://github.com/TanStack/form/blob/75d25dd16790f88768da75b48a87f0cfc0211fc3/packages/form-core/src/FormApi.ts#L44-L64

Balastrong:
Thanks for the PR! Looks good so far but let's doublecheck, have you tested also with validator adapters like zod? Could you also write some unit tests for that?

ogawa0071:
@Balastrong 
Thank you for your feedback!

- I have updated the branch to follow the latest main and added tests for the zod-form-adapter. Please review them.
- There are type tests for the FormApi in the zod-form-adapter. Do you think type tests for createServerValidate are necessary?
- I haven’t added tests for Valibot and yup yet. Are these required? Also, there are no existing tests for createServerValidate within react-form itself. Should I add them as well?

Looking forward to your guidance!

Balastrong (commented):
The PR is already ok, if you want to add more tests for valibot and yup it'd be great!

Tests on types are also welcome, up to you if you want to make them on this branch or on a new PR later.

Thank you!

ogawa0071:
@Balastrong 
I don't have time to act on it right now so I won't be adding any new tests in this PR.
Thanks for checking.

Balastrong:
I'll take care of that and merge then, thanks! :)

Balastrong (approved):

PR #868, opened 2024-07-18
feat: export ReactFormApi for better typescript package support 

Makisuo:
This solves the following Typescript error when building a react typescript package exporting useForm.

`has or is using name 'ReactFormApi' from external module @tanstack/react-form" but cannot be named.`


mistyharsh:
@timon-kosmoy If you need a temporary workaround and do not wish to change the code, you can do the module augmentation as follows:

```ts
// <REPO>/src/augment.d.ts

// Important for module augmentation
import * as form from '@tanstack/react-form';

declare module '@tanstack/react-form' {
  import { FormApi, FieldComponent } from '@tanstack/react-form';

  export interface FormApi<TFormData, TFormValidator extends Validator<TFormData, unknown> | undefined = undefined> {
    Field: FieldComponent<TFormData, TFormValidator>;
  };
}

```

Balastrong (approved):
We're probably going to export this anyway in #825 without breaking changes so let's get this merged :)

PR #869, opened 2024-07-19
docs: update the syntax highlighting for angular

mgechev:
Shikijs supports [TypeScript](https://github.com/shikijs/textmate-grammars-themes/blob/main/packages/tm-grammars/grammars/angular-ts.json) and [HTML](https://github.com/shikijs/textmate-grammars-themes/blob/main/packages/tm-grammars/grammars/angular-html.json) dialects, which improves the syntax highlighting for the Angular samples.

Screenshot 2024-07-18 at 16 58 21


Balastrong (approved):
Looks so much better now, thanks!

PR #892, opened 2024-08-08
Expose `SolidFormApi` type

oscartbeaumont:
I'm currently building a `InputField` abstraction wrapping a form and it would be useful to be able to directly name the type.

```ts
function InputField(opts: { form: SolidFormApi<T> } ) {
   return <opts.form.Field ... />
}
```

For now i'm doing the following but it's kinda annoying:
```ts
type FormResult<
	TData,
	T extends Validator<TData, unknown> | undefined = undefined,
> = ReturnType<typeof createForm<TData, T>>;
```


Balastrong:
Thanks for the PR! Can you also export the type from `packages/solid-form/src/index.tsx`?

---

On a separate note, we're drafting a recommended approach for building custom/wrapper components in React, but I'd say the needs and challenges are the same for all frameworks.

Since you're also doing something similar, if you have feedback please share them in #825 
An example component is on that PR: https://github.com/TanStack/form/pull/825/files#diff-b2e0803ded7363bf3432f160f7c7f6e9bf4f982ad13b08ecb4789c228774f956R22

Balastrong (approved):
Thank you!

Issue #903, opened 2024-08-13
Using both `onMount` and `onChange` validators on a field doesn't work


oscartbeaumont:
### Describe the bug

Providing both an `onMount` and `onChange` validators to `form.Field` does not run validation on change events.

For some reason removing the `onMount` validator, causes the `onChange` validator to work correctly.

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-form-cgm7xb?file=src%2Findex.tsx

### Steps to reproduce

Type more than 2 characters into the input field and notice the error doesn't go away despite a `onChange` validator being applied.

Now comment out the `onMount` validator and notice the `onChange` validator works, but obviously you get no error on initial load.

### Expected behavior

Typing more than 2 characters into the input field should cause the validator error to go away with both the `onMount` and `onChange` validator defined.

### How often does this bug happen?

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/1727e044-2515-465d-9f1f-a5a7c73d09ae

### Platform

Reproduced in Stackblitz so N/A.

### TanStack Form adapter

solid-form

### TanStack Form version

v0.29.1

### TypeScript version

v5.4.2

### Additional context

It looks like `onMount`'s validator error is just never cleared on user input. It's possible this is expected behavior and if so would you know if I can workaround this?

oscartbeaumont:
Okay I actually think this issue is different than I originally thought.

My assumption would be that all previous validation errors are cleared when running the next validation but that doesn't seem to be the case.

So `onMount` returning an error means the error will always exist and `onBlur` returning an error is only cleared on the next blur event.

Can I confirm if this is expected behaviour or not?

Balastrong:
I also think there's something we should look at with the `onMount` validator. Some time ago I started exploring a possible option that is removing the mount validation as soon as a field is touched (#726) but I also like your idea of removing all validation errors of a field as soon as any of its validators kicks in.

Maybe this can also be configured with an optional flag for the `validators` object? Looking for feedback :)

oscartbeaumont:
I have opened #907 with an implementation behind a configuration flag.

EyMaddis:
More annoyingly it `onMount` will keep the initial state forever. So even `form.validate('mount')` will effectively no nothing based on user input

vdvukhzhilov:
Is anything being done to resolve stuck `onMount` validation error?

Balastrong:
Fixed with #726 (version 0.34.4)

[...]

PR #907, opened 2024-08-17
Option for automatically resetting validators

oscartbeaumont:
A potential solution for #903.

I think this is a more general solution than #726 but they both cover the same issue.

My assumption coming into Tanstack Form was that all previous validation errors would be cleared out when running the next validation but that doesn't seem to be the case.

So `onMount` returning an error means the error will *always exist* and `onBlur` returning an error is only cleared on the next blur event.

## New behavior

When `form.options.automaticallyResetValidators === true`  previous validation errors are cleared when running the validator,
else retain the previous behaviour.

## Known problems

Calling `.validateSync` or `.validateAsync` will not cause the fields to be reset. These fields are however marked as `@private` so I think this is fine.

I think fixing this would also add a lot more complexity because the `.validate` method calls `.validateSync` and then `.validateAsync`, so we would need to know *not* to clear the errors that were just added by `.validateSync` in `.validateAsync`.

## Questions

- [ ] One major question I have is what should the default be? Personally I think it should be `true` but this would be a breaking change.

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -172,7 +172,7 @@ export interface FormOptions<
    *
    * @default false
    */
-  automaticallyResetValidators?: boolean;
+  automaticallyResetValidators?: boolean
 }
 
 /**
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -169,6 +169,7 @@ export interface FormOptions<
    * While `false`:
    *  - An `onMount` error will never be cleared.
    *  - An `onBlur` error will only be cleared on the next `onBlur` event.
+   *  - An `onChange` error will only be cleared on the next `onChange` event.
    *
    * @default false
    */
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -160,6 +160,19 @@ export interface FormOptions<
     formApi: FormApi<TFormData, TFormValidator>
   }) => void
   transform?: FormTransform<TFormData, TFormValidator>
+  /**
+   * When true, all previous validation errors will be cleared when running validation.
+   *
+   * While `true`:
+   *  - `onMount` and `onBlur` errors will be cleared when `onChange` triggers validation.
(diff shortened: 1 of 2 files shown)


packages/form-core/src/FormApi.ts
```diff
@@ -160,6 +160,20 @@ export interface FormOptions<
...
+  /**
+   * When true, all previous validation errors will be cleared when running validation.
+   *
+   * While `true`:
+   *  - `onMount` and `onBlur` errors will be cleared when `onChange` triggers validation.
+   *
+   * While `false`:
+   *  - An `onMount` error will never be cleared.
+   *  - An `onBlur` error will only be cleared on the next `onBlur` event.
+   *  - An `onChange` error will only be cleared on the next `onChange` event.
```
Balastrong, line 172: I'd probably add an extra line on top like "Some examples:" since those aren't hardcoded rules.

packages/form-core/src/FormApi.ts
```diff
@@ -160,6 +160,20 @@ export interface FormOptions<
...
+   *
+   * @default false
+   */
+  automaticallyResetValidators?: boolean
```
Balastrong, line 176: The action this flag enables is that errors are removed when a new validator is run, I think we should better explain this.

I like `resetOnRevalidate` to indicate that an action (reset) is performed when something happens (onRevalidate) or `resetErrorsOnRevalidate` if we want to be extra explicit.

What do you think?

Balastrong (commented):
Thanks for the PR, the overall idea looks good!

To answer the question of what the default behaviour should be, I think it makes more sense to keep this false for primarily two reasons:
1. As you also mentioned, that would be a breaking change
2. This adds extra complexity on setting validators, most notably every time you trigger blur on an input it will remove the onChange errors (if a blur validator is set) which might not be expected.

This last point is also the main reason why my initial idea was to only focus on the mount errors.

Would you mind adding some unit tests? I'll be happy to merge this!

Balastrong:
Superseded by #726 but we can keep discussing in #1005 for improvements in the API

PR #908, opened 2024-08-17
Deduplicate field and form `errors` property

oscartbeaumont:
The current behavior of `form.state.errors` and `field.getMeta().errors` is to collect up all errors within the `errorMap`. The implementation *does not* take care of removing duplicates and Form's unit tests assert this behaviour as expected.

I think this behaviour should potentially be reconsidered.

Personally I think it's very common as an application developer to apply the same validator to multiple events (Eg, `onMount` & `onChange`) and if that is done the `.errors` array will likely end up with duplicates.

If the application developer is to render a list of all errors for the current field/form using `.errors`, they will end up showing the user duplicates. I don't think a user should ever be shown duplicates because it is only one error for them to fix. Right now the application developer would be responsible for implementing deduplication logic and I feel like `.errors` should just do this by default.

If a developer really want the previous behavior they can easily do `Object.values(.errorMap)` but I feel like keeping duplicates is something you should need to opt-in to, as opposed to opt-out like it is right now.

Happy to close this PR if the discussion around changing this behavior decides it's not a great idea.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -467,9 +467,13 @@ export class FieldApi<
         onUpdate: () => {
           const state = this.store.state
 
-          state.meta.errors = Object.values(state.meta.errorMap).filter(
-            (val: unknown) => val !== undefined,
-          )
+          state.meta.errors = [
+            ...new Set(
+              Object.values(state.meta.errorMap).filter(
+                (val: unknown) => val !== undefined,
+              ),
+            ),
+          ]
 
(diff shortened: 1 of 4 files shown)


Balastrong (commented):
Nice catch, having duplicate errors isn't great from an user's perspective.

Could you please tweak the unit tests on this? At least to have one that keeps passing with different errors from different validators, and one that shows how duplicate errors are shown just once.

crutchcorn:
This should be much less of a problem now with the `return anything`. I'm also philosophically opposed to filtering out the errors array any more than we currently are 😅

PR #909, opened 2024-08-17
`canSubmit` with `onMount` validators

oscartbeaumont:
[Stackblitz](https://stackblitz.com/edit/tanstack-form-rms8qz?file=src%2Findex.tsx) reproduction.

Notice how the "Submit" button is not disabled, but the `onMount` validator returned an error (as shown next to the field).

I would expect `canSubmit` to return `false` when an `onMount` error is present instead of `true` like it does right now until the field is touched.

It seems wrong it's possible for the user to press submit without the form being in a valid state.


Balastrong:
Fixed with #726

PR #919, opened 2024-08-21
feat: add isSubmitSuccessful prop

s-coimbra21:
This adds a new prop the form state called `isSubmitSuccessful`. My use case for this is that there is no good way to keep input/submission disabled when a form that calls a server action which redirects the user when the form is successful.

### DIFF
--- a/docs/reference/classes/formapi.md
+++ b/docs/reference/classes/formapi.md
@@ -7,7 +7,7 @@ title: FormApi
 
 # Class: FormApi\<TFormData, TOnMount, TOnChange, TOnChangeAsync, TOnBlur, TOnBlurAsync, TOnSubmit, TOnSubmitAsync, TOnServer, TSubmitMeta\>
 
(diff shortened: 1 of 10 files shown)


crutchcorn (approved):
Awesome work, great value add!

Balastrong:
Sorry for arriving here one month late, @s-coimbra21 what was case where `isSubmitted` wasn't enough and we needed to add `isSubmitSuccessful`?

s-coimbra21:
I have a login form that redirects the user after logging in, sometimes this page load can be very slow (>10s), and in those cases, the submit button would become clickable again whilst the browser is still navigating to the next page

Balastrong:
How is `isSubmitSuccessful` behaving differently from `isSubmitted` in your scenario?

I'm trying to understand because I see both flags and they look exactly the same to me.

s-coimbra21:
If you enter a wrong password you wouldn't want it to block indefinitely

Balastrong:
Can you please make a stackblitz where isSubmitSuccessful works as you expect and isSubmitted doesn't?

PR #925, opened 2024-08-28
[feat] Field validation from form schema

crutchcorn:
This PR introduces the ability to pass a schema to the `validators` object **on a form** and have it validate each specific field:

```jsx
const form = useForm({
  defaultValues: {
    firstName: '',
    lastName: '',
  },
  onSubmit: async ({ value }) => {
    console.log(value)
  },
  validators: {
    onChange: z.object({
      // Errors here are passed to the <form.Field name="firstName"> component
      firstName: z.string().min(3, 'First name must be at least 3 characters'),
      lastName: z.string().min(3, 'Last name must be at least 3 characters'),
    }),
  },
  // Add a validator to support Zod usage in Form and Field
  validatorAdapter: zodValidator({}),
})
```

# TODO

- [x] Update Valibot adapter
- [x] Update Yup adapter
- [x] Fix issues with `createServerValidate`
- [x] Update docs


packages/form-core/src/types.ts
```diff
@@ -2,13 +2,21 @@ import { type DeepKeys } from './util-types'
...
+
 /**
  * If/when TypeScript supports higher-kinded types, this should not be `unknown` anymore
  * @private
  */
 export type Validator<Type, Fn = unknown> = () => {
-  validate(options: { value: Type }, fn: Fn): ValidationError
-  validateAsync(options: { value: Type }, fn: Fn): Promise<ValidationError>
+  validate(
+    options: { value: Type; api: APITypes },
```
Balastrong, line 13: I find `api: APITypes` too generic (apis of what?), maybe something like `validationSource` or `source`?

The idea of passing that value here is great btw :)

packages/zod-form-adapter/src/validator.ts
```diff
@@ -1,27 +1,47 @@
...
+
 export const zodValidator =
   (params: Params = {}): Validator<unknown, ZodType> =>
   () => {
     return {
-      validate({ value }, fn) {
+      validate({ value, api }, fn) {
+        const transformErrors =
+          params.transformErrors ??
+          (api === 'form' ? defaultFormTransformer : mapIssuesToSingleString)
```
Balastrong, line 32: I feel we're defeating the purpose of `params.transformErrors` if used this way, that was to inject a different mapper than the default (now `mapIssuesToSingleString`).
If a user wants to handle everything manually just don't use the adapter, but then they have to manually put errors on each field with this implementation.

It should be either there's no custom transformer and we run `mapIssuesToSingleString` on each field/errorArray or we run `params.transformErrors` if provided. This is the use case when someone doesn't want to join all errors on a single string but wants to handle the array of errors (on each field).

packages/form-core/src/FormApi.ts
```diff
@@ -234,7 +236,7 @@ export type FormState<TFormData> = {
   /**
    * The error map for the form itself.
    */
-  errorMap: ValidationErrorMap
+  errorMap: ServerValidationErrorMap
```
Balastrong, line 239: I'm not fully convinced this should be called `Server` since we're using it client side too, but I couldn't come up with a better name so I'm just leaving a note here :D

Maybe `Detailed` or `Expanded`? I don't know

Balastrong (commented):
Thanks for jumping so fast on expanding adapters to the new validation API :D

IIRC @jljorgenson18 was also looking for this feature, feedback is welcome!

packages/form-core/src/FormApi.ts
crutchcorn: On refection; terrible name. Let's call it `FormValidationErrorMap`

packages/form-core/src/types.ts
crutchcorn: I like `validationSource`!

bompi88:
What is the current status of this PR?

crutchcorn:
@bompi88 I'm... Actively working on it? Like, have made commit less than an hour ago that marked a lot of things off?

We'll get it reviewed soon, but timing just felt weird with "status?" right after I'd worked on it lol

bompi88:
@crutchcorn I'm sorry I did not see that. I agree that is weird 😂

Balastrong (approved):
All three adapters now should be able to inject errors for:
- Primitives
- Nested objects
- Arrays

A couple extra notes:
1. I kinda went bruteforce when adapting each validator's path system into our own and used the exact same approach on all of them. Pros, easier to maintain, cons, there might be a smarter (faster?) way for each validator (e.g. zod has a flatten function, could it help? I don't know)
2. There's quite a lot of repeated code across the validators. Are we ok keeping it like that or refactoring a bit?

Anyway, I don't think these are blockers. We now have the feature working without any breaking change to our API, there's plenty of time to refine the internals later imho.

crutchcorn:
This LGTM @Balastrong - great work! Merging now

bompi88:
@crutchcorn Nice work! I love to test it out 😁 I think the squash commit message was on the wrong format `[feat]` and not `feat:` therefore no release? I might be wrong here

Balastrong:
@bompi88  you're right! :D

I fixed it now, 0.33 should be available 🚀

bompi88:
Thanks @Balastrong ! So far this version works really well🎉

Issue #939, opened 2024-09-05
Add isBlurred state check


Pascalmh:
Add a isBlurred state check so a field can be validated onChange after it has been blurred once.

Usage example once it is implemented:

```ts
validators={{
	onChange: ({ value, fieldApi }) => {
		if (fieldApi.form.getFieldMeta("firstName")?.isBlurred) {
			// validation code goes here
		}

		return undefined;
	},
}}
```

Idea came up in Discord: https://discord.com/channels/719702312431386674/1263506167901065246

PR #940, opened 2024-09-05
fix: only set fields to isBlurred after field blur or form submit

Pascalmh:
this is a followup for [#939](https://github.com/TanStack/form/issues/939)


packages/form-core/src/FormApi.ts
```diff
@@ -562,7 +562,7 @@ export class FormApi<
         }
 
         // If any fields are not blurred
-        if (!field.instance.state.meta.isBlurred) {
+        if (!field.instance.state.meta.isBlurred && cause === 'submit') {
```
crutchcorn, line 565: Per previous previous convos via Discord, let's move this logic out of this block and into the `onChange` function itself so that we don't have to specially edgecase the `cause === 'submit'` behavior here:

https://discord.com/channels/719702312431386674/1263506167901065246/1281270529289883801

crutchcorn (changes requested):
See comment for change required

packages/form-core/src/FormApi.ts
Pascalmh: @crutchcorn I'm not sure where to move this… 

I can't find an `onChange` function. Do I need to create it somewhere?

Pascalmh: I think I figured out what you meant, could you please re-check - thanks!

Pascalmh:
@crutchcorn I implemented the changes as requested and added some more tests, could you please re-check?

packages/form-core/src/FormApi.ts
```diff
@@ -986,6 +974,19 @@ export class FormApi<
...
+      void (
+        Object.values(this.fieldInfo) as FieldInfo<any, TFormValidator>[]
+      ).forEach((field) => {
+        if (!field.instance) return
+
+        if (!field.instance.state.meta.isBlurred) {
+          field.instance.setMeta((prev) => ({ ...prev, isBlurred: true }))
+        }
+      })
+    })
```
Balastrong, line 988: I'd like to push the fix in this PR forward but I'm missing a bit, why are we setting all fields to `isBlurred: true` on submit?

I see blur as an explicit action that is: user focused the form, then moved the focus away. This isn't necessarily happening for all fields on submit.

Pascalmh: My idea was: If the fields are implemented to only show the Error Messages after onBlur and the User submits the Form without blurring all field first the Messages wouldn't show up for those Fields.

**But** that could be handled differently on the field by checking if the form was submitted, that would give us as developers a more fine-grained option when to show the errors.

Balastrong: You could do something like showing the error if `field.isBlurred || form.isSubmitted` right?
Then maybe we can remove this part, what do you think?

Pascalmh:
@Balastrong the isBlurred state of the fields in the form are kept now (removed the `batch` part and updated the test) - 

FYI: you'll need to use `submissionAttempts` instead of `isSubmitted` to display the errors as `isSubmitted` is set to `false` again inside the `handleSubmit`-function:

```ts
// FormApi.ts

handleSubmit = async () => {
    this.store.setState((old) => ({
      ...old,
      // Submission attempts mark the form as not submitted
      isSubmitted: false,
      // Count submission attempts
      submissionAttempts: old.submissionAttempts + 1,
    }))
    
    // …
}
```

### Example Code (JSX):

Only shows errors once the field is blurred or the form has been attempted to submit at least once.

```ts
{(field.form.state.submissionAttempts || field.state.meta.isBlurred) && field.state.meta.errors.length ? (
  <em>{field.state.meta.errors.join(',')}</em>
) : null}
```

### Demo:
https://stackblitz.com/edit/wyqmwdxk?file=src%2Findex.tsx

Balastrong (approved):
That's great, thank you!

PR #947, opened 2024-09-11
minor docs fix

soundrez:
In the React native example, it wasn't clear where `handleChange` came from


Balastrong (approved):
Nice catch, thanks!

Issue #949, opened 2024-09-16
[Feature request] Pass on return value of onSubmit inside handleSubmit


zoodirector:
I would like to pass on the return value of `options.onSubmit()` inside `form.handleSubmit()` to be able to process returned values of the server response. For example the form could be valid but an error occurs when the form should be saved on the server. Or the form is stored successfully and I want to do something with the returned value.

Imagine the following use case: In an administration interface I have a view to create a new item with a form. To save I have two buttons: "Save and return to list view" and "Save and keep editing". Both buttons call `form.handleSubmit()` in their onClick handler. But depending on the result of the function they will continue in a different way. 

I made the according commit here https://github.com/TanStack/form/compare/main...zoodirector:tanstack-form:handle-submit-results

and to illustrate the use case with an example: https://github.com/TanStack/form/compare/main...zoodirector:tanstack-form:handle-submit-results-example

Thanks for looking into it!
Kindly,
Nico

crutchcorn:
Apologies, but we're going to close this issue in favor of the API as documented here: https://github.com/TanStack/form/issues/989

PR #954, opened 2024-09-20
feat: pass on return value of onSubmit inside handleSubmit

zoodirector:
To be able to directly use the results returned by onSubmit the handleSubmit function should pass on the returned value after calling onSubmit. Issue here #949


crutchcorn:
Closing, as implementing this as-is would require a breaking change when we introduce this: https://github.com/TanStack/form/issues/989

PR #971, opened 2024-10-04
Expose Vue specific interfaces

sztyup:
Much like in https://github.com/TanStack/form/pull/892 I'm working on wrappers in Vue for fields and it would be a great improvement if I could access these types.

edit: typo


Balastrong (approved):
Thanks :)

PR #972, opened 2024-10-04
chore: update valibot dependency to v1

fabian-hiller:
I've updated the Valibot dependency on TanStack Form to v1. We've changed the type signature, so this change is necessary. All other changes in our v1 beta version are only internal.

---

An union of schema library authors (Zod, Valibot, ArkType, ...) collaborating on a standard interface for schema libraries called [Standard Schema](https://github.com/standard-schema/standard-schema). This simplifies implementation and prevents vendor lock-in. So far Valibot v1 supports this spec and Zod will follow soon. In the long run, the explicit implementation of a Valibot and Zod adapter may no longer be necessary.

### DIFF
--- a/examples/angular/valibot/package.json
+++ b/examples/angular/valibot/package.json
@@ -22,7 +22,7 @@
     "@tanstack/valibot-form-adapter": "^0.33.0",
     "rxjs": "^7.8.1",
     "tslib": "^2.7.0",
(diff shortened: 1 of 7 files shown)


Balastrong (approved):
Thanks for the update!

Also great idea on the standard schema, that will help a lot ❤️

PR #982, opened 2024-10-17
docs: update react-zod form example defining a zod schema

vicentematus:
Based on the discussion on https://github.com/TanStack/form/discussions/978 with @Balastrong 

This PR updates the docs example on `react/zod` using the most common case which is defining a `Zod` schema at the top level, and then using it on the `useForm` hook.

### DIFF
--- a/examples/react/zod/src/index.tsx
+++ b/examples/react/zod/src/index.tsx
@@ -17,8 +17,10 @@ function FieldInfo({ field }: { field: FieldApi<any, any, any, any> }) {
 }
 
 const userSchema = z.object({
-  firstName: z.string(),
-  lastName: z.string(),
+  firstName: z.string().refine((val) => val !== 'John', {
+    message: '[Form] First name cannot be John',
+  }),
+  lastName: z.string().min(3, '[Form] Last name must be at least 3 characters'),
 })
 
 type User = z.infer<typeof userSchema>
@@ -57,7 +59,7 @@ export default function App() {
             validators={{
               onChange: z
                 .string()
-                .min(3, 'First name must be at least 3 characters'),
+                .min(3, '[Field] First name must be at least 3 characters'),
               onChangeAsyncDebounceMs: 500,
               onChangeAsync: z.string().refine(
                 async (value) => {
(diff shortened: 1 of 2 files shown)


examples/react/zod/src/index.tsx
```diff
@@ -16,18 +16,28 @@ function FieldInfo({ field }: { field: FieldApi<any, any, any, any> }) {
   )
 }
 
+const formSchema = z.object({
+  firstName: z.string(),
+  lastName: z.string(),
+})
+
+type Form = z.infer<typeof formSchema>
```
Pascalmh, line 24: Renaming `formSchema` to e.g. `userSchema` and `Form` to `User` should make this example easier to follow. Otherwise the word "form" is used a lot…

examples/react/zod/src/index.tsx
```diff
@@ -16,18 +16,28 @@ function FieldInfo({ field }: { field: FieldApi<any, any, any, any> }) {
...
+
 export default function App() {
   const form = useForm({
     defaultValues: {
       firstName: '',
       lastName: '',
-    },
+    } as Form,
```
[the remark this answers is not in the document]
Balastrong: I'd probably keep the type assertion in the example for two reasons:
1) It's just one type you have to specify (and you know what type it is) instead of two, with ZodValidator being something you must know/library specific
2) On cases where the defaultValues come from an existing (typed) object you won't need any cast at all

examples/react/zod/src/index.tsx
vicentematus: fixed :)

Pascalmh (approved):

Pascalmh (approved):

Balastrong (approved):
LGMT, thanks!

I added a couple more validations on zod to make it super clear in the example when the error comes from the form validator and when it's from the field validator instead :)

Pascalmh:
@Balastrong what if we had something like this:

```ts
type EmptyStringOr<T> = T | "";

type CheckProperties<T> = {
  [K in keyof T]: EmptyStringOr<T[K]>;
};
```

Either as a type that's shipped with the Library or the `defaultValues` would accept values like according to the `CheckProperties` type?

PR #992, opened 2024-10-24
Update FormApi.ts

b-steel:
## Problem
When updating a field, as soon as the use edits the fields value the `state.meta.isBlurred` becomes `true`.

## Example
https://codesandbox.io/p/devbox/silly-cdn-fzgz2p

## Changes
Change `setFieldValue` to only update `isTouched` and `isDirty` when a field value changes.  
Leave `isBlurred` unchanged so that it's only updated on blur.


b-steel:
I just saw that there's already PR's in for this

PR #995, opened 2024-10-28
fix: dont merge arrays in `mergeForm`

rburgst:
- this causes form submissions to be duplicated when having an array field in your form

Example of the problem: https://stackblitz.com/edit/tanstack-form-vdj4yh?file=src%2Fapp%2Fshared-code.ts

Just try to submit the form and watch the todos to be duplicated


rburgst:
not sure why codecov/project is still failing

Balastrong (approved):
I'm not 100% sure this is the right direction but also I can't think of a better one to prevent duplication so... let's go this way! Thanks for the fix :)

PR #997, opened 2024-10-28
fix: Valibot peer dependency to support dist-tags

fabian-hiller:
Looks like the npm install broke with the upgrade to Valibot v1. This PR should fix that.


Balastrong (approved):
Thanks!

Issue #902, opened 2024-08-12
Accessing form error array using form.useStore and destructuring can cause react update depth issue


craigbehnke:
### Describe the bug

When using destructuring, the form.useStore hook returns an error array that appears to change on every render. This then causes a react update depth issue.

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-form-oryfsj?file=README.md

### Steps to reproduce

1. Access the errors array using destructuring, like so:
```ts
 const {errors}: {errors: ValidationError[]} = form.useStore(state => ({
    errors: state.errors
  }))
```
2. Use the errors array as a dependency to a react hook, such as useEffect or useMemo
3. Observe that this react hook is perpetually called.

Please see the provided repro.

### Expected behavior

As a user, I want the error array to be updated only when the error array changes.

### How often does this bug happen?

Every time

### Platform

- OS: MacOS
- Browser: Chrome/Safari
- Typescript/React

### TanStack Form adapter

react-form

### TanStack Form version

v0.29.1

### TypeScript version

5.5.4

jackcoup:
I also get this, almost all the time I use `form.useStore((state) => state.values)`  really not ideal.

SiddOnKeys:
did u find a fix or any way around it, its causing issues for me and I'm on a deadline lol

jackcoup:
I believe I used `form.useField()` for each prop

Balastrong:
I'm sorry but `const errors = form.useStore((state) => state.errors)` seems to be working fine, I think I don't understand where's the problem.

What feature are you trying to implement?

craigbehnke:
@Balastrong It's been a bit, but what I was trying to achieve was to have 1 hook that would return the errors and other properties, i.e.
```ts
// My one hook to rule them all
const {isValid, isPristine, isSubmitted, isSubmitting,errors} = form.useStore(
        state => ({
            canSubmit: state.canSubmit,
            isPristine: state.isPristine,
            isSubmitted: state.isSubmitted,
            isSubmitting: state.isSubmitting,
            errors: state.errors
        })
    )
```

Because of this bug (and after a extended debugging session) I had to split it into 2 different hook calls so that it would work:
```ts
// The main hook
const {isValid, isPristine, isSubmitted, isSubmitting} = form.useStore(
        state => ({
            canSubmit: state.canSubmit,
            isPristine: state.isPristine,
            isSubmitted: state.isSubmitted,
            isSubmitting: state.isSubmitting
        })
[...]

PR #1038, opened 2024-11-25
TanStack Store Internals Refactor

crutchcorn:
> This PR should be merged with the message
>
> ```
> fix: useStore should cause fewer (if any) update depth issues in React
> ```

# TODOs

- [x] Move `new Derived` side effects to dedicated `mount` method
- [x] Add back `transformArray`
- [x] Add back broken tests
- [x] Add tests to validate fix for #902
    - Make sure that `errors` doesn't change unless `errorMap` has changed
- [ ] Make derived state more modular to avoid problems with over-generating related code


crutchcorn:
OK, I figured out what the problem was:

`batch` sends out `prevDepVals` in one single update. Meaning that when we end up receiving the new value, it will inevitably be `===` consistent. By removing all instances in `form` for `batch` by doing:

```typescript
function batch(fn: () => void) {
  return fn()
}
```

It works as-intended and _all_ tests pass.

This isn't ideal tho, so I'll try to patch it in `store` instead if possible

packages/form-core/src/FormApi.ts
```diff
@@ -393,103 +387,172 @@ export class FormApi<
...
-              prev.push(curr.form)
+        const isValidating = !!isFieldsValidating
+
+        // As `errors` is not a primitive, we need to aggressively persist the same referencial value for performance reasons
+        let errors = prevVal?.errors ?? []
+        if (
+          !prevBaseStore ||
+          currBaseStore.errorMap !== prevBaseStore.errorMap
+        ) {
+          errors = Object.values(currBaseStore.errorMap).reduce(
```
Balastrong, line 490: Can we check earlier if `onMount` errors have to be removed? So that we avoid pushing them now and then filtering them out on lines 515-521

Something like moving up this block
```ts
let errorMap = currBaseStore.errorMap
if (shouldInvalidateOnMount) {
    errorMap = Object.assign(errorMap, { onMount: undefined })
}
```
and then using the local `errorMap` on `Object.values`

Balastrong (commented):
I see there are also some failing tests on Solid which seem to be related to the initial form state, I couldn't see if it's something on the solid-form adapter or even from solid-store

packages/form-core/src/FormApi.ts
crutchcorn: Let's make a follow-up PR on this, seeing that it works today and I'm struggling to think of an easy way to move this around ATM

Balastrong: Sure!

Balastrong (approved):

PR #1068, opened 2024-12-12
chore: allow unusedLocals in standard-schema example

mmkal:
there were a bunch ts-ignores, which looks bad on an example

Personally I prefer policing unused locals via a linter, which helps assign more appropriate "seriousness" to tsc errors, and so unused local errors can be more easily suppressed, but that's just me so I didn't make a sweeping change here.

### DIFF
--- a/examples/react/standard-schema/src/index.tsx
+++ b/examples/react/standard-schema/src/index.tsx
@@ -17,7 +17,6 @@ function FieldInfo({ field }: { field: FieldApi<any, any, any, any> }) {
   )
 }
 
-// @ts-ignore - Might be unused for demo purposes
 const ZodSchema = z.object({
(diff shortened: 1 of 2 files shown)


Balastrong (commented):
I don't think the ts-ignores look bad in the example, in fact we're intentionally writing _wrong_ code with unused variables for demo purposes.

Another option I evaluated was to keep all the unused stuff commented, but then it meant you had to uncomment both the schema and the import each time, not ideal.

With that said, I see it might cause some confusion so maybe allowing the error at a linter level might make the example easier to read but still, we're allowing an error.

I'll think about it, the example is just about showing how easy it's to swap them, I might merge this. Thanks for the suggestion!

PR #1074, opened 2024-12-13
fix: simplify field validator type

Balastrong:
Adding StandardSchema removed type inference on validators with plan functions, the type might be simplified and seems to also fix the issue.

Note: type tests as the following were still passing on the repo, while the exact same configuration was failing on the examples in the docs, I still don't know why.


Balastrong:
Superseded by #1076

PR #1075, opened 2024-12-14
Comply with `react/no-children-prop` ESLint rule

rinarakaki:
Let’s do this before v1.


Pascalmh:
This should be a stylistics choice. It was discussed before in the discord-channel: https://discord.com/channels/719702312431386674/1278341822883823626/1278342479879340146

@crutchcorn said:

`<Comp children={() => null} />`

is the same as:

`<Comp>{() => null}</Comp>`

> Just more shortform preference for us. It's a pretty minor change that doesn't impact speed or anything else

-----

Thad said: my personal preference would be to set the Eslint rule to `allowFunctions` (see https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-children-prop.md#allowfunctions)

crutchcorn:
+1 to what @Pascalmh said. Nothing to add on my end. Closing per stylistic decision.

rinarakaki:
Sorry, but

> Just more shortform preference for us

means @crutchcorn prefers this change, right?

crutchcorn:
More or less.

PR #1076, opened 2024-12-14
fix: remove dependency from standard schema and internalize it

Balastrong:
Attempt number 2 fixing the missing types in validators


lnhrdt:
Thank you @Balastrong! Fixed several type errors in our projects that came up in recent versions.

Balastrong:
That's great, thanks for the feedback!

PR #1077, opened 2024-12-16
test: Standard Validator should handle multiple errors on a field

Pascalmh:
As the ZodAdapter is getting deprecated I refactored the failing test to go against the Standard Validator.

see: #943


Balastrong:
Returning multiple errors means dealing with arrays instead of simple strings pretty much everywhere. I included this in #1090 and the test seems to be green :)

Issue #999, opened 2024-10-30
[BUG] When using zodValidator, even if I set a valid value with setFieldValue, canSubmit does not become true


koki-iwaizumi:
### Describe the bug

When using zodValidator, even if I set a valid value with setFieldValue, canSubmit does not become true

```
import { zodValidator } from '@tanstack/zod-form-adapter'
import { z } from 'zod'

// ...

const formSchema = z.object({
  age:   z.string().optional().refine((data) => data != null && data !== '' && !isNaN(Number(data))),
  ageSub:   z.string().optional().refine((data) => data != null && data !== '' && !isNaN(Number(data))),
})

const form = useForm({
  validatorAdapter: zodValidator(),
  validators: {
    onChange: formSchema
  },
})

const onChangeAge = (handleChange) => (event: React.ChangeEvent<HTMLInputElement>) => {
  handleChange(event.target.value)
  form.setFieldValue('ageSub', event.target.value)  // The validation for ageSub is not being executed here."
}

return (
  <form>
      <form.Field name="age">
        {(field): React.ReactNode => {
          return (
            <Input
              type="number"
              onChange={(e): void => onChangeAge(field.handleChange)}
              value={field.state.value?.toString()}
            />
          )
        }}
    </form.Field>
    <form.Subscribe
      selector={(state): { isDisabled: boolean } => {
        return { isDisabled: state.isPristine || !state.canSubmit }
      }}
    >
      {({ isDisabled }): React.ReactNode => {
        return (
          <Button isDisabled={isDisabled} >
            submit
          </Button>
        )
      }}
    </form.Subscribe>
  </form>
)
```

In version 0.32.0, canSubmit became true with valid values (e.g., '5' or '6'), but in version 0.33.0, canSubmit remains false.

### Your minimal, reproducible example

none

### Steps to reproduce

none

### Expected behavior

I want validation to be executed when setFieldValue is called.

### How often does this bug happen?

Every time

### Platform

- OS: Mac
- Browser: Chrome

### TanStack Form adapter

react-form

### TanStack Form version

v0.33.0

### TypeScript version

v5.6.3

### Additional context

If I execute validateField after every setFieldValue call, the validation runs and behaves as expected. However, this was not necessary in version 0.32.0. I would like to know if this behavior is intentional or a bug.

like this.

```
import { zodValidator } from '@tanstack/zod-form-adapter'
import { z } from 'zod'

// ...

const formSchema = z.object({
[...]

PR #1082, opened 2024-12-18
[WIP] fix: errors not resetting once values are correct for Zod

Gambitboy:
This is an attempt to fix the issue raise in https://github.com/TanStack/form/issues/999
I had a similar issue I think is linked to this.

This passes my test I added but it breaks other tests.
I'm not sure if I'm either handling it incorrectly or if fixing this for zod causes other things to break.

Please let me know if i'm heading in the right direction.

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -750,7 +750,7 @@ export class FormApi<
         } else {
           for (const [field] of Object.entries(this.fieldInfo)) {
             this.setFieldMeta(field as DeepKeys<TFormData>, (prev) => {
-              const newErrorMap = prev.errorMap;
+              const newErrorMap = prev.errorMap
 
               if (prev.isDirty) {
                 for (const [errorKey] of Object.entries(prev.errorMap)) {
@@ -898,7 +898,7 @@ export class FormApi<
           } else {
             for (const [field] of Object.entries(this.fieldInfo)) {
               this.setFieldMeta(field as DeepKeys<TFormData>, (prev) => {
-                const newErrorMap = prev.errorMap;
+                const newErrorMap = prev.errorMap
 
                 if (prev.isDirty) {
                   for (const [errorKey] of Object.entries(prev.errorMap)) {
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -747,6 +747,26 @@ export class FormApi<
               }))
             }
           }
+        } else {
+          for (const [field] of Object.entries(this.fieldInfo)) {
+            this.setFieldMeta(field as DeepKeys<TFormData>, (prev) => {
+              const newErrorMap = prev.errorMap;
+
+              if (prev.isDirty) {
+                for (const [errorKey] of Object.entries(prev.errorMap)) {
+                  newErrorMap[errorKey as ValidationErrorMapKeys] = undefined
+                }
+              }
+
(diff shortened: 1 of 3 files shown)


packages/form-core/src/FormApi.ts
```diff
@@ -747,6 +747,26 @@ export class FormApi<
               }))
             }
           }
+        } else {
```
Balastrong, line 750: Here the form validator should clear the field errors only if the validator is one that returns the object version of `FormValidationError`.
This means you should go with `else if (isFormValidationError(rawError))`.

However, when the form has no errors the adapters are returning `undefined`. With this information alone you don't know if it was a form object validator or a simple validator returning a string, as in that case you shouldn't clear fields (the new failing tests you're getting) but only the form error.

The first approach that comes to my mind is to slightly refactor the adapters and make them return something instead of undefined. It could be `{ fields: {} }`

With this information you know that the form validator returned no errors and you should clear the fields.

Note: this might break the initial `if (fieldErrors)` but something like `if (Object.values(fieldErrors).length)` should do the trick.

packages/form-core/src/FormApi.ts
```diff
@@ -747,6 +747,26 @@ export class FormApi<
...
+          for (const [field] of Object.entries(this.fieldInfo)) {
+            this.setFieldMeta(field as DeepKeys<TFormData>, (prev) => {
+              const newErrorMap = prev.errorMap
+
+              if (prev.isDirty) {
+                for (const [errorKey] of Object.entries(prev.errorMap)) {
+                  newErrorMap[errorKey as ValidationErrorMapKeys] = undefined
+                }
+              }
```
Balastrong, line 759: Are those lines still needed or were they left by some other attempts?

packages/form-core/src/FormApi.ts
```diff
@@ -747,6 +747,26 @@ export class FormApi<
               }))
             }
           }
+        } else {
+          for (const [field] of Object.entries(this.fieldInfo)) {
```
Balastrong, line 751: Tiny thing, `Object.keys` might be enough here

packages/form-core/src/FormApi.ts
```diff
@@ -875,7 +895,28 @@ export class FormApi<
                 }))
               }
             }
+          } else {
```
Balastrong, line 898: Comments for sync also apply here, but let's make sync work first, then we'll replicate in the async :)

packages/zod-form-adapter/tests/FormApi.spec.ts
```diff
@@ -184,4 +184,55 @@ describe('zod form api', () => {
...
+      },
+    })
+    field2.mount()
+
+    field1.setValue('password')
+    expect(field1.getMeta().errors).toStrictEqual(['Passwords must match'])
+    expect(form.state.canSubmit).toBe(false)
+
+    field2.setValue('password')
+    expect(field2.getMeta().errors).toStrictEqual([])
```
Balastrong, line 235: This is always going to be empty, you should check for `field1` errors

Balastrong (commented):
You're on the right track but this might be a bit more tricky than I initially thought. I added some notes, let me know what you think.

Thank you!

crutchcorn:
This should be fixed as of 0.45.1, but thank you for giving this a shot!

Gambitboy:
Pleasure, glad it's fixed now.

PR #1084, opened 2024-12-20
feat(docs): reusable fields & reusable field utility type

harry-whorlow:
A continue on from #636, this PR introduces docs for reusable components/fields as well as introduces a Ts utility type for inferring valid field names, for the relevant component. 

Much like my last PR this is something I was trying to do for a work project, but not having any docs or examples hindered the process, so I figured I'd give it a go.

---
**Guidance wanted:**

Specifically with this line `onChange={(e) => handleChange(e.target.value as any)}` (line 28 in reusable-fields) I can't stand the casting of any, however I can figure a way for Ts to infer this value. Any suggestions is greatly appreciated!

---
**Sandbox**

I have a code sandbox for playing around with this [here](https://codesandbox.io/p/sandbox/yv7zmh), you'll see theres a folder called solutions that has two files: 
- one.tsx works and is a refined version of all my test components 
- two.tsx is another version that I haven't quite got working but in terms of user Ux I think is an improvement. 

The rest of the files can be ignored but if you want to see how I got here they kind of explain my reasoning.

### DIFF
--- a/docs/framework/react/guides/reusable-fields.md
+++ b/docs/framework/react/guides/reusable-fields.md
@@ -17,8 +17,8 @@ export default function GenericTextField<
   TName extends InferValidFormKeys<TForm, string>,
   TFormValidator extends Validator<TForm, unknown> | undefined,
 >({ name, form }: {
-   name: TName;
-   form: ReturnType<typeof useForm<TForm, TFormValidator>
+  name: TName;
+  form: ReturnType<typeof useForm<TForm, TFormValidator>
 > }): JSX.Element {
   return (
     <form.Field name={name}>
@@ -65,8 +65,8 @@ export default function GenericTextField<
   TName extends InferValidFormKeys<TForm, string>,
   TFormValidator extends Validator<TForm, unknown> | undefined,
 >({ name, form }: {
-   name: TName;
-   form: ReturnType<typeof useForm<TForm, TFormValidator>
+  name: TName;
+  form: ReturnType<typeof useForm<TForm, TFormValidator>
 > }): JSX.Element {
   return (
     <form.Field name={name}>
@@ -83,7 +83,7 @@ export default function GenericTextField<
 
 function App() {
   const form = useForm({
-    defaultValues: { name: '', id: 0 },
+    defaultValues: { name: '', id: 0, interests: {hobbies: 'climbing'} },
     onSubmit: ({ value }) => {
       console.log(value);
     },
--- a/docs/framework/react/guides/reusable-fields.md
+++ b/docs/framework/react/guides/reusable-fields.md
@@ -1,0 +1,94 @@
+---
+id: reusable-fields
+title: Creating reusable fields
+---
+
+As TanStack form is a headless library, we provide you the core building blocks and then give you the freedom to build on top of them. This page introduces the concept of creating reusable field-components for your form, allowing you to create fields that you can reuse throughout your app.
+
+## Basic Usage
+
+To create a reusable fields, you can do the following.
+
+```tsx
+import {  useForm, Validator, InferValidFormKeys } from '@tanstack/react-form';
+
+export default function GenericTextField<
+  TForm,
+  TName extends InferValidFormKeys<TForm, string>,
+  TFormValidator extends Validator<TForm, unknown> | undefined,
+>({ name, form }: {
+   name: TName;
+   form: ReturnType<typeof useForm<TForm, TFormValidator>
+> }): JSX.Element {
+  return (
+    <form.Field name={name}>
+      {({ handleChange, handleBlur, state }) => (
+        <input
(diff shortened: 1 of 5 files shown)


packages/react-form/src/types.ts
```diff
@@ -28,3 +29,12 @@ export type UseFieldOptions<
...
 }
+
+/**
+ * The return type use useForm with pre-populated generics
+ */
+export type UseFormReturnType<TForm, TFormValidator> = TFormValidator extends
+  | Validator<TForm, unknown>
+  | undefined
+  ? ReturnType<typeof useForm<TForm, TFormValidator>>
+  : never
```
harry-whorlow, line 40: I'm not certain this is required, but it cleans up the types on the components. Let me know what you think 🤟

harry-whorlow:
So I was playing around with my utility types and fixed [two.tsx](https://codesandbox.io/p/sandbox/yv7zmh?file=%2Fsrc%2Fsolutions%2FTwo.tsx%3A26%2C50). 

However, I'm at a bit of a crossroads’s, my utility type abstracts aways some of the Ts so that the interface is cleaner which I think the average dev would be appreciated. However, you lose some of the error feedback.

with the current implementation the interface looks like this:

`export default function GenericTextField<
  TForm,
  TFormValidator,
  TName extends InferValidFields<TForm, string>,
>({
  name,
  form: { Field },
}: {
  name: TName;
  form: FormReturnType<TForm, TFormValidator>;
})`

which can be simplified to this with the updated utility: 

`export default function GenericTextField<TForm, TName, TFormValidator>({
  name,
  form: { Field },
}: {
  name: InferValidFieldsName<TName, TForm, string>;
  form: FormReturnType<TForm, TFormValidator>;
})`

For me, I don't mind the Ts, but my concern is for people that aren’t too we'll versed in it. What do you guys think?

packages/react-form/src/types.ts
harry-whorlow: ah, this Isn't needed this can just be simplified on the user to  ReactFormExtendedApi<TForm, any>;

packages/react-form/src/types.ts
```diff
@@ -28,3 +28,12 @@ export type UseFieldOptions<
 > & {
   mode?: 'value' | 'array'
 }
+
+/**
+ * The return type use useForm with pre-populated generics
+ */
+export type UseFormReturnType<TForm, TFormValidator> = TFormValidator extends
```
harry-whorlow, line 35: I would love to get rid of the TFormValidator, since TFormValidator is just Validator<TForm, unknown> | undefined. But so far my efforts have been fruitless.

docs/framework/react/guides/reusable-fields.md
```diff
@@ -0,0 +1,92 @@
...
+>({ name, form }: {
+  name: InferValidFormKeys<TForm, string>;
+  form: UseFormReturnType<TForm, TFormValidator>,
+> }): JSX.Element {
+  return (
+    <form.Field name={name}>
+      {({ handleChange, handleBlur, state }) => (
+        <input
+          value={state.value}
+          onChange={(e) => handleChange(e.target.value as any)}
```
harry-whorlow, line 27: Another improvement would be to get Ts to infer the handleChange value so the update is type safe.

Talent30:
Might be a dup with https://github.com/TanStack/form/pull/825 ?

harry-whorlow:
@Talent30 yeah, I was talking with crutchcorn about this on Friday and he made me aware of it. He said to leave it open for now, and he'll link it to his issue. Though, I'm not sure how useful it'll be, as after reading though the linked Pr, they've run into all the same problems I have. 🙃

bzbetty:
wonder if it'd be possible to do something like ```form.Field``` as a way to provide your components.

```
   function useMyForm(args) {
       const form = useForm(args);
       return {
          ...form
         TextField: function ({ name }) {
              return (
                  <form.Field name={name}>
                        {({ handleChange, handleBlur, state }) => (
                          <input
                            value={state.value}
                            onChange={(e) => handleChange(e.target.value as any)}
                            onBlur={handleBlur}
                          />
                        )}
                  </form.Field>
              );
          }
       }
   }
```

harry-whorlow:
@bzbetty I think the issue with the proposed solution is that if the form is apart of the component then your re-instantiating the form on every component. So instead of having one form controlling the fields, you will have multiple standalone forms controlling a single field then passing the data back to whatever the args is. 

After seeing the proposed solutions in #825, and after banging my head against the keyboard for a couple weeks, I think its probably the best way to go. Though I will admit if you could get the form typing your solution would probably address the internal typing issues I ran into in my attempt.

crutchcorn:
Closing in favor of an alternative API we're proposing here: #1179

PR #1090, opened 2024-12-29
feat: return an array of errors from a single validator (custom & schema library)

Balastrong:
This PR addresses the frequent request of being able to return multiple errors from a single validator.

Validator functions (aka what you put inside `validators: { onChange: () => ... }`) can now return a `ValidationResult` that is `undefined | false | null | ValidationError | ValidationError[]` with `ValidationError` being a `string`.

Existing validators returning a single string will keep working, additionally you can now also return an array of strings.

As a side effect, validators defined with schema libraries will no longer be cut to the first error by default, but all errors will be shown instead.

### Breaking Changes

Being able to return an array of strings from a validator means `errorMap.onChange` & co will no longer be a string but an array.

Documentation has been updated showing the quickest change to keep the previous behaviour:

```diff
- {meta.errorMap.onChange ? (<em>{meta.errorMap.onChange}</em>) : null}
+ {meta.errorMap.onChange?.length ? (<em>{meta.errorMap.onChange[0]}</em>) : null}
```

In case you were using `meta.errors` to display all your errors there won't be any breaking change since this value will always be a (flat) aggregate array of errors.

Feedback is welcome!


crutchcorn:
Closing this in favor of https://github.com/TanStack/form/pull/1104, which offers more flexibility over how types are handled.

This PR led to the idea I had for that PR tho! @Balastrong continues to be a legend! 🚀

Issue #1122, opened 2025-01-21
UseStore - Maximum update depth exceeded when selector is not present


harry-whorlow:
### Describe the bug

The useStore hook, when not provided the "selector" prop results in a "Maximum update depth exceeded" error. Though the interface shows the prop is optional, when not provided it results is the aforementioned render loop error.

Apparently this is a know issue and was supposed to of been resolved, but for me at least, its reared its head again.

_Interface with option "selector" prop_
![Image](https://github.com/user-attachments/assets/6c7f6de8-e1de-41f9-a832-d1e105b113ad)

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/yv7zmh?file=%2Fsrc%2FStoreTest%2FWithStore%2Findex.tsx

### Steps to reproduce

- Using the latest version of form
- Create a basic form app (or visit my Codesandbox demo)
- Instantiate the useStore hook and provide it the form store, however leave the selector prop undefined. 
- This will trigger "Maximum update depth exceeded" error.

### Expected behavior

Maximum update depth exceeded shouldn’t happen.

### How often does this bug happen?

Every time

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/1fc7d2d1-df4a-4367-97b0-31aa2e8cc233)
![Image](https://github.com/user-attachments/assets/68988fd4-41ec-46a0-8ac8-d68aa74e720d)

### Platform

OS: MacOs
Browser: Chrome
Version: Version 130.0.6723.70 (Official Build) (arm64)

Also CodeSandbox

### TanStack Form adapter

react-form

### TanStack Form version

0.41.2

### TypeScript version

8.18.1

### Additional context

I should note that every instance of useStore throughout the tanstack/form repo has been provided a selector, perhaps a test to ensure that a "selectorless" (its now a word) useForm don't result in an render loop would be beneficial.

crutchcorn:
Keep in mind, despite the fix in #1123, **you should always have a selector for `useStore` when possible**. Not having one can cause **far** more rerenders than expected, even with the patch

harry-whorlow:
Fastest patch in the west. Thanks! 🤟

Perhaps, it would be good to have this noted in the docs somewhere, as the only documentation on this hook I can find is this

![Image](https://github.com/user-attachments/assets/e8de0318-86fb-4cab-8347-b7c6bc8fded1)

and its not exactly clear that leaving it out can cause re-renders. As always I'm happy to draft a Pr, but it might be not warranted.

[...]

PR #1123, opened 2025-01-21
fix: useStore without a selector should no longer crash in React

crutchcorn:
Fixes: https://github.com/TanStack/form/issues/1122


rburgst:
This causes https://github.com/TanStack/form/issues/1135

dufia:
I think this is still an issue on the latest version

Issue #1111, opened 2025-01-13
Meta api for form fields


harry-whorlow:
Continuation from the discussion #709. [Listeners](https://github.com/TanStack/form/pull/1032) are implemented, however the meta field is yet to be started. 

The proposal is to have the meta data available to the field component: 

_(example shamelessly ripped form #709)_

`<form.Field
  name="foo"
  meta={({ fieldApi }) => ({
    disabled: fieldApi.form.getFieldValue('someOtherField') == null,
    ...
  })}
  // or maybe expose values directly like form.Subscribe
  meta={({ values }) => ({
    disabled: values.someOtherField == null,
    ...
  })}
>
  {(field, meta) => <Comp
    disabled={meta.disabled}
    ... `

You can subscribe to the meta from the Form but its pretty handy to have it available to the Field component.

I'm happy to give this a go, I just need the go ahead from a maintainer. 🤟

fulopkovacs:
Hey 👋 , this sounds like a great idea. Personally, I prefer the second version (less verbose and more straight-forward). Are you still ready to create a PR? If yes, then this is the "go ahead"! ☺️☺️

harry-whorlow:
Yeah, me too. 

Absolutely, I'll start it this week, and hopefully have a PR ready by next weekend. 🫡

Thanks for the go ahead 🤘

fulopkovacs:
Thank *you* for your enthusiasm! 🙌  Feel free to ping me in the comments when the PR is ready!

fulopkovacs:
> [@fulopkovacs](https://github.com/fulopkovacs), l've left a couple questions on [#1125](https://github.com/TanStack/form/pull/1125). If you get a chance to take a look it would be appreciated 🤟

Awesome, thanks for your work!  I'll look at them tomorrow, this week was pretty busy...

itayperry:
✨ Hi everyone, 

I commented on a post by @crutchcorn on LinkedIn and we talked about disabling form-fields. I was told it is okay to have a discussion about it here as a part of this issue.

I have this request/idea that I would like to suggest, a `setDisabled()` method for form-fields.
**I don't know if it's a good idea**.. but I had this problem where I was building a complicated form using `react-hook-form`, I had a field containing data about a heart's rhythm disorder, when you pick one you must `setValue()` and also disable other fields, for example: a patient's pulse value.
[...]

PR #1125, opened 2025-01-22
feat(core): field meta API

harry-whorlow:
**Field.meta**

From issue #1111 and discussion #709, this pr Introduces the Field.meta api to allow for extensible and easily accessible meta data to the field component.

*example api*
```tsx
<form.Field
  name="foo"
  meta={({ values }) => ({
    disabled: values.someOtherField == null,
    ...
  })}
>
  {(field, meta) => 
  <Comp
    disabled={meta.disabled}
    ... >
  </Comp>
```

---
**Tasks**

- [x] Meta available to Field.children 
- [ ] MetaFn available to form.Field  
- [ ] Tests 
- [ ] LGTM 🚀


harry-whorlow:
@fulopkovacs PR draft

packages/form-core/src/FieldApi.ts
```diff
@@ -661,6 +685,8 @@ export class FieldApi<
 
   getMeta = () => this.store.state.meta
 
+  meta = this.store.state.meta
```
harry-whorlow, line 688: Is there a reason that the getMeta is a function to be called, as opposed to directly exposed like I've done here?

packages/form-core/src/FieldApi.ts
```diff
@@ -396,10 +410,15 @@ export type FieldMetaDerived = {
   isPristine: boolean
 }
 
+export type FieldMetaExtension<TMetaExtension extends object = {}> =
```
harry-whorlow, line 413: The idea here is to have a optional generic so that the meta field is TypeSafe when accessing the additional meta information.

fulopkovacs: Hmmm, maybe I'm haven't thought about this issue enough, but I don't see a reason why we need the `FieldMetaExtension` type. 🤔 

For example, when we use it here:
```ts
export type FieldMeta<TMetaExtension extends object = {}> = FieldMetaBase &
  FieldMetaDerived &
  FieldMetaExtension<TMetaExtension>
```
it looks redundant:
```ts
export type FieldMeta<TMetaExtension extends object = {}> = FieldMetaBase &
  FieldMetaDerived & TMetaExtension
```

If I'm not missing something obvious (I might be 😆 ), then can we start without using `FieldMetaExtension`, and introduce it only when it seems necessary?

fulopkovacs (commented):
Thanks for the work, I think I answered all the questions in the first round! ☺️

packages/form-core/src/FieldApi.ts
harry-whorlow: yeah that makes sense.... I'm still living in react land where I just expect the re-render to update the values. Thanks!

packages/form-core/src/FieldApi.ts
harry-whorlow: I would agree 🤟

harry-whorlow:
Morning @fulopkovacs! 

So as you can see from the previous commit, I've gotten a working MVP, it's working in Js land, I'm still trying to figure out why Ts isn't inferring the meta generic.

Screenshot 2025-01-29 at 10 01 57

If you have any ideas shoot me a message, if your busy I'll continue to poke away at this issue. But if I can get this working then it should be ready relatively shortly for review. 🤟

packages/form-core/tests/FieldApi.spec.ts
```diff
@@ -1760,4 +1760,32 @@ describe('field api', () => {
...
+    })
+    form.mount()
+
+    const nameField = new FieldApi({
+      form,
+      name: 'name',
+      meta: ({ values }) => ({
+        dinosaur: values.name === 'Stegosaurus' ? 'dino' : 'notDino',
+      }),
+    })
```
harry-whorlow, line 1778: Where Ts is throwing the incorrectly inferred generic, previously mentioned

packages/form-core/src/FieldApi.ts
```diff
@@ -656,10 +682,15 @@ export class FieldApi<
       fieldApi: this,
     })
 
+    this.setMeta((prev) => ({
+      ...prev,
+      ...this.options.meta?.(this.form.state),
+    }))
+
```
harry-whorlow, line 689: Update meta, I've added to mount and setValue... Can you think of anywhere else it should be updated?

harry-whorlow:
Looking for insight... Should the fields meta be inferable from form? or should it be a "local" meta? 

Currently implementation is as follows

_(taken from my test)_
Screenshot 2025-01-29 at 11 02 25

Looking at formApi if it should be global, I would imagine it'll be a lot of work. So I figured I'd get conformation first, before attempting it, though it might be beyond my capabilities as theres a lot of internal stuff I'm not familiar with. 🤟

Either way I'll give it a shot.

harry-whorlow:
@fulopkovacs, I'm going to open a new branch after the v1 update.... jumping through all the conflicts is breaking my brain, I'll ping you when it's ready

fulopkovacs:
> @fulopkovacs, I'm going to open a new branch after the v1 update.... jumping through all the conflicts is breaking my brain, I'll ping you when it's ready
> 
> I'll leave this open for now as it makes it easier to see what I did. I'll be sure to close it when the new branch is open for cr 🤟

Awesome, thanks for the note! Sorry for not being active in this thread for so long, I'll make sure to pay more attention to the new branch! 😅 I can do a review when you ping me if you want (or answer any questions).

harry-whorlow:
hey @fulopkovacs check out #1216, if you’re about I'd really appreciate some input. I've kinda hit a wall figuring out how to infer the user defined field meta from the form and could really use someone to bounce some ideas off.

PR #1126, opened 2025-01-22
feat(docs): risks of omitting useStore selector

harry-whorlow:
From issue #1122 and a conversation with @crutchcorn, omitting the selector prop of the useStore hook will result in unnecessary re-renders.

Screenshot 2025-01-22 at 15 06 57

---

The interface of useStore shows it as an optional prop, and while the hook will work without the prop, it is not the recommended approach.

```tsx
export declare function useStore<TState, TSelected = NoInfer<TState>>(store: Store<TState, any>, selector?: (state: NoInfer<TState>) => TSelected): TSelected;
```

---

I've searched the Docs for references to this, and the only documentation relating to this hook was under `Reactivity` of `basic concepts`. However, it does not highlight the risks of omitting selector.

Screenshot 2025-01-22 at 15 13 36

The only other references to this hook, I found in the documentation, were examples demonstrating the use of `useStore`. However, none of them highlighted this potential pitfall.

This PR adds clarification of useStore and examples of correct use.

Screenshot 2025-01-22 at 15 07 07

### DIFF
--- a/docs/framework/react/guides/basic-concepts.md
+++ b/docs/framework/react/guides/basic-concepts.md
@@ -204,6 +204,16 @@ const firstName = useStore(form.store, (state) => state.values.firstName)
 />
 ```
 
+It is important to remember that while the `useStore` hook's `selector` prop is optional, it is strongly recommended to provide one, as omitting it will result in unnecessary re-renders.
+
+```tsx
+// Correct use
+const firstName = useStore(form.store, (state) => state.values.firstName)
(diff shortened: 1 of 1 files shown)


harry-whorlow:
Personally I feel it might be good to have a section for the library's hooks in the docs, with documentation on their use and explanation of their props, kind of like the api-reference but a bit more human friendly.

But, for now this is a good stop gap. 🤟

crutchcorn (approved):

Issue #989, opened 2024-10-22
[FEATURE] Form callback API


crutchcorn:
Imagine a usecase where the user wants two buttons, they both send the item to the backend (after form validation) but one goes back to the list and the other goes to a different page, like a step 2 or a detail page.

To solve this, we're thinking of introducing a new API called `onSubmitMeta` that allows you to pass arguments to the `onSubmit` function via a `meta` field.

This API might look something like this:

```tsx
  const form = useForm({
    defaultValues: {
      firstName: '',
      lastName: '',
    },
    // Is 'never' by default unless you provide a type
    // {} is the default value passed to `onSubmit`'s `meta` property
    onSubmitMeta: {} as { someValue: string },
    onSubmit: async ({ value, meta }) => {
      // Do something with form data
      console.log(value)
      // Do something with the values passed via handleSubmit
      console.log(meta)
    },
  })

  return (
    <div>
      <h1>Simple Form Example</h1>
      <form
        onSubmit={(e) => {
          e.preventDefault()
          e.stopPropagation()
          form.handleSubmit({
            someValue: 'value',
          })
        }}
      >
```

By having `onSubmitMeta` it enables us to have a few features on top of passing properties such as:

- Type safety
- Default arguments to `meta` field from the FW itself

This API should be purely additive, so it may not make it into v1 of TanStack Form

thejasviES:
Hi @crutchcorn  can I be assigned this ticket?

zoodirector:
Hey @crutchcorn . Thank you very much for taking up my ideas of #949 ! However, I don't find this API proposal of yours particularly advantageous. It would require me to put the business logic that handles the server response (e.g. navigating to different routes after submissions) **inside** the `onSubmit` function of the form. It is an inversion of control where not necessary. For me it would feel more natural if the form API would return control after the successful submission of the form and not wrap around the business logic that follows afterwards.  I think that my suggested design in #949 would allow for a better decoupling of code. For example if I want to reuse the same redirection logic in different forms.

But that is just my opinion. I guess I could also live with the API suggested above.

godd9170:
I've got this exact use case, and I was able to solve it by looking at the id of the submitting button.

```jsx
() => {
  let form = useForm({
    defaultValues: {},
    onSubmit: async ({ value }) => await post(value),
  });
  return (
    <form
      onSubmit={async (event) => {
        event.preventDefault();
        event.stopPropagation();
        await form.handleSubmit();
[...]

PR #1131, opened 2025-01-28
feat(core) onSubmit meta

harry-whorlow:
_reopened in https://github.com/TanStack/form/pull/1167_

From issue #989, this PR introduces onSubmitMeta to the formApi. 

_Example API shamelessly ripped from #989 and @crutchcorn_ 
```
const form = useForm({
    defaultValues: {
      firstName: '',
      lastName: '',
    },
    // Is 'never' by default unless you provide a type
    // {} is the default value passed to `onSubmit`'s `meta` property
    onSubmitMeta: {} as { someValue: string },
    onSubmit: async ({ value, meta }) => {
      console.log(value)
      // Do something with the values passed via handleSubmit
      console.log(meta)
    },
  })

  return (
    <div>
      <h1>Simple Form Example</h1>
      <form
        onSubmit={(e) => {
          e.preventDefault()
          e.stopPropagation()
          form.handleSubmit({
            someValue: 'value',
          })
        }}
      >
```

---
Screenshot 2025-01-28 at 10 17 44

---
- [x] working overloads
- [x] Tests


packages/form-core/src/FormApi.ts
```diff
@@ -1102,7 +1120,18 @@ export class FormApi<
...
+          value: this.state.values,
+          formApi: this,
+          meta: submitMeta,
+        })
+      } else {
+        await this.options.onSubmit?.({
+          value: this.state.values,
+          formApi: this,
+        })
+      }
```
harry-whorlow, line 1134: I've run into this wall where 
```  onSubmit?: (
    props: {
      value: TFormData
      formApi: FormApi<TFormData, TFormValidator>
    } & SubmitMeta<TFormSubmitMeta>,
  ) => any | Promise<any>```
  
  However Ts can't infer that as TFormSubmitMeta is not undefined that meta should be of TFormSubmitMeta, and vice versa in the else block

File: packages/form-core/src/FormApi.ts [hunk 2]
```diff
@@ -177,17 +181,26 @@ export interface FormOptions<
...
-    formApi: FormApi<TFormData, TFormValidator>
-  }) => any | Promise<any>
+  onSubmit?: (
+    props: {
+      value: TFormData
+      formApi: FormApi<TFormData, TFormValidator, TFormSubmitMeta>
+    } & TFormSubmitMeta extends object
+      ? { meta: TFormSubmitMeta }
+      : {},
+  ) => any | Promise<any>
```
harry-whorlow, line 191: This is the type causing me issues, no matter how I do the conditional it'll throw a Ts error on the onSubmit props.

packages/form-core/src/FormApi.ts
```diff
@@ -1102,7 +1122,12 @@ export class FormApi<
 
     try {
       // Run the submit code
-      await this.options.onSubmit?.({ value: this.state.values, formApi: this })
+
+      await this.options.onSubmit?.({
+        value: this.state.values,
+        formApi: this,
+        ...(submitMeta ? { meta: submitMeta } : {}),
+      } as any)
```
harry-whorlow, line 1130: I'm not sure this casting is acceptable, however i cant find a workaround. Open to suggestions 🤟

packages/form-core/src/FormApi.ts
```diff
@@ -632,7 +646,9 @@ export class FormApi<
       ) as never
     }
 
-    return (props.validate as FormValidateFn<any, any>)(props.value) as never
+    return (props.validate as FormValidateFn<any, any>)(
+      props.value as any,
```
harry-whorlow, line 650: Same goes here, either I specify on line 623 `formApi: FormApi<any, any, any>` and this errors or i specify it as `formApi: FormApi<any, any, never>` and it'll error below.

As always I'm open to suggestions.

harry-whorlow:
So I've gone down the rabbit hole of trying to get the adapters working and starting with react I'm a little lost as to why the onSubmit: ({value}) is inferred as never, when it's outside the conditional of the onSubmit interface. Perhaps someone with a little more familiarity could have a look, as I'm not sure what else to adjust.

harry-whorlow:
I'm going to give it a go as the generic not extending object, I'm a little unsure why I started with it like this. But, I think restricting the user to what type they can define is a miss-step.

packages/lit-form/src/tanstack-form-controller.ts
```diff
@@ -137,9 +149,10 @@ export class TanStackFormController<
         TName,
         TFieldValidator,
         TFormValidator,
-        TData
+        TData,
+        TParentMetaExtension
       >
-    )(this.api, fieldConfig, render)
+    )(this.api as any, fieldConfig, render)
```
harry-whorlow, line 155: is this acceptable?

crutchcorn: Let's do `as never` if we can

packages/form-core/src/FormApi.ts
crutchcorn: Casting is a-okay in internal code. `any` and `never` are not banned unless it's a boundary type

PR #1152, opened 2025-02-16
feat(core): validateFields method for FormApi

lucas-dtrn:
I habe added the `validateFields()` method to the FormApi so you can validate multiple fields at once.


crutchcorn:
At first glance: "I'm not opposed to this" but then I thought about it longer and I **really** strongly want to discourage users from using `validate` and similar APIs manually. Can you provide a usecase (and minimal repro) for what this would do in an application?

Thanks!

LeCarbonator:
This PR has been stale for multiple months. Until a proper use case and minimal repro is found, I'll mark it as closed.

PR #1155, opened 2025-02-18
feat(core): add getAllErrors to form api

harry-whorlow:
Add's a utility function to form.api to return form errors combined with a reduced fieldMeta.errorMap. 

example

```tsx
form.getAllErrors() => {form: FormErrors, fields: Record<string, FieldErrors>}
```

---
Tasks

- [x] utility
- [x] tests


Balastrong (changes requested):
Thanks for the great PR!

I think we're on the right direction but we can probably make it even better. Here I see you're returning the flat `errors` array for the form but the `validationMap` for the fields. We should probably return _both for both_.

What do you think about a return type similar to that:
```ts
{
  form: { errors: ValidationError[]; errorMap: ValidationErrorMap }
  fields: Record<
    DeepKeys<TFormData>,
    {
      errors: ValidationError[]
      errorMap: ValidationErrorMap
    }
  >
}
```

Feel free to make the types smarter if you think it's needed, the overall idea is to return both errors and errorMap for form and each field :)

packages/form-core/src/FormApi.ts
```diff
@@ -1431,6 +1431,54 @@ export class FormApi<
...
+        },
+        {} as Record<ValidationErrorMapKeys, ValidationError>,
+      )
+    }
+    console.log(this.state.errors)
+
+    return {
+      form: {
+        errors: this.state.errors,
+        errorMap: reduceErrorMap(this.state.errorMap as ValidationErrorMap),
```
harry-whorlow, line 1456: @Balastrong where dose FormValidationErrorMap used as `Record<ValidationErrorMapKeys, FormValidationError>`? 

Screenshot 2025-02-20 at 14 18 12

Im guessing when the form validation returns an object of validations.

Balastrong: Right, that type is gonna change a bit in #1104

In this case we should probably wait for the other PR to land first

harry-whorlow: yeah, and increase the type complexity quite some too 😂

harry-whorlow:
> This also somewhat ties into my PR where we need to determine which field errors came from form validations vs field validations. I think if we're getting all errors in a function like this, it'd be nice to know where the error came from as well. If this could be included within this functionally we could also use it internally to clear stale field errors sourced from the form.

you mean in the form validations? 

It should work as form FormValidationError is a union of either

`
const err = {
  onChange: ValidationError,
  onMount: ValidationError,
  //...
}

const err = {
  // name and age being fields in form
  onChange: { name: ValidationError, age: ValidationError },
  onMount: { name: ValidationError, age: ValidationError },
  //...
}
`

Screenshot 2025-02-21 at 08 34 01

packages/form-core/src/FormApi.ts
harry-whorlow: @Balastrong, correct me if I'm work... but with the changes to the errors will this have been resolved?

Balastrong: Yep, that ambiguity is gone :)

packages/form-core/src/FormApi.ts
```diff
@@ -1865,6 +1865,55 @@ export class FormApi<
...
+            acc[fieldName] = {
+              errors: (fieldMeta as AnyFieldMeta).errors,
+              errorMap: fieldErrors,
+            }
+          }
+
+          return acc
+        },
+        {} as Record<
+          string,
```
Balastrong, line 1911: Do you think we can make this `string` become `DeepKeys<TFormData>`?

packages/form-core/src/FormApi.ts
```diff
@@ -1865,6 +1865,55 @@ export class FormApi<
...
+        (errorAcc, [errorName, errorValue]) => {
+          if (errorValue != null) {
+            errorAcc[errorName as ValidationErrorMapKeys] = errorValue
+          }
+
+          return errorAcc
+        },
+        {} as Record<ValidationErrorMapKeys, ValidationError>,
+      )
+    }
```
Balastrong, line 1884: I'm not entirely sure about this step, do we really need it?

Balastrong (commented):
Well done! I think we're almost there :)

packages/form-core/src/FormApi.ts
harry-whorlow: yeah, we can do... that was the intention but I forgot to add it.

packages/form-core/src/FormApi.ts
harry-whorlow: The reducing of errors? in terms of simplicity I'm all for it... but if your calling getAllErrors and you're returning `onChange: undefined` I'm not not so convinced. But I'm happy to change it.


Screenshot 2025-02-23 at 17 55 31

Balastrong: I agree that only fields with errors should be returned and this is already done nicely.

This piece of code is not filtering fields but it's filtering the errorMap keys on each field if I'm not mistaken, but this would produce an inconsistent result compared to accessing the errorMap directly from `field.state.meta.errorMap` in other parts of the code.

harry-whorlow: ah okay..... my bad, sorry for the misunderstanding, yeah if we return the full error map then this can be a lot simpler

packages/form-core/src/FormApi.ts
harry-whorlow: @Balastrong so, I haven’t seen this before.... but typing the index of the object causes this 
Screenshot 2025-02-24 at 08 13 29

Do you have any ideas?

harry-whorlow: ah, awesome... thats something that wouldn’t of crossed my mind.

Balastrong (approved):
LGTM!

Do you want to add a few more unit tests to cover edge cases and also some type tests? Otherwise we can just merge and add them later, let me know :)

harry-whorlow:
@Balastrong as always it's been a pleasure. 

Normally, I would add the tests but I've got three open pr's and it's getting hard to track and keep them up to date, plus I'm away at the moment. So if we can merge it and I'll come back to the tests in two weeks or so. 
🤟

Balastrong:
Oki, no worries, I added a type test :)

Thanks again, merging it now 🚀

PR #1157, opened 2025-02-19
fix(form-core): clear previous invalid sync validation errors on FormApi

juanvilladev:
Fixes: https://github.com/TanStack/form/issues/999

When performing sync validation, ensure that previous field errors are cleared if they are no longer present in the new validation result. This prevents stale validation errors from persisting in the form state.

To achieve this I have added a new prevFieldsErrorMap:

```TypeScript
  /**
   * @private fields previously validated via form level validations
   */
  prevFieldsErrorMap: FieldsErrorMapFromValidator<TFormData> = {}
```

This allows us to keep track of field level errors that were added at the form level.
We then use this map to clear them out if no new errors appear for the same field.

```TypeScript
        for (const field of Object.keys(this.prevFieldsErrorMap) as Array<
          DeepKeys<TFormData>
        >) {
          const fieldMeta = this.getFieldMeta(field)
          if (
            fieldMeta?.errorMap[errorMapKey] &&
            !newFieldsErrorMap[field]?.[errorMapKey]
          ) {
              ...prev,
              errorMap: {
                ...prev.errorMap,
                [errorMapKey]: undefined,
              },
            }))
          }
        }
```
All tests are passing. The tricky part was keeping track of what field errors were added from the form level. 
Added a test covering this specific scenario within FormApi.spec.ts


harry-whorlow:
So you'll see under this comment this warning:

Screenshot 2025-02-19 at 22 15 44

I'm going to go out on a limb and guess, that because I've contributed before, my account has been approved for workflows. So I presume until a maintainer is available for review that the workflows will not run until approved.

But you can check it will pass before by running the `test:ci` script which consists of: 
- pnpm sherif
- pnpm test:knip
- pnpm test:eslint
- pnpm test:lib
- pnpm test:types
- pnpm test:build
- pnpm build

If in the future you want to figure it out yourself you check this out by heading into the .github/workflows/ci.yml and seeing what test scrip they call, in this case `test:ci`.

Screenshot 2025-02-19 at 22 31 27


Then head to the package.json and calling that script, or the individual parts.

Screenshot 2025-02-19 at 22 30 44

juanvilladev:
Thanks! @harry-whorlow that makes sense. Cool, I just ran test:ci and got a successful pass:
` NX   Successfully ran targets test:sherif, test:knip, test:eslint, test:lib, test:types, test:build, build for 39 projects (1m)`

harry-whorlow:
Just be aware that now the ci is running, it'll add a commit to your branch. 

This commit will contain generated docs and type definitions. So git pull before you push, again as your remote will be one commit ahead from your local. 

Good luck with the pr 🫡

juanvilladev:
Another approach I thought of was to update the errorMap type to store the validation source. This would allow us to know which errors should be clearable at the form level once stale. This would be quite a breaking change though so unsure...

Something like changing the current type below to:
```TypeScript
export type ValidationErrorMap = {
  [K in ValidationErrorMapKeys]?: ValidationError
}
```

```TypeScript
export type ValidationErrorMap = {
  [K in ValidationErrorMapKeys]?: { error: ValidationError, source: ValidationSource }
}
```

We already have a type for Validation Source [here.](https://github.com/TanStack/form/blob/c00898499b26f4f886f9abf109b64a88b18b423a/packages/form-core/src/types.ts#L3)

Since @crutchcorn is already reworking the ValidationErrors maybe something like this could be considered, else I'm not sure how else to know which errors on the map came from the form and should be cleared.

Balastrong (approved):

juanvilladev:
Hey @Balastrong, just checking in—let me know if there's anything else I need to do for this PR to be merged, or if it's just a matter of waiting. Appreciate your time!

PR #1167, opened 2025-02-21
feat(core): onSubmit meta 

harry-whorlow:
_new branch after from https://github.com/TanStack/form/pull/1131_

From issue #989, this PR introduces onSubmitMeta to the formApi. 

_Example API shamelessly ripped from #989 and @crutchcorn_ 
```
const form = useForm({
    defaultValues: {
      firstName: '',
      lastName: '',
    },
    // Is 'never' by default unless you provide a type
    // {} is the default value passed to `onSubmit`'s `meta` property
    onSubmitMeta: {} as { someValue: string },
    onSubmit: async ({ value, meta }) => {
      console.log(value)
      // Do something with the values passed via handleSubmit
      console.log(meta)
    },
  })

  return (
    <div>
      <h1>Simple Form Example</h1>
      <form
        onSubmit={(e) => {
          e.preventDefault()
          e.stopPropagation()
          form.handleSubmit({
            someValue: 'value',
          })
        }}
      >
```

---
Screenshot 2025-01-28 at 10 17 44

---
- [x] working overloads
- [x] Tests


packages/form-core/src/FormApi.ts
```diff
@@ -281,23 +287,49 @@ export interface FormOptions<
...
-      TOnChangeAsync,
-      TOnBlur,
-      TOnBlurAsync,
-      TOnSubmit,
-      TOnSubmitAsync,
-      TOnServer
-    >
-  }) => any | Promise<any>
+  onSubmit?: (
+    props: [TSubmitMeta] extends [never]
```
crutchcorn, line 300: Do we need to do this `extends [never]` or could we just have:

```typescript
          meta: TSubmitMeta
```

Always there?

crutchcorn (commented):
Great work!!

Just checking, do we want to make these changes? WDYT?

packages/form-core/src/FormApi.ts
```diff
@@ -1536,7 +1590,11 @@ export class FormApi<
 
     try {
       // Run the submit code
-      await this.options.onSubmit?.({ value: this.state.values, formApi: this })
+      await this.options.onSubmit?.({
+        value: this.state.values,
+        formApi: this,
+        ...(submitMeta ? { meta: submitMeta } : {}),
```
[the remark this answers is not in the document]
harry-whorlow: yeah... That makes sense actually

packages/form-core/src/FormApi.ts
harry-whorlow: yeah, I think I over engineered it... I've made the change

crutchcorn (approved):
**Incredible** work here, @harry-whorlow - sincerely this is great. Let's get this in!

harry-whorlow:
@crutchcorn thanks! looking forward to getting this functionality into my works codebase.

 Yeah I will do, just give give a couple days and I'll get around to it 🤟. I'm thinking of adding a page 'submission handling' but having just the meta might leave it feeling a bit bare, is there anything you think I could also add to it?

PR #1190, opened 2025-02-28
errors.map((error) => error.message) before .join(', ')

rinarakaki:


crutchcorn (changes requested):
This is only needed when using standard schema, not for other validation

PR #1204, opened 2025-03-03
docs(form-core): Update philosophy heading tag levels h1 to h2.md

KevinVandy:
Unless there was a reason for h1s, I just converted headings to h2 to be consistent with the rest of the tanstack docs


Balastrong (approved):
Yeah I think they should be h2 as we do in the other docs pages.
It also seems h1 don't have much space on top doing a weird effect if used inside the page.

PR #1214, opened 2025-03-04
docs: improve get started guide for React to be more representative

crutchcorn:
Per a conversation I had with @phryneas, this "get started" guide for React might serve as a better jumping off point for users.

### DIFF
--- a/docs/framework/react/quick-start.md
+++ b/docs/framework/react/quick-start.md
@@ -3,54 +3,132 @@ id: quick-start
 title: Quick Start
 ---
 
-The bare minimum to get started with TanStack Form is to create a form and add a field. Keep in mind that this example does not include any validation or error handling... yet.
+TanStack Form is unlike most form libraries you've used before. It's designed for large-scale production usage, with a focus on type safety, performance, and developer experience.
+
+As a result, we've developed [a philosophy around the library's usage](/form/latest/docs/philosophy) that values scalability and long-term developer experience over short and sharable code snippets.
+
+Here's an example of a form following many of our best practices, which will allow you to rapidly develop even high-complexity forms after a short onboarding experience:
+
+```tsx
+import React from 'react'
+import ReactDOM from 'react-dom/client'
+import { createFormHook, createFormHookContexts } from '@tanstack/react-form'
+import { TextField, NumberField, SubmitButton } from '~our-app/ui-library'
+// We also support Valibot, ArkType, and any other standard schema library
+import { z } from 'zod'
+
+const { fieldContext, useFieldContext, formContext, useFormContext } =
+  createFormHookContexts()
+
+// Allow us to bind components to the form to keep type safety but reduce production boilerplate
+const { useAppForm, withForm } = createFormHook({
+  fieldComponents: {
+    TextField,
+    NumberField,
+  },
+  formComponents: {
+    SubmitButton,
+  },
+  fieldContext,
+  formContext,
+})
+
+const PeoplePage = () => {
+  const form = useAppForm({
+    defaultValues: {
+      username: '',
+      age: 0,
+    },
+    validators: {
+      // Pass a schema or function to validate
+      onChange: z.object({
+        username: z.string(),
+        age: z.number().min(13),
+      }),
+    },
+    onSubmit: ({ value }) => {
+      // Do something with form data
+      alert(JSON.stringify(value, null, 2))
+    },
+  })
+
+  return (
+    <form
+      onSubmit={(e) => {
+        e.preventDefault()
+        form.handleSubmit()
+      }}
+    >
+      <h1>Personal Information</h1>
+      {/* Components are bound to `form` and `field` to ensure extreme type safety */}
+      <form.AppField
+        name="username"
+        children={(field) => <field.TextField label="Full Name" />}
+      />
+      {/* The "name" property will throw a TypeScript error if typo'd  */}
+      <form.AppField
+        name="age"
+        children={(field) => <field.NumberField label="Age" />}
+      />
+      {/* Our customized hook enables usage of `form.AppField` and `form.AppForm` components */}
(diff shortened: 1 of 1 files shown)


docs/framework/react/quick-start.md
```diff
@@ -3,54 +3,132 @@ id: quick-start
...
+As a result, we've developed [a philosophy around the library's usage](/form/latest/docs/philosophy) that values scalability and long-term developer experience over short and sharable code snippets.
+
+Here's an example of a form following many of our best practices, which will allow you to rapidly develop even high-complexity forms after a short onboarding experience:
 
 ```tsx
 import React from 'react'
 import ReactDOM from 'react-dom/client'
-import { useForm } from '@tanstack/react-form'
+import { createFormHook, createFormHookContexts } from '@tanstack/react-form'
+import { TextField, NumberField, SubmitButton } from '~our-app/ui-library'
```
phryneas, line 16: I'd like to see one of those as an example (can you do multiple tabs in code fields?) or at least a link/mention towards the composition page to explain what these would look like.

crutchcorn: No tabs for our docs site, unfortunately, but ya fair enough. I'll link out to the composition page

docs/framework/react/quick-start.md
```diff
@@ -3,54 +3,132 @@ id: quick-start
...
+// We also support Valibot, ArkType, and any other standard schema library
+import { z } from 'zod'
 
-export default function App() {
-  const form = useForm({
+const { fieldContext, useFieldContext, formContext, useFormContext } =
```
phryneas, line 20: `useFieldContext` and `useFormContext` are not used here - would they normally be used? If yes, does it fit the example? If no, maybe it might make sense to skip them here?

docs/framework/react/quick-start.md
```diff
@@ -3,54 +3,132 @@ id: quick-start
...
+  createFormHookContexts()
+
+// Allow us to bind components to the form to keep type safety but reduce production boilerplate
+const { useAppForm, withForm } = createFormHook({
```
phryneas, line 24: Same for `withForm` here

crutchcorn: Fixed in newest commit

docs/framework/react/quick-start.md
```diff
@@ -3,54 +3,132 @@ id: quick-start
...
+        e.preventDefault()
+        form.handleSubmit()
+      }}
+    >
+      <h1>Personal Information</h1>
+      {/* Components are bound to `form` and `field` to ensure extreme type safety */}
+      <form.AppField
+        name="username"
+        children={(field) => <field.TextField label="Full Name" />}
+      />
```
phryneas, line 67:
```suggestion
      <form.AppField
        name="username"
      >
         {(field) => <field.TextField label="Full Name" />}
      </form.AppField>
```

Probably a style preference, but I tend to feel that a `children` prop might confuse users as opposed to nesting.

crutchcorn: Yeah, it's a style preference. We'd have to change all of our docs to update the style and `children` is less verbose

docs/framework/react/quick-start.md
```diff
@@ -3,54 +3,132 @@ id: quick-start
...
+      // Do something with form data
+      alert(JSON.stringify(value, null, 2));
+    },
+  });
+
+  return (
+    <form.Field
+      name="age"
+      validators={{
+        // We can choose between form-wide and field-specific validators
```
phryneas, line 108: I think this comment is a bit stray here since it doesn't show two different ways of doing things.

crutchcorn: In comparison to the last example it does - just wanted to clarify for those who were looking between the two confused

docs/framework/react/quick-start.md
```diff
@@ -3,54 +3,132 @@ id: quick-start
 title: Quick Start
 ---
 
-The bare minimum to get started with TanStack Form is to create a form and add a field. Keep in mind that this example does not include any validation or error handling... yet.
+TanStack Form is unlike most form libraries you've used before. It's designed for large-scale production usage, with a focus on type safety, performance, and developer experience.
```
Balastrong, line 6: In the "focus" list I'd probably also mention something like "composability" or "reusable components" or... something in general that hints about the fact that we're giving an API to configure and define the form once and then just assemble the components on all the forms around the application.

Sure it's shown later, but I see it as a huge feature and better defines how we support large-scale

docs/framework/react/quick-start.md
```diff
@@ -3,54 +3,132 @@ id: quick-start
...
+import { TextField, NumberField, SubmitButton } from '~our-app/ui-library'
+// We also support Valibot, ArkType, and any other standard schema library
+import { z } from 'zod'
 
-export default function App() {
-  const form = useForm({
+const { fieldContext, formContext } = createFormHookContexts()
+
+// Allow us to bind components to the form to keep type safety but reduce production boilerplate
+const { useAppForm } = createFormHook({
```
Balastrong, line 24: It's probably worth mentioning that you don't need to run createFormHook with all the parameters for each form instance but most likely it's done once for the entire app.

docs/framework/react/quick-start.md
```diff
@@ -3,54 +3,132 @@ id: quick-start
...
+      <form.AppField
+        name="username"
+        children={(field) => <field.TextField label="Full Name" />}
+      />
+      {/* The "name" property will throw a TypeScript error if typo'd  */}
+      <form.AppField
+        name="age"
+        children={(field) => <field.NumberField label="Age" />}
+      />
+      {/* Our customized hook enables usage of `form.AppField` and `form.AppForm` components */}
```
Balastrong, line 73: We can add here a couple extra words to explain why AppField and AppForm are two different components

PR #1216, opened 2025-03-04
Meta api 2

harry-whorlow:
*reopened from #1125*

**Field.meta**

From issue #1111 and discussion #709, this pr Introduces the Field.meta api to allow for extensible and easily accessible meta data to the field component.

*example api*
```tsx
<form.Field
  name="foo"
  meta={({ values }) => ({
    disabled: values.someOtherField == null,
    ...
  })}
>
  {(field, meta) => 
  <Comp
    disabled={meta.disabled}
    ... >
  </Comp>
```

---
**Tasks**

- [x] Meta available to Field.children 
- [ ] MetaFn available to form.Field  
- [ ] Tests 
- [ ] LGTM 🚀

### DIFF
--- a/packages/form-core/tests/FieldApi.spec.ts
+++ b/packages/form-core/tests/FieldApi.spec.ts
@@ -1850,4 +1850,27 @@ describe('field api', () => {
     expect(field.getMeta().errors).toStrictEqual([])
     expect(form.state.canSubmit).toBe(true)
   })
+
+  it('should have user defined meta and react to value change', () => {
+    const form = new FormApi({
+      defaultValues: {
+        name: 'Stegosaurus',
+      },
+    })
+    form.mount()
+
+    const nameField = new FieldApi({
+      form,
+      name: 'name',
+      meta: ({ values }) => ({
+        dinosaur: values.name === 'Stegosaurus' ? 'dino' : 'notDino',
+      }),
+    })
+
+    nameField.mount()
+    expect(nameField.getMeta().dinosaur).toEqual('dino')
+
+    nameField.handleChange('Cat')
+    expect(nameField.getMeta().dinosaur).toEqual('notDino')
+  })
 })
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -14,6 +14,7 @@ import type {
   FieldInfo,
   FormApi,
   FormAsyncValidateOrFn,
+  FormState,
   FormValidateAsyncFn,
   FormValidateFn,
   FormValidateOrFn,
@@ -77,6 +78,7 @@ export type FieldValidateFn<
     TData,
     // This is technically an edge-type; which we try to keep non-`any`, but in this case
     // It's referring to an inaccessible type from the field validate function inner types, so it's not a big deal
+    any,
     any,
     any,
     any,
@@ -176,6 +178,7 @@ export type FieldValidateAsyncFn<
     any,
     any,
     any,
+    any,
     any
   >
   signal: AbortSignal
@@ -259,9 +262,38 @@ export type FieldListenerFn<
     any,
     any,
     any,
+    any,
     any
   >
 }) => void
+
+/**
+ * @private
+ */
+export type FieldMetaFn<
+  TParentData,
+  TFormOnMount extends undefined | FormValidateOrFn<TParentData>,
+  TFormOnChange extends undefined | FormValidateOrFn<TParentData>,
+  TFormOnChangeAsync extends undefined | FormAsyncValidateOrFn<TParentData>,
+  TFormOnBlur extends undefined | FormValidateOrFn<TParentData>,
+  TFormOnBlurAsync extends undefined | FormAsyncValidateOrFn<TParentData>,
+  TFormOnSubmit extends undefined | FormValidateOrFn<TParentData>,
+  TFormOnSubmitAsync extends undefined | FormAsyncValidateOrFn<TParentData>,
+  TFormOnServer extends undefined | FormAsyncValidateOrFn<TParentData>,
+  TFieldMetaExtension extends object,
+> = (
+  props: FormState<
+    TParentData,
+    TFormOnMount,
+    TFormOnChange,
+    TFormOnChangeAsync,
+    TFormOnBlur,
+    TFormOnBlurAsync,
+    TFormOnSubmit,
+    TFormOnSubmitAsync,
+    TFormOnServer
+  >,
+) => TFieldMetaExtension
 
 export interface FieldValidators<
   TParentData,
@@ -375,6 +407,15 @@ export interface FieldOptions<
   TOnSubmitAsync extends
     | undefined
     | FieldAsyncValidateOrFn<TParentData, TName, TData>,
+  TFormOnMount extends undefined | FormValidateOrFn<TParentData>,
+  TFormOnChange extends undefined | FormValidateOrFn<TParentData>,
+  TFormOnChangeAsync extends undefined | FormAsyncValidateOrFn<TParentData>,
+  TFormOnBlur extends undefined | FormValidateOrFn<TParentData>,
+  TFormOnBlurAsync extends undefined | FormAsyncValidateOrFn<TParentData>,
+  TFormOnSubmit extends undefined | FormValidateOrFn<TParentData>,
+  TFormOnSubmitAsync extends undefined | FormAsyncValidateOrFn<TParentData>,
+  TFormOnServer extends undefined | FormAsyncValidateOrFn<TParentData>,
+  TFieldMetaExtension extends object,
 > {
   /**
    * The field name. The type will be `DeepKeys<TParentData>` to ensure your name is a deep key of the parent dataset.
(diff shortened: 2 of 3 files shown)


packages/form-core/tests/FieldApi.spec.ts
```diff
@@ -1850,4 +1850,27 @@ describe('field api', () => {
...
+        dinosaur: values.name === 'Stegosaurus' ? 'dino' : 'notDino',
+      }),
+    })
+
+    nameField.mount()
+    expect(nameField.getMeta().dinosaur).toEqual('dino')
+
+    nameField.handleChange('Cat')
+    expect(nameField.getMeta().dinosaur).toEqual('notDino')
+  })
```
harry-whorlow, line 1875: @fulopkovacs 

so this works nicely, but the issue is if I do something like `form.state.fieldMeta.name` I can't infer the user defined meta. `

Do you have any suggestions on how to get it to form? I've tried a few things but can't get it to work. 🙃

fulopkovacs: Ugh, this does seem to be complex problem. I spent about 2 hours thinking about it, but couldn't come up with a solution.

## Here's what I see now

Up until this point we had no need to infer the type of a field's `meta`, because it was the same for all of them (`isDirty`, `errorMap`, etc.). Because of this, it was enough to create a _generic_ `AnyFieldMeta` type that we could use in the methods of `FormApi`, like `getFieldMeta`:

```ts
  getFieldMeta = <TField extends DeepKeys<TFormData>>(
    field: TField,
  ): AnyFieldMeta | undefined => {
    return this.state.fieldMeta[field]
  }
```

Compare this to the `getFieldValue` method, where the return type depends on the _specific_ field whose value we're looking for:

```ts
  getFieldValue = <TField extends DeepKeys<TFormData>>(
    field: TField,
  ): DeepValue<TFormData, TField> => getBy(this.state.values, field)
```

`DeepValue` receives the name of the field (`TField`) as a type argument and uses it to get the type of the returned value from the form's data (`TFormData`).

I think `DeepValue` has access to the type of the field's value, because you initiate the `FormApi` class with it: 

```ts
    const form = new FormApi({
      defaultValues: {
        name: 'Stegosaurus',
      } // this form can have only 1 field whose name must be `"name"` 
    })
    
    const formWithAge = new FormApi({
      defaultValues: {
        name: 'Stegosaurus',
      } as {name: string, age?: number} // this form can have an `age` field too
    })
    
    const hobbyField = new FieldApi({
      form,
      name: 'hobby', // throws an error, because it's not a valid field name
    })
```

## How to solve this problem?

In order to have a proper `FormFieldMeta` type instead of a generic `AnyFieldMeta` type in the `FormApi` class, we need to define the field's `meta` types during the initialization of the `FormApi` class.

So it could look something like this:

```ts
    const formWithAge = new FormApi({
      defaultValues: {
        name: 'Stegosaurus',
      },
      fieldMeta: {
       name: ({ values }) => ({
        dinosaur: values.name === 'Stegosaurus' ? 'dino' : 'notDino',
      })
    })
```

Ofc, it would be better to only pass down the types of the field's custom meta properties, and define those later.

```ts
    const formWithAge = new FormApi({
      defaultValues: {
        name: 'Stegosaurus',
      },
      fieldMeta: {} as {name: {dinosaur: 'dino' | 'notDino'})
    })
```

Honestly, this is not the best possible API we could have, but the best one I could come up with at 0:37 🤣...

If the `FormApi` is aware of the possible meta fields mapped to field names, then we can have a type like `FormFieldMeta<TFormFieldMetaData, TField>`, where `TFormFieldMetaData` is the type of `fieldMeta` in the example above, and `TField` is the name of the field.

## Bad news

I'll be away from the keyboard for 4 days from today (maybe I can be still reached today, but definitely not from tomorrow). 😢 You probably have more questions/comments, feel free to write them, and I'll answer when I'm back!

harry-whorlow: @fulopkovacs I was thinking the exact same thing, I think there was a small dwindling light of hope that perhaps it was just my inability to code. Maybe I've just sold myself short...

Given you're suggested answer would it also make sense to have a baseMetaExtension too? I ask this because of #715 where a proposed solution was to store a Ref in the meta. Which makes sense to me, but having to add the Ref meta to every field, especially if you have a lot of fields could result in lots of boilerplate.

I'll have a poke around in the mean time, enjoy the time away from the keyboard 🤟

fulopkovacs: >  Maybe I've just sold myself short...

You probably did! 🤣 

> Given you're suggested answer would it also make sense to have a `baseMetaExtension` too?

🤔 Hmmm, do you image the API look like this 👇 ?

```ts
    const formWithAge = new FormApi({
      defaultValues: {
        name: 'Stegosaurus',
        age: 0,
      },
      fieldMeta: {
          name: ({ values }) => ({
            dinosaur: values.name === 'Stegosaurus' ? 'dino' : 'notDino',
          }),
       },
      baseMetaExtension: {
         ref: null as null | HTMLInputElement
      }
    })
```

After reading the discussion you mentionted (https://github.com/TanStack/form/discussions/715), I understand why we would need something like this, and I support it, but I still have a weird feeling about it for some reason. Maybe I'm just not fond of the name `baseMetaExtension`. What do you think about something like `defaultCustomFieldMeta`? From what I understand now, we would have to assign default values to the shared custom field meta properties when we create the form, so a name with `default` in it sounds fitting, plus it looks similar to `defaultValues`.

harry-whorlow: @fulopkovacs sounds good to me, if it's cool with you I'll spin the defaultCustomFieldMeta into its own separate PR and get that out as it seems #715 is asked every other week.

harry-whorlow: @fulopkovacs actually... you've got a good point it would. we shall keep it as one then... and yes I will drop it into the discord.

fulopkovacs:
@harry-whorlow 

Heyy 👋 ,

I'd just like to check in to see if I can help you with anything regarding this PR. ☺️ If you're stuck, please let me know, I'd love to help! :))

harry-whorlow:
@fulopkovacs 

Leonardo suggested this:

>Speaking of typescript, have you tried doing something similar to what we're doing with validators? With two generics on FieldApi, one for the meta coming from form and the other with the meta coming from that field

and thinking about it this could be actually exactly what were looking for. I'll try and throw something together 😄

iceikiru:
I think I have a use case that might be similar to this

I have a need for every field in some forms to know the same info (in this case so they can disable and show info about the permissions they're missing)

```
const newValidationSchema = z.object({
  sectionComplete: z.boolean(),
  foregroundColor: rgbColorSchema(),
  backgroundColor: rgbColorSchema(),
  logoColor: rgbColorSchema(),
});

export function VirtualCardArtView() {
  const { user } = useUserContext();
  const isBank = user?.role === EnumUserTenantType.BANK;

  const { outstandingPermissions } = useCheckPermissions({
    requiredPermissions: [Permission.Profilewrite],
  });

  const { AppField, AppForm, ButtonForCompanyProfile, FormRow, handleSubmit } = useAppForm({
    defaultValues,
    onSubmit,
    validators: {
      onSubmit: newValidationSchema,
      onBlur: newValidationSchema,
    },
  });

  return (
    <AppForm>
      <form
        onSubmit={(e) => {
          e.preventDefault();
          e.stopPropagation();
          handleSubmit();
        }}
      >
        <FormRow>
          <AppField
            name="foregroundColor"
            children={({ Text }) => (
              <Text
                i18nKey="base.labels.foregroundColor"
                isDisabled={isBank}
                outstandingPermissions={outstandingPermissions}
              />
            )}
          />
        </FormRow>
        <FormRow>
          <AppField
            name="backgroundColor"
            children={({ Text }) => (
              <Text
                i18nKey="base.labels.backgroundColor"
                isDisabled={isBank}
                outstandingPermissions={outstandingPermissions}
              />
            )}
          />
        </FormRow>

        <ButtonForCompanyProfile
          isDisabled={isBank}
          outstandingPermissions={outstandingPermissions}
        />
      </form>
    </AppForm>
  );
}
```

I would love to be able to pass in some form level meta data that the components can just pull out of a hook . So it would look like this

```
export function VirtualCardArtView() {
  const { user } = useUserContext();
  const isBank = user?.role === EnumUserTenantType.BANK;

  const { outstandingPermissions } = useCheckPermissions({
    requiredPermissions: [Permission.Profilewrite],
  });

  const { AppField, AppForm, ButtonForCompanyProfile, FormRow, handleSubmit } = useAppForm({
    defaultValues,
    onSubmit,
    validators: {
      onSubmit: newValidationSchema,
      onBlur: newValidationSchema,
    },
    formMeta: {
      isDisabled: isBank,
      outstandingPermissions: outstandingPermissions,
    },
  });

  return (
    <AppForm>
      <form
        onSubmit={(e) => {
          e.preventDefault();
          e.stopPropagation();
          handleSubmit();
        }}
      >
        <FormRow>
          <AppField
            name="foregroundColor"
            children={({ Text }) => <Text i18nKey="base.labels.foregroundColor" />}
          />
        </FormRow>
        <FormRow>
          <AppField
            name="backgroundColor"
            children={({ Text }) => <Text i18nKey="base.labels.backgroundColor" />}
          />
        </FormRow>

        <ButtonForCompanyProfile />
      </form>
    </AppForm>
  );
}

```

or is that more of a `formMeta` thing?

harry-whorlow:
@JosieRice, thanks for the input! I got distracted for a bit closing some bug tickets. But I actually have a need, quite similarly to you, for the meta to be provided to all fields at work. 

I'll pick this up after closing out my current tickets 😄

iceikiru:
> @JosieRice, thanks for the input! I got distracted for a bit closing some bug tickets. But I actually have a need, quite similarly to you, for the meta to be provided to all fields at work.
> 
> I'll pick this up after closing out my current tickets 😄

I've been meaning to get involved with more open source contributions. Would this be something I could help with in any way.

harry-whorlow:
hi @JosieRice,
I've actually been working on this for the past week, but if things change I'll let you know. 🙂

harry-whorlow:
I've gotten sidetracked with adding the devtools to form, if someone wants to pick this up in the mean time feel free.

Issue #1177, opened 2025-02-25
New API - resetField('fieldName')


crutchcorn:
### Discussed in https://github.com/TanStack/form/discussions/1159

<div type='discussions-op-text'>

<sup>Originally posted by **wilson12365** February 19, 2025</sup>
right now the only way to handle reset a field based on other actions of the form is the following

```
listeners={{
  onChange: ({value}) => {
    const defaultMeta = {
      isValidating: false,
      isTouched: false,
      isBlurred: false,
      isDirty: false,
      errorMap: {}
    }

    if (value === ‘personal’) {
      setFieldValue(‘businessName’, ‘’)
      setFieldValue(‘businessLabel’, ‘’)
      setFieldMeta(’businessName’, defaultMeta)
      setFieldMeta(’businessLabel’, defaultMeta)
    }
}}
```

@crutchcorn has suggested to add `resetField('fieldName')`</div>

blipk:
maybe an `updateFieldMeta` function along with `setFieldMetaAndValue` function would make this a lot more ergonomic but still flexible

harry-whorlow:
@crutchcorn, I could pick this up if you could give a little more detail. I'm back from holidays so I should have some time around the weekend to open a pr

selah-EN-HC-glitch:
How can I utilize this to my advantage? Please help.

harry-whorlow:
@selah-EN-HC-glitch I just released it to main, so update your packages then you can use form.resetField(<field name>)

Laura1590:
Thank you for adding this! For nested fields, this only works on the leaves , and not on the parent field. Is there a solution that propagates the reset from the parent field to all of its nested fields?

harry-whorlow:
Closed in #1223

PR #1223, opened 2025-03-05
feat(core): form reset field utility function

harry-whorlow:
From issue #1177 and discussion #1159 this PR introduces form.resetField() utility


packages/form-core/src/FormApi.ts
```diff
@@ -1956,6 +1959,36 @@ export class FormApi<
...
+          ...prev.fieldMetaBase,
+          [field]: {
+            isValidating: false,
+            isTouched: false,
+            isBlurred: false,
+            isDirty: false,
+            isPristine: true,
+            errors: [],
+            errorMap: {},
+          },
```
crutchcorn, line 1980: We have these values elsewhere in the `getDefaultMeta` or something, don't we? Let's reuse to avoid the bundle size increase

PR #1230, opened 2025-03-06
fix code example typo in reactivity.md

Seemspyo:
I found some strange code in the guide document and fixed it. Congratulations on the official launch!


crutchcorn (approved):
Thanks! And for the kind words as well

PR #1231, opened 2025-03-06
docs(react): emphasise form composition exports

harry-whorlow:
The quick guides docs don't emphasise the importance of exporting and importing the same useFieldContext as the form context, which if not emphasised I'm pretty convinced will lead to unnecessary issues. This PR fixes that, plus a small typo.


crutchcorn (approved):
Thanks!

PR #1234, opened 2025-03-06
feat(form-core): add canSubmitWhenInvalid form option

### LINKED ISSUE #1109: Option to trigger onSubmitInvalid on every submit
**Description**
Currently, `onSubmitInvalid` is designed to be called only once, due to [this line of code](https://github.com/TanStack/form/blob/3e889183acd6d7d2b9acbb1e67a844ae33a2e19c/packages/form-core/src/FormApi.ts#L1026). This behavior limits flexibility for handling invalid form submissions, as it prevents custom logic from being executed during validation.

**Use Case**
I want to implement functionality where, if the form is resubmitted, the next field that still has an error gets focused. However, since `onSubmitInvalid` is only called once, this behavior isn't currently possible without workarounds.

**Proposal**
Remove the guard to allow `onSubmitInvalid` to be called multiple times, or provide an explicit option to opt out of the current restriction.

matthiasschwarz:
Allows to submit a form when invalid with the new `canSubmitWhenInvalid` form option.

Resolves #1109.


ColinKrist:
test

harry-whorlow:
Out of curiosity, is there a reason for the two tests? onMount and onChange?

packages/form-core/src/FormApi.ts
```diff
@@ -941,8 +945,10 @@ export class FormApi<
         const canSubmit =
           (currBaseStore.submissionAttempts === 0 &&
             !isTouched &&
-            !hasOnMountError) ||
-          (!isValidating && !currBaseStore.isSubmitting && isValid)
+            (!hasOnMountError || this.options.canSubmitWhenInvalid)) ||
+          (!isValidating &&
+            !currBaseStore.isSubmitting &&
+            (isValid || this.options.canSubmitWhenInvalid))
```
harry-whorlow, line 951: This logic dose not work, if you run `pnpm test:lib` you will get failing tests. 

would something like this work?

```ts
const submitInvalid = this.options.canSubmitWhenInvalid ?? false
const canSubmit =
  (currBaseStore.submissionAttempts === 0 &&
    !isTouched &&
    (!hasOnMountError || submitInvalid)) ||
  (!isValidating &&
    !currBaseStore.isSubmitting &&
    (isValid || submitInvalid))
```

packages/form-core/src/FormApi.ts
```diff
@@ -1089,10 +1096,11 @@ export class FormApi<
...
 
-    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-    const shouldUpdateReeval = !!options.transform?.deps?.some(
-      (val, i) => val !== this.prevTransformArray[i],
-    )
+    const shouldUpdateReeval =
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      !!options.transform?.deps?.some(
+        (val, i) => val !== this.prevTransformArray[i],
+      ) || oldOptions.canSubmitWhenInvalid !== options.canSubmitWhenInvalid
```
harry-whorlow, line 1103: @Balastrong what do you think of this? I feel that this should be handled in a way that is universal for all the options not just chaining `||` operators.

Balastrong: Yeah I'd revert this change, `shouldUpdateReeval` is there for a completely different purpose.

If we want to imperatively update the form options that's probably another story (do we even want that?)

The test will need to be updated too :)

packages/form-core/src/FormApi.ts
```diff
@@ -275,6 +275,10 @@ export interface FormOptions<
    * Optional time in milliseconds if you want to introduce a delay before firing off an async action.
    */
   asyncDebounceMs?: number
+  /**
+   * If true, allows to submit when form is invalid. Defaults to undefined.
```
Balastrong, line 279: Just one tiny detail here, I think we should be explicit and mention that this is affecting the how the `canSubmit` flag behaves.
Basically to explain that when this flag is true, canSubmit remains true even if there are some validation errors.

With that sorted out I'd say you can merge :)

Balastrong (approved):

packages/form-core/src/FormApi.ts
harry-whorlow: Awesome, thanks for the review! enjoy the weekend 😄

harry-whorlow:
LGTM 🚀  thanks @matthiasschwarz

PR #1241, opened 2025-03-07
fix(form-core): listen to array onChange events

### LINKED ISSUE #1226: Listener not run on array form field
### Describe the bug

An `onChange` listener on a `<form.Field mode="array">` is not run when there are changes to the input

https://stackblitz.com/edit/vitejs-vite-n6la3gan?file=src%2FApp.tsx

### Steps to reproduce

Example Steps:
1. Add a person
2. Notice that the household size does not increase

General Steps:

1. add an `onChange` listener to a `<form.Field mode="array">`
2. use `field.pushValue` to update the field
3. Observe that the `onChange` is not called

### Expected behavior

Would expect the `onChange` listener to be called when a value is changed.

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/d3655700-a7ed-4d8d-9b67-c7fc460d6c49

### Platform

- OS: MacOS
- Browser: Arc v1.82.1 
- Chromium Engine Version: 133.0.6943.35

### TanStack Form adapter

react-form

### TanStack Form version

v1.0.0

### TypeScript version

v5.7.2

### Additional context

_No response_

pavle99:
Listen to array field events

Closes #1226

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -1233,8 +1233,6 @@ export class FieldApi<
       value: this.state.value,
       fieldApi: this,
     })
-
-    this.validate('change')
   }
 
   /**
@@ -1251,8 +1249,6 @@ export class FieldApi<
       value: this.state.value,
       fieldApi: this,
     })
-
-    this.validate('change')
   }
 
   /**
@@ -1269,8 +1265,6 @@ export class FieldApi<
       value: this.state.value,
       fieldApi: this,
     })
-
-    this.validate('change')
   }
 
   /**
@@ -1283,8 +1277,6 @@ export class FieldApi<
       value: this.state.value,
       fieldApi: this,
     })
-
-    this.validate('change')
   }
 
   /**
@@ -1297,8 +1289,6 @@ export class FieldApi<
       value: this.state.value,
       fieldApi: this,
     })
-
-    this.validate('change')
   }
 
   /**
@@ -1311,8 +1301,6 @@ export class FieldApi<
       value: this.state.value,
       fieldApi: this,
     })
-
-    this.validate('change')
   }
 
   /**
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -1226,7 +1226,16 @@ export class FieldApi<
   pushValue = (
     value: TData extends any[] ? TData[number] : never,
     opts?: UpdateMetaOptions,
-  ) => this.form.pushFieldValue(this.name, value as any, opts)
+  ) => {
+    this.form.pushFieldValue(this.name, value as any, opts)
+
+    this.options.listeners?.onChange?.({
+      value: this.state.value,
+      fieldApi: this,
+    })
+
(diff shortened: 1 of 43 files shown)


packages/form-core/src/FieldApi.ts
```diff
@@ -1244,26 +1262,59 @@ export class FieldApi<
...
-  removeValue = (index: number, opts?: UpdateMetaOptions) =>
+  removeValue = (index: number, opts?: UpdateMetaOptions) => {
     this.form.removeFieldValue(this.name, index, opts)
 
+    this.options.listeners?.onChange?.({
+      value: this.state.value,
+      fieldApi: this,
+    })
+
+    this.validate('change')
```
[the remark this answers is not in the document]
pavle99: You are right, will remove the uneccessary validations.

packages/form-core/tests/FieldApi.spec.ts
```diff
@@ -1156,6 +1156,196 @@ describe('field api', () => {
...
+    field.moveValue(0, 2)
+
+    expect(form.getFieldValue('itemsModified')).toStrictEqual(true)
+    expect(form.getFieldValue('items')).toStrictEqual([
+      'second item',
+      'third item',
+      'first item',
+    ])
+  })
+
```
harry-whorlow, line 1348: I'm not sure these need to be all separate tests, they could all just be one test where at each step the utils are tested then be grouped under 'should listen to array field changes'. 

Saves re-instantiating form each and every time, and saves some line count, whilst also testing the same stuff. 

What do you think?

harry-whorlow: Yeah, I see your point. To be honest I don't really have a strong opinion on it. Looking back on my tests I did the same... so we can leave it as it.

pavle99: I can extract the tests into a separate function, since they repeat themselves as well and are in this form:
```ts
expect(form.getFieldValue('itemsCount')).toStrictEqual(<number>)
expect(form.getFieldValue('items')).toStrictEqual(<string_array>)
expect(form.getFieldValue('itemsModified')).toStrictEqual(<boolean>)
```

WDYT?

pavle99: For eg:
```ts
type CheckFieldValuesArgs = {
  itemsCount: number
  items: string[]
  itemsModified: boolean
}

const checkFieldValues = ({ itemsCount, items, itemsModified }: CheckFieldValuesArgs) => {
  expect(form.getFieldValue('itemsCount')).toStrictEqual(itemsCount)
  expect(form.getFieldValue('items')).toStrictEqual(items)
  expect(form.getFieldValue('itemsModified')).toStrictEqual(itemsModified)
}
```

harry-whorlow:
@pavle99 looks good! Thanks for the contribution... and fixing my mistake 🙃. Just hang tight for a bit and I'll merge it once I get a chance to double check it all 🤟

harry-whorlow:
@pavle99 Thanks for the contribution!

PR #1246, opened 2025-03-07
fix(solid-form): use { equals: false }

aadito123:
it was using a weird createMemo hack. Solid has a { equals: false } now that will track regardless of any changes so we can avoid an unnecesary memo and array creation (yay)


crutchcorn:
What minimum version of Solid.js is needed for this?

aadito123:
[0.26.0](https://github.com/solidjs/solid/blob/main/CHANGELOG.md#0260---2021-04-09)

Had no idea this was there when I made the initial solid-form implementation.

crutchcorn:
Nice! Good work!

PR #1107, opened 2025-01-10
[WIP] add svelte integration

43081j:
This is just an attempt at adding a svelte integration.

It is a draft until, if ever, we get it building and tested.

TODO:

- [ ] successfully build it
- [ ] figure out how to deal with `api.Field` being a curried `<Field>` (i.e. it needs to have `<Field ...props form={form}>`)
- [ ] add tests
- [ ] figure out how to use generics (since our generics are complex and need to reference other types, can svelte support that?)
- [ ] add standard README

PR #1247, opened 2025-03-07
feat(svelte-form): Add Svelte adapter

dummdidumm:
This implements a Svelte adapter for the form package. API-wise it's somewhere inbetween Solid and React (`createForm` is used like Solid's, getting the field state from within a snippet is more like React). Added basic tests and two examples.

Superseeds #1107


packages/svelte-form/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@tanstack/svelte-form",
+  "version": "1.0.0",
```
dummdidumm, line 3: marked this as 1.0 but if you wanna play it safe you could do `0.5.0` or sth first to see if there are any glaring mistakes made that require a breaking change

KevinVandy: I think our TanStack Config currently forces all packages in the repo to have the same version, so it will release as v1.x.0 when released

examples/svelte/simple/src/App.svelte
```diff
@@ -0,0 +1,150 @@
...
+      onclick={() => {
+        form.reset()
+      }}
+    >
+      Reset
+    </button>
+  </div>
+</form>
+
+<pre>{JSON.stringify(form.state, null, 2)}</pre>
```
lachlancollins, line 150: This doesn't seem to be reactive

packages/svelte-form/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/TanStack/form.git",
+    "directory": "packages/svelte-form"
+  },
+  "homepage": "https://tanstack.com/form",
+  "scripts": {
+    "clean": "premove ./dist ./coverage",
+    "test:eslint": "eslint ./src ./tests",
+    "test:types": "pnpm run \"/^test:types:ts[0-9]{2}$/\"",
```
lachlancollins, line 16: Can you replace these type tests with `svelte-check`?

lachlancollins (changes requested):

43081j:
Let me know if you want any help with this 👍 my pr went stale since I was in need of someone to pair/review with but I am still very much interested in landing this

examples/svelte/simple/src/App.svelte
dummdidumm: This is a leftover from some early explorations - I just removed it. FWIW for Solid and Vue this is also unreactive, from what I understand you gotta use `form.useStore(...)` to get a reactive version of (a part of) the state.

packages/svelte-form/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "./package.json": "./package.json"
+  },
+  "sideEffects": false,
+  "files": [
+    "dist",
+    "src"
+  ],
+  "dependencies": {
+    "@tanstack/form-core": "workspace:*",
+    "@tanstack/svelte-store": "^0.7.0"
```
43081j, line 44: this should come from `workspace:*` I think

packages/svelte-form/src/Field.svelte
```diff
@@ -0,0 +1,270 @@
...
+    type DeepValue,
+    FieldApi,
+    type FieldAsyncValidateOrFn,
+    type FieldValidateOrFn,
+    type FormAsyncValidateOrFn,
+    type FormValidateOrFn,
+  } from '@tanstack/form-core'
+  import { useStore } from '@tanstack/svelte-store'
+  import { onMount, type Snippet } from 'svelte'
+  import Field from './Field.svelte'
```
43081j, line 13: this is really, really odd. importing the file we're in, from the file we're in

i'm surprised this did something. even if this somehow works, we really shouldn't be doing it imo

this is why `createField` and `Field` were originally separate

packages/svelte-form/src/createForm.svelte.ts
```diff
@@ -0,0 +1,235 @@
...
+      TSubmitMeta
+    > = api as never
+
+  // @ts-expect-error constructor definition exists only on a type level
+  extendedApi.Field = (internal, props) =>
+    Field(internal, { ...props, form: api })
+  extendedApi.createField = (props) =>
+    createField(() => {
+      return { ...props(), form: api }
+    }) as never
```
43081j, line 222: how come you weren't able to type this properly? (i.e. without `as never`)

packages/svelte-form/package.json
dummdidumm: This isn't part of the monorepo, so like this is the correct way

packages/svelte-form/src/Field.svelte
dummdidumm: Why shouldn't we? This is valid JavaScript, you can self-import in ESM, it's part of the spec. This is just a different way of doing what the other adapters are doing, where they have `Field` and `createField` in the same file.

packages/svelte-form/src/createForm.svelte.ts
dummdidumm: added a comment

packages/svelte-form/tests/simple.svelte
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  export const sampleData: Employee = {
+    firstName: 'Christian',
+    lastName: '',
+  }
+</script>
+
+<script lang="ts">
+  import { createForm } from '@tanstack/svelte-form'
```
lachlancollins, line 14: There's a type error coming from here, I think it should be a relative import from `src` instead

Dreaming-Codes:
Is this ready for testing it in a project?

niemyjski:
> Is this ready for testing it in a project?

And if it is, would love to know if there is a nightly build :), Otherwise what is left / where we can help.

JonathonRP:
> > Is this ready for testing it in a project?
> 
> what is left / where we can help.

I see an error in types that has an @ts/error with no actual error, and as someone else pointed that import failing. I think the reviews need to approve and best guess they will once those two failing tests pass from the fixes I mentioned...

packages/svelte-form/src/Field.svelte
43081j: its basically a hacky workaround for the fact there's a circular dependency here (`createField` needs `Field` and vice versa), and in svelte the component is the module as a whole

ideally, we'd figure it out properly. though im not sure how off the top of my head

packages/svelte-form/src/Field.svelte
```diff
@@ -0,0 +1,270 @@
...
+      const cleanupFn = api.mount()
+      mounted = true
+      return () => {
+        cleanupFn()
+        mounted = false
+      }
+    })
+
+    // TODO (43081j): does this do what i think? we don't access anything
+    // svelte is aware of, so maybe it'll never call this?
```
43081j, line 97: presumably the `Object.values(current)` is why this `$effect.pre` works?

so can we remove the `TODO` now?

dummdidumm: We can yes - though we can also simplify this. calling the function is enough. Also added a test for this

packages/svelte-form/src/Field.svelte
niemyjski: Are you running into a problem with the circular ref? In general it's ok due to the way js imports work (e.g., you can import yourself), but ideally something we resolve.

43081j: im fully aware that this "works", its just not great for code quality/readability and is inconsistent with the rest of the codebase

also its a clear workaround for the circular dependency. you did this because you couldn't define `Field` in the file itself (since it _is_ the file), whereas in all other integrations we would just have it in a named export

i disagree with doing it this way but will leave it up to other reviewers to decide

dummdidumm: You're not gonna get a "clean" solution because there's a cyclic dependency here. So either you have a cyclic import or you have a self-import - I prefer the latter due to colocation.

examples/svelte/array/src/App.svelte
```diff
@@ -0,0 +1,55 @@
...
+    },
+    onSubmit: ({ value }) => alert(JSON.stringify(value)),
+  }))
+</script>
+
+<form
+  id="form"
+  onsubmit={(e) => {
+    e.preventDefault()
+    e.stopPropagation()
```
43081j, line 16: nit: we probably don't need to stop the propagation here. doesn't make much difference in these examples but would be good to avoid people thinking this is necessary

examples/svelte/array/src/App.svelte
```diff
@@ -0,0 +1,55 @@
...
+              <div>
+                <label>
+                  <div>Name for person {i}</div>
+                  <input
+                    value={person.name}
+                    oninput={(e: Event) => {
+                      const target = e.target as HTMLInputElement
+                      subField.handleChange(target.value)
+                    }}
+                  />
```
43081j, line 37: it may be me missing something, but html doesn't use self-closing tags anymore (they're ignored). stuff like this should be a void tag (`<input>`)

packages/svelte-form/eslint.config.js
```diff
@@ -0,0 +1,5 @@
+// @ts-check
+
+import rootConfig from '../../eslint.config.js'
+
+export default [...rootConfig]
```
43081j, line 5: do you think there's opportunity here to enable svelte rules for the svelte files in here?

not in this PR, but we could open an issue to do it separately if you think that makes sense

packages/svelte-form/tests/simple.test.ts
```diff
@@ -0,0 +1,74 @@
+/// <reference lib="dom" />
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import '@testing-library/jest-dom'
+import { userEvent } from '@testing-library/user-event'
+import { mount, unmount } from 'svelte'
+// @ts-ignore tsc doesn't know about named exports from svelte files
```
43081j, line 6: i'd just separate this out, it isn't really anything to do with the component. rather it is input

so it could go in its own `sample-data.ts` or whatever you prefer to call it. then import it here and in `simple.svelte`

43081j:
some of the CI failure seems unrelated, some `eslint.config.js` stuff from the react package

this all looks good other than the failing CI though 👍  i left a few very minor comments but nothing much

there's an unfortunate amount of `any`, unsafe casts (`as never`) and such. but there's not a huge amount we can do about them given the circular references and overly deep type inference

packages/svelte-form/eslint.config.js
dummdidumm: honestly I don't care, Eslint isn't of much use in this repo in my opinion, so I'd rather just keep it as is

packages/svelte-form/tests/simple.test.ts
dummdidumm: Now that we use svelte-check to type-check this we can just remove the ts-ignore comment
```suggestion
```

packages/svelte-form/eslint.config.js
43081j: to be clear, we were always going to keep it as it is _in this pr_. i was asking in case its worth us opening an issue to do it in a separate PR

> not in this PR, but we could open an issue to do it separately

43081j:
i tried this out locally and dug around the types a bit etc, all seems pretty good

maybe we can get some kind of nightly build sorted soon

niemyjski:
nightly build would be great. Any thing the community can help with?

lachlancollins (approved):

Issue #1134, opened 2025-01-30
React warning when setting defaultValue on conditionally rendered Field.


asegarra:
### Describe the bug

React crashes if you set a Field level defaultValue on a conditionally rendered Field.

```
 Cannot update a component (`App`) while rendering a different component (`Field`). To locate the bad setState() call inside `Field`, follow the stack trace as described in https://react.dev/link/setstate-in-render 
    at Field (https://tkj4dz.csb.app/node_modules/@tanstack/react-form/dist/esm/useField.js:43:23)
    at APIField
    at div
    at form
    at div
    at App
```

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/optimistic-golick-tkj4dz

### Steps to reproduce

1. Type something in First name so last name gets rendered

### Expected behavior

Should not crash

### How often does this bug happen?

Every time

### Platform

- OS: Windows 10

### TanStack Form adapter

None

### TanStack Form version

v0.41.3

asegarra:
I just went back and you're absolutely right it doesn't 'crash' not sure why I described it that way, my apologies.

juanvilladev:
Pretty sure this is caused by us calling setFieldValue directly on the constructor of FieldApi here:
https://github.com/TanStack/form/blob/72b4572ed644f6699eacb186fdd6c2f536c73219/packages/form-core/src/FieldApi.ts#L999

When you defined a defaultValue, and the component is rendering, within the same cycle we're attempting to update state which seems to violate a React rule about this. You'll notice in your stackblitz if you comment out defaultValue on lastName issue goes away.

Typically you'd end up having a useEffect that runs once the component is mounted, but this won't work for us since this is framework agnostic. I'll take a crack at this after #1201!

juanvilladev:
@asegarra has this been resolved for you in latest?

asegarra:
> [@asegarra](https://github.com/asegarra) has this been resolved for you in latest?

Yes it's fixed, thank you!

PR #1250, opened 2025-03-08
fix(core, react-form): Avoid set state error in React by avoiding set…

juanvilladev:
…ting default value until mount #1134 

Due to the FieldApi constructor setting defaultValue in the constructor we violate: https://react.dev/link/setstate-in-render

UPDATE:  We're going with option 2 below: moving set default value to mount function.

To avoid this I added a new flag which will prevent setting the default value within the constructor:
```ts
  /**
   * If true, the default value will not be set in the constructor.
   */
  deferDefaultValue?: boolean
```

Within the react react-form package, we then set that flag true and do the following
```ts
  const [fieldApi] = useState(() => {
    const api = new FieldApi({
      ...opts,
      form: opts.form,
      name: opts.name,
      deferDefaultValue: true, // Prevents https://react.dev/link/setstate-in-render by setting the default value after initial mount for React
    })

  //...
  useIsomorphicLayoutEffect(() => {
    if (fieldApi.options.defaultValue !== undefined) {
      fieldApi.form.setFieldValue(
        fieldApi.name,
        fieldApi.options.defaultValue as never,
        {
          dontUpdateMeta: true,
        },
      )
    }
  }, [fieldApi])
```
Please let me know if you prefer this implementation ^^

Or a different implementation where we can simply move the following code into the mount function of the FieldApi class:
https://github.com/TanStack/form/blob/135b88686bf9d10115828cd852ef2bd1201f5776/packages/form-core/src/FieldApi.ts#L999-L1003

This also solves the problem, though now we'd be affecting all other packages not just the React one. Tests pass for either implementation!

Thanks in advance!

### DIFF
--- a/docs/reference/classes/fieldapi.md
+++ b/docs/reference/classes/fieldapi.md
@@ -163,7 +163,7 @@ The current field state.
 getInfo(): FieldInfo<TParentData>
 ```
 
-Defined in: [packages/form-core/src/FieldApi.ts:1222](https://github.com/TanStack/form/blob/main/packages/form-core/src/FieldApi.ts#L1222)
+Defined in: [packages/form-core/src/FieldApi.ts:1217](https://github.com/TanStack/form/blob/main/packages/form-core/src/FieldApi.ts#L1217)
 
 Gets the field information object.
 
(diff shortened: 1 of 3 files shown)


crutchcorn (changes requested):
I think I'd rather add the `defaultValue` setting inside of `mount`.

Also, ideally we don't have to set this flag to do that. If we set it inside of `mount`, do all tests still pass?

juanvilladev:
> I think I'd rather add the `defaultValue` setting inside of `mount`.
> 
> Also, ideally we don't have to set this flag to do that. If we set it inside of `mount`, do all tests still pass on all adapters?

All tests passing on adapters, moved to simpler version of setting field value on mount function. Kept test within React that specifically tests the issue.

Issue #1256, opened 2025-03-08
Broken form validations in SolidJS?


mrlubos:
### Describe the bug

Hey! I am literally following the [guide for validations](https://tanstack.com/form/latest/docs/framework/solid/guides/validation) and I am unable to accomplish a simple validation.

### Your minimal, reproducible example

https://stackblitz.com/edit/solidjs-templates-bv3rc2x1?file=src%2FApp.tsx

### Steps to reproduce

1. Follow the instructions in the guide
2. Observe field validation doesn't fire
3. Observe submit button doesn't get conditionally disabled

### Expected behavior

Probably resolving the issue with validator not running would resolve the button state?

### How often does this bug happen?

Every time

### Platform

N/A

### TanStack Form adapter

solid-form

### TanStack Form version

1.0.5

### TypeScript version

5.1.3

theVedanta:
`onInput` is the listener you are looking for here.

So the problem here is how SolidJS treats `onChange`.

`onInput` triggers immediately when the input value changes, whether by user typing, pasting, or any other form of content modification. This event fires with each keystroke or value change.

`onChange` triggers when the input loses focus after being modified (on blur), or when the user explicitly commits a change (like pressing Enter in some inputs).

As for the docs, there are 5 instances where I see this error is made, I will verify them and make a commit based on whether their usage is OK. Thank you for bringing this up!

mrlubos:
@theVedanta found one more issue (potentially?). Notice both `canSubmit` and `isValid` are `true` by default. Even if I set `defaultState` to `false`, that doesn't change. I modified the validator to always fail, yet the form remains submittable until the first keystroke.

For DX, I wouldn't expect to have to set `defaultState` at all. The validators should be executed with default values and that's what should determine the initial state. Forcing the default state through `defaultState` might trip some people up, but would be still an acceptable workaround as long as it works.

theVedanta:
I think this is a good insight!
However, the defaultState is not something I would expect to have toggled to `false` for canSubmit and isValid. From what I get, it SHOULD be set based on the initial values. In the stackblitz example, the `onChange` checks for the validation not some validator.

If you had a validator check it, for example `onMount` on the form-level, you would get the desired output.

You can check out this example: https://stackblitz.com/edit/solidjs-templates-g56ng1la?file=src%2FApp.tsx

This has both `canSubmit` and `isValid` as `false` without the defaultState being set because it validates on-mounting not on-changing.

theVedanta:
[...]

PR #1257, opened 2025-03-09
docs(solid-form): onChange listener -> onInput

theVedanta:
With respect to #1256 

Changed in all the places I found in the guides. The examples probably remain but they need a more thorough check.


mrlubos (approved):

mrlubos:
@theVedanta updated the original issue with new context

crutchcorn (approved):
LGTM! Thanks for fixing this for us

PR #1259, opened 2025-03-09
docs(form): Effect Schema also supports Standard Schema

rehos:
Effect Schema also supports Standard Schema.

The [StackBlitz](https://tanstack.com/form/latest/docs/framework/react/examples/standard-schema) example can also be modified with:

```ts
import { Schema as S } from 'effect';

const EffectSchema = S.standardSchemaV1(
  S.Struct({
    firstName: S.String.pipe(
      S.minLength(3),
      S.annotations({
        message: () => '[Effect/Schema] You must have a length of at least 3',
      })
    ),
    lastName: S.String.pipe(
      S.minLength(3),
      S.annotations({
        message: () => '[Effect/Schema] You must have a length of at least 3',
      })
    ),
  })
);
```

See it [live](https://stackblitz.com/edit/tanstack-form-7b3dyjma?file=package.json,src%2Findex.tsx&preset=node)


Balastrong (approved):
Thank you!

PR #1261, opened 2025-03-09
feat(core): form listeners

harry-whorlow:
Adds form listeners to formApi.

### DIFF
--- a/docs/framework/react/guides/listeners.md
+++ b/docs/framework/react/guides/listeners.md
@@ -15,10 +15,10 @@ In this example, when the user changes the country, the selected province needs 
 
 Events that can be "listened" to are:
 
-- onChange
-- onBlur
-- onMount
-- onSubmit
+- `onChange`
+- `onBlur`
+- `onMount`
+- `onSubmit`
 
 ```tsx
 function App() {
@@ -90,18 +90,18 @@ We enable an easy method for debouncing your listeners by adding a `onChangeDebo
 </form.Field>
 ```
 
-### form listeners
+### Form listeners
 
 At a higher level, listeners are also available at the form level, allowing you access to the `onMount` and `onSubmit` events, and having `onChange` and `onBlur` propagated to all the form's children.
 
 `onMount` and `onSubmit` listeners have to following props:
 
-- formApi
+- `formApi`
 
-`onChange` and `onBlur` listeners have access to formApi, as well as:
+`onChange` and `onBlur` listeners have access to:
 
-- fieldName
-- fieldApi
+- `fieldApi`
+- `formApi`
 
 ```tsx
 const form = useForm({
@@ -111,11 +111,14 @@ const form = useForm({
       loggingService('mount', formApi.state.values)
     },
 
-    onChange: ({ formApi }) => {
+    onChange: ({ formApi, fieldApi }) => {
       // autosave logic
       if (formApi.state.isValid) {
         formApi.handleSubmit()
       }
+
+      // fieldApi represents the field that triggered the event.
+      console.log(fieldApi.name, fieldApi.state.value)
     },
   },
 })
--- a/docs/framework/react/guides/listeners.md
+++ b/docs/framework/react/guides/listeners.md
@@ -89,3 +89,34 @@ We enable an easy method for debouncing your listeners by adding a `onChangeDebo
   )}
 </form.Field>
 ```
+
+### form listeners
+
+At a higher level, listeners are also available at the form level, allowing you access to the `onMount` and `onSubmit` events, and having `onChange` and `onBlur` propagated to all the form's children.
+
+`onMount` and `onSubmit` listeners have to following props:
+
+- formApi
+
+`onChange` and `onBlur` listeners have access to formApi, as well as:
+
+- fieldName
+- fieldApi
+
+```tsx
+const form = useForm({
+  listeners: {
+    onMount: ({ formApi }) => {
+      // custom logging service
+      loggingService('mount', formApi.state.values)
+    },
+
+    onChange: ({ formApi }) => {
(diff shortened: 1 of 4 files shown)


harry-whorlow:
do not mere until 'infinite type' bug is fixed 🤟

packages/form-core/tests/FormApi.spec.ts
```diff
@@ -1956,7 +1956,143 @@ describe('form api', () => {
...
+        age: 0,
+      },
+      listeners: {
+        onChange: ({ formApi, fieldName }) => {
+          arr = formApi.state.values[fieldName]
+          name = fieldName
+        },
+      },
+    })
+    form.mount()
```
harry-whorlow, line 2039: I need a pointer as to why fieldName is "excessively deep and possibly infinite"

juanvilladev: Commented about this on a new discussion #1275, I was trying to figure this out as well but couldn't really find a way to break the endless types either without the refactor I mentioned :/

packages/form-core/src/FieldApi.ts
```diff
@@ -1656,12 +1656,22 @@ export class FieldApi<
           value: this.state.value,
           fieldApi: this,
         })
+
+        this.form.options.listeners?.onBlur?.({
+          fieldName: this.name,
+          formApi: this.form,
+        })
```
harry-whorlow, line 1663: Here I used to have fieldValue also being returned, however since it could be any of the field values I couldn’t type it. 

I though it best to remove it and let the user access it with something like formApi.values[fieldName] so either its any, or if its a defined value such as formApi.values.age it can be inferred.

Balastrong: Challenge: why don't we pass the entire `fieldApi` to the callback?

harry-whorlow: sure thing, I would also like to provide the formApi without having to de-structure form field... since one of the uses of the form-listener, is for logging and having the formApi directly to me would be helpful. Also, keeps it more consistent with the onMount and onSubmission.

So something like:
this.form.options.listeners?.onBlur?.({
  fieldName: this.name,
  formApi: this.form,
  filedApi: this
})

What do you think?

harry-whorlow:
Note to self: move onChange out of the form listeners debounce logic and implement it in form

packages/form-core/src/FieldApi.ts
```diff
@@ -1704,6 +1704,11 @@ export class FieldApi<
 
   private triggerOnBlurListener() {
     const debounceMs = this.options.listeners?.onBlurDebounceMs
+    this.form.options.listeners?.onBlur?.({
+      fieldName: this.name,
```
crutchcorn, line 1708: Nit: Do we need `fieldName`? Couldn't we just do `fieldApi.options.name`?

docs/framework/react/guides/listeners.md
```diff
@@ -89,3 +89,34 @@ We enable an easy method for debouncing your listeners by adding a `onChangeDebo
   )}
 </form.Field>
 ```
+
+### form listeners
```
crutchcorn, line 93:
```suggestion
### Form listeners
```

docs/framework/react/guides/listeners.md
```diff
@@ -89,3 +89,34 @@ We enable an easy method for debouncing your listeners by adding a `onChangeDebo
...
+
+At a higher level, listeners are also available at the form level, allowing you access to the `onMount` and `onSubmit` events, and having `onChange` and `onBlur` propagated to all the form's children.
+
+`onMount` and `onSubmit` listeners have to following props:
+
+- formApi
```
crutchcorn, line 99:
````suggestion
- `formApi`
````

File: docs/framework/react/guides/listeners.md [hunk 6]
```diff
@@ -89,3 +89,34 @@ We enable an easy method for debouncing your listeners by adding a `onChangeDebo
...
+
+`onChange` and `onBlur` listeners have access to formApi, as well as:
+
+- fieldName
+- fieldApi
```
crutchcorn, line 104:
````suggestion
- `fieldApi`
````

crutchcorn (changes requested):
Two minor

packages/form-core/src/FieldApi.ts
harry-whorlow: Sure, for me it was more of a convenience thing, but as I said to @Balastrong, it's not a hill I want to die on. I'll change it.

MartinCura:
Hey, awesome addition! If this also had debounceMs i could replace a listener i've added to all fields in a form 🚀 (though of course it's a nit!)

harry-whorlow:
@MartinCura hmm that makes sense to me... would also keep it more consistent across the form and field levels, give me a day or two. 😄

Balastrong:
@MartinCura might be worth exploring [TanStack Pacer](https://tanstack.com/pacer/latest/docs/overview) in the meantime 👀

MartinCura:
> @MartinCura might be worth exploring [TanStack Pacer](https://tanstack.com/pacer/latest/docs/overview) in the meantime 👀 

Uuuh didn't know about that one! Though it makes sense as it looks quite new.

Don't worry, used field listeners (replacing a `useDebounce` from another library). Grazie! Keep up the good work (and yt channel)!

MartinCura:
> @MartinCura I've implemented the debounced listeners just now... I'm waiting for a review and I'll deploy it 😄 
> 
> Enjoy the weekend!

You're a machine. Danke schön !

harry-whorlow:
Kein problem, It was just deployed... enjoy the new feature!

PR #1262, opened 2025-03-09
perf(form-core): reduce instantiations when type checking

chorobin:
I have used the large example in the repo (this isn't very large so difference is not so large) to check reduction in instantiations and type checking time. I am using `extendedDiagnostics` but also traced the example

Before:

```
> tsc "--extendedDiagnostics"

Files:                         215
Lines of Library:            43159
Lines of Definitions:        80538
Lines of TypeScript:           209
Lines of JavaScript:             0
Lines of JSON:                   0
Lines of Other:                  0
Identifiers:                105350
Symbols:                    118110
Types:                       18129
Instantiations:             205206
Memory used:               157045K
Assignability cache size:     9324
Identity cache size:           223
Subtype cache size:              5
Strict subtype cache size:       0
I/O Read time:               0.05s
Parse time:                  0.43s
ResolveModule time:          0.09s
ResolveTypeReference time:   0.00s
ResolveLibrary time:         0.02s
Program time:                0.66s
Bind time:                   0.21s
Check time:                  0.71s
printTime time:              0.00s
Emit time:                   0.00s
Total time:                  1.58s
```

After

```
> tsc "--extendedDiagnostics"

Files:                         215
Lines of Library:            43159
Lines of Definitions:        80528
Lines of TypeScript:           209
Lines of JavaScript:             0
Lines of JSON:                   0
Lines of Other:                  0
Identifiers:                105504
Symbols:                     85611
Types:                        6188
Instantiations:              27643
Memory used:               133987K
Assignability cache size:     1640
Identity cache size:           166
Subtype cache size:              5
Strict subtype cache size:      12
I/O Read time:               0.05s
Parse time:                  0.44s
ResolveModule time:          0.10s
ResolveTypeReference time:   0.00s
ResolveLibrary time:         0.02s
Program time:                0.68s
Bind time:                   0.21s
Check time:                  0.37s
printTime time:              0.00s
Emit time:                   0.00s
Total time:                  1.26s
```

There are three main things contributing to lower instantiations:
1) Variance annotations have been added, this means when comparing two types with different type parameters, the variance worker takes less time as we are hinting to the compiler how our type parameters are going to be used
2) `TName` is not constrained and uses a `ValidateName` utility instead. I prefer not to use contraints at inference sites against complex types that could be unions. This is because `TName` could default to the union and this can slow things down in a few situations.
3) I re-worked `DeepKeys` and `DeepValue`. Instead of creating a union and then parsing `TName` again, a map is constructed so `TValue` can be grabbed by `TName`. `TName` -> `TValue`


packages/form-core/tests/FieldApi.test-d.ts
```diff
@@ -32,7 +32,7 @@ it('should type value when nothing is passed into constructor', () => {
   })
 
   assertType<string | undefined>(field.state.value)
-  assertType<'name'>(field.options.name)
```
Balastrong, line 35: I think `field.option.name` should be `'name'` only and that's also true on the code, maybe this change was lost there during a refactor? (same for the two instances below on this file).

I see the test is still passing with `assertType<'name'>`

packages/form-core/tests/util-types.test-d.ts
```diff
@@ -24,8 +24,13 @@ assertType<
 type ArraySupport = DeepKeys<{ users: User[] }>
 assertType<
   | 'users'
+  | `users[number]`
```
Balastrong, line 27: Question: the types with `[number]` are added to `ArrayAccessor` on purpose for the autocomplete, right?

I see it goes from this (on main)
image

to this
image

Balastrong (commented):
Thanks for the improvements and sorry for taking so long for the review!

I just added a couple comments, overall looks great! (and _slightly_ beyond my TS knowledge, but I trust you :D)

packages/form-core/tests/util-types.test-d.ts
chorobin: Yes, what do you think about this? Would you prefer it is not here?

chorobin: Hm, yeah. Thinking about it. I don't think this is a good idea to have for now. I think ideally it should show as a suggestion but then type error if you actually use it. But this then requires the user to edit the type after they have selected it which I'm not sure if I like

harry-whorlow: Is this something that we've seen requested? So far to me it seems like potentially adding complexity in a situation where the benefit isn't that apparent... But either way I don't have a strong opinion. 😊

Balastrong (approved):

PR #1268, opened 2025-03-11
feat(types): make types on functions extend off the base classes generics

s-hoff:
The goal here is to eventually allow a `withForm` component to accept "extending" forms, or those which are bigger then what they know about but do contain what they need.
I suspect that even an alternative solution to `withForm` would still need those changes to work.

While looking at the TS error that comes when trying this, I noticed that it was always about the types in the functions but never about the object attributes. So after some experimenting, it turns out that this fails:
```ts
type Test<T> = {
  defaultValues: T
  fun: (props: { value: T }) => {}
}
const t: Test<{ firstName: string }> =
  {} as Test<{ firstName: string, lastName: string }>
```

While this works:
```ts
type Test2<T> = {
  defaultValues: T,
  fun: <TF extends T = T>(props: { value: TF }) => void
}

// ist fine, the actual type has all that was asked for and then some
const t2: Test2<{ firstName: string }> =
  {} as Test2<{ firstName: string, lastName: string }>
```

and still ensures that `firstName` is included on t2. The main idea is unless `props.value` only extends T, TypeScript checks the equality both ways. If it does extend T, the checking becomes unidirectional.

This obviously still needs a lot of work to be done as most likely all function signatures would have to be changed in such a matter and more type errors will come up along the way.

But at least this works to prove the concept:
```ts
const f1 = {} as FormOptions<
  { firstName: string, lastName: string },
  undefined,
  undefined,
  undefined,
  undefined,
  undefined,
  undefined,
  undefined,
  undefined
>

const f2: FormOptions<
  { firstName: string },
  undefined,
  undefined,
  undefined,
  undefined,
  undefined,
  undefined,
  undefined,
  undefined
> = f1
```
with surprisingly few errors in `FieldApi.ts` to track down.

Once its all the way done, the inference should still work just fine.

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -193,40 +193,43 @@ export interface FormValidators<
  */
 export interface FormTransform<
   TFormData,
-  TOnMount extends undefined | FormValidateOrFn<TFormData>,
-  TOnChange extends undefined | FormValidateOrFn<TFormData>,
-  TOnChangeAsync extends undefined | FormAsyncValidateOrFn<TFormData>,
-  TOnBlur extends undefined | FormValidateOrFn<TFormData>,
-  TOnBlurAsync extends undefined | FormAsyncValidateOrFn<TFormData>,
-  TOnSubmit extends undefined | FormValidateOrFn<TFormData>,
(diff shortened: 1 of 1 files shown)


harry-whorlow:
@s-hoff draft is more a type of pr than the title 🤟, just updated the pr for you

PR #1271, opened 2025-03-12
docs(core): add comparison to angular forms

forivall:
Simple addition to the comparison table for angular's first-party forms. Needs some additional work, and verification of "SSR integration". (I presumed this was the case since it's first party and angular has first-party SSR). The "fully inferred typescript" is marked as "extra user-code" since that was added recently, and it sometimes requires some additional work to get the types correct, from my experience so far.


crutchcorn:
Not inherently related to this PR, but I wonder if we should move to a top-down comparison chart or some other visualization given how many form libraries there are 😅

crutchcorn:
I'm going to close this, but only because we need a better story for tabbed comparison tables between many framework's alternatives.

Something like:

```
<ANGULAR> <REACT> <VUE>
```

Where you pick the framework and see the competition that way.

@LadyBluenotes I know you have some stuff written there, can you make a new PR for our Form docs and add tabs/an initial comparison chart for the supported frameworks?

PR #1280, opened 2025-03-15
docs(vue): use event modifiers to enhance readability

Psycarlo:
Instead of using the arrow function:

```vue
(e) => {
  e.preventDefault()
  e.stopPropagation()
  form.handleSubmit()
}
```

we can just use [event modifiers](https://vuejs.org/guide/essentials/event-handling#event-modifiers) like this:

```vue
@submit.prevent.stop="form.handleSubmit"
```

enhancing readability ✨

### DIFF
--- a/docs/framework/vue/quick-start.md
+++ b/docs/framework/vue/quick-start.md
@@ -23,15 +23,7 @@ const form = useForm({
 
 <template>
   <div>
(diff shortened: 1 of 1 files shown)


crutchcorn (approved):
Thanks for this!

Issue #1130, opened 2025-01-27
`Field`s re-render twice on each keystroke, based on validation state


aditya-lenze:
### Describe the bug

While trying out the [Simple](https://tanstack.com/form/latest/docs/framework/react/examples/simple) example on the website in the docs, as well as locally, I stumbled upon the following behavior, which I am finding difficult to reason about:
- If there is a validation defined on the `Field`, it renders once on each `onChange`, as long as the data is **invalid**.
    - Once the data becomes valid, it starts rendering twice on each keystroke.
- If there is no validation defined, the `Field` renders twice on each keystroke.

**Please note that:**
- `React.strictMode`, which runs all effects twice, is not used.
- Renders are measured using a simple `RenderCount` component and `console.log`, both of which are in the CSB link below.
- This is issue is not linked to the number of inputs. For example, even with 3, 4, 5 inputs, the re-render occurs twice.

**CSB details**:
- Contains two fields, `firstName` and `lastName`.
- `firstName` has no validation defined.
- `lastName` has a validation of **minimum length: 5.**

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/great-yonath-3z6vqt

### Steps to reproduce

1. Go to the CSB link.
2. Try typing a `firstName`, observe the render count and `console.log`s. They should increment by two each time.
3. Try typing a `lastName`. As long as it is invalid and error message is displayed, it will render once on each keystroke. Once length > 5, and input is valid, it will start re-rendering twice.

### Expected behavior

- I would expect that the inputs are re-rendered only once on each keystroke, instead of two.
- This problem becomes peculiar when paired with validation, i.e., the issue happens only if the field is valid or no validation defined.

### How often does this bug happen?

Every time

### Platform

- OS: Windows
- Browser: Chrome
- Version: 131.0.6778.265

### TanStack Form adapter

react-form

### TanStack Form version

0.41.3

PR #1291, opened 2025-03-16
perf(react-form): skip updating isValidating by default

MVaik:
fixes https://github.com/TanStack/form/issues/1130

BREAKING CHANGE: isValidating is now opt-in

### DIFF
--- a/packages/react-form/tests/useField.test.tsx
+++ b/packages/react-form/tests/useField.test.tsx
@@ -290,26 +290,27 @@ describe('useField', () => {
                   </form.Field>
                 )
               }
-              return null
+              return (
+                <form.Field name="secondField">
+                  {({ handleChange, state }) => (
+                    <label>
+                      second field
+                      <input
+                        data-testid="second-field"
+                        value={state.value}
+                        onChange={(e) => handleChange(e.target.value)}
+                      />
+                    </label>
+                  )}
+                </form.Field>
+              )
             }}
           </form.Subscribe>
-          <form.Field name="secondField">
-            {({ handleChange, state }) => (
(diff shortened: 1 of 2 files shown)


MVaik:
Unsure if this is even the right direction, as it's essentially flipping an established default purely for react performance.

packages/react-form/tests/useField.test.tsx
```diff
@@ -290,28 +290,26 @@ describe('useField', () => {
                   </form.Field>
                 )
               }
-
```
MVaik, line 293: Moved secondField out of subscribe as React reused the field, which after removing double rendering caused firstField's value to end up in secondField.
Kind of a weird bug.

MVaik:
Switched to draft because "should properly update conditionally rendered fields" was acting weird after the change and I made an unprompted fix, but it was likely intentionally set up that way.

packages/react-form/tests/useField.test.tsx
MVaik, line 1: "should properly update conditionally rendered fields" is now failing after this change, the "hello" that is written into firstField somehow ends up in secondField after the checkbox is toggled and firstField is no longer shown.

crutchcorn:
We absolutely cannot (and should not) force an API shift simply for minor perf gains.

However, we can check if this field is read by using a proxy if that's a route we want to investigate.

Closing this PR on that principal.

> Side note; I'm like 99% sure this can be fixed simply by adding a `batch` method call from TanStack Store somewhere... I just dunno where at a quick glance. Your test will help figure this out - keep digging! 😄🤘

PR #1298, opened 2025-03-17
docs: Change some words on react guides documents

nafasebra:


crutchcorn:
These docs changes don't feel substantial enough to change the experience with our docs - closing.

PR #1318, opened 2025-03-21
feat(react-form): return parsed data from serverValidate function

SebbeBroman:
Change serverValidate return type from `Promise<void>` to `Promise<TData>` to provide direct access to validated and parsed data. Eliminates the need to decode formData twice, enhancing developer experience with immediate typed access to validation results.

Closes #1317

### DIFF
--- a/docs/framework/react/guides/ssr.md
+++ b/docs/framework/react/guides/ssr.md
@@ -69,6 +69,11 @@ export const handleForm = createServerFn({
     try {
       const validatedData = await serverValidate(ctx.data)
       console.log('validatedData', validatedData)
+      // Persist the form data to the database
+      // await sql`
+      //   INSERT INTO users (name, email, password)
+      //   VALUES (${validatedData.name}, ${validatedData.email}, ${validatedData.password})
+      // `
     } catch (e) {
       if (e instanceof ServerValidateError) {
         // Log form errors or do any other logic here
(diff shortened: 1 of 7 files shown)


harry-whorlow:
@SebbeBroman nice work on the PR, I'm just waiting on some conformation from the other maintainers and this should get merged. 🤟

SebbeBroman:
I don't really know how to show you the intended usage. Maybe just a comment? At the console log you have decoded and validated the data meaning it is ready to be inserted into a db or something.
Current implementation is like this:

```
// I "have to" explicitly add this internal dependency
import { decode } from "decode-formdata";

handler(async (ctx) => {
    try {
      const thisIsNull = await serverValidate(ctx.data)
    } catch (e) {
      if (e instanceof ServerValidateError) {
        // Log form errors or do any other logic here
        return e.response
      }

      // Some other error occurred when parsing the form
      console.error(e)
      setResponseStatus(500)
      return 'There was an internal error'
    }
    // We need to manually decode the form data and set it's type
    const decoded = decode<{name: string}>(data); 
    
    // await sql`insert into mytable name values ${decoded.name}`

    return 'Form has validated successfully'
  })
```

The new implementation would be
```
handler(async (ctx) => {
    try {
      // This has the type from serverValidate and the formData is already parsed
      const decodedData = await serverValidate(ctx.data)
       // await sql`insert into mytable name values ${decoded.name}`
    } catch (e) {
      if (e instanceof ServerValidateError) {
        // Log form errors or do any other logic here
        return e.response
      }

      // Some other error occurred when parsing the form or doing other logic
      console.error(e)
      setResponseStatus(500)
      return 'There was an internal error'
    }

    return 'Form has validated successfully'
  })
```

Balastrong:
Thanks for the example! It would be great to have a really minimal usecase to add in the docs. Like, what can you do now that you have that returned data? If you can highlight that in a couple of lines instead of the console.log that might be enough

SebbeBroman:
I added an example as a comment, cant really think of a way to showcase it without adding a dependency (database adapter, filesystem access etc)

Balastrong (approved):
That's ok, thank you!

Issue #1171, opened 2025-02-22
Removal of objects from nested arrays doesn't work as expected


Oudwins:
### Describe the bug

When removing an object from a nested array only the "id" field is removed from the element. This happens both if you use arrayField.removeValue and form.setFieldValue.

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-form-dbycl9vt?file=src%2Findex.tsx

### Steps to reproduce

1. Head to the stack bliz
2. Click on any of the rendered codes (except the last one, that one works correctly)
3. You will see in the console the logs of the data before and after each step. You will also see a react key error and can verify that although the object was removed the log that shows the data after the update shows the object that was removed without a id field

### Expected behavior

The object is correctly removed. No key error is triggered. React renders the new list correctly.

### How often does this bug happen?

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/004bb73a-6824-4f55-8bb0-0f8b1fc3fcce

### Platform

- OS: Linux
- Browser: chrome, brave

### TanStack Form adapter

None

### TanStack Form version

^0.43.2

### Additional context

Example is a little chaotic. Monday I can clean it up. But I think the video shows what is happening quite well.

crutchcorn:
Yeah, apologies @Oudwins I'm not clear what's happening in this example or what the bug you're suggesting exists is 😅

Oudwins:
> Yeah, apologies @Oudwins I'm not clear what's happening in this example or what the bug you're suggesting exists is 😅

Oh, thats a shame. I can try to make an even more minimal example and explain it further.

But for that it would be very helpful if you could explain a little bit further what you have understood (if any) and what you are confused about specifically

MinnDevelopment:
From what I can tell, the `fieldMeta` still has a key for one of the fields inside the removed element:

![Image](https://github.com/user-attachments/assets/a70c9291-9958-4d90-b88b-22749efac7c3)

On a subsequent render cycle, it's then re-added to the values:

![Image](https://github.com/user-attachments/assets/2dfc2457-afb3-44b2-b680-c5bba3116ba3)

Changing the key on the div from `item.id` to `itemIndex` fixes it (see screenshot below). I assume there is a problem in the state handling on `form.Field`.

![Image](https://github.com/user-attachments/assets/479a59de-06b2-4491-9b20-9f3a2f037b99)

crutchcorn:
Oh, in that case; does v0.48.1 solve this issue for removeValue?

[...]

PR #1319, opened 2025-03-21
fix(react-form): removing array items (#1171)

niba:
It is impossible to remove an item from an array while properly using React keys.

Here is an [example](https://stackblitz.com/edit/tanstack-form-gcsxeyng?file=src%2Findex.tsx,src%2Fminimal.tsx) taken from the issue that demonstrates the problem. Try to remove the middle item.

This issue occurs because we use the ⁠index inside the field name. When you remove any item other than the last one, the array is re-iterated, and in the next render, the field names change.
In the current code, you update the name as the last action and use the old name in earlier actions. 

Let's assume you have an array:` ⁠[{ id: 1, name: "one"}, {id: 2, name: "two"}, {id: 3, name: "three"}]`, and you remove the middle item. In the next render loop, you want to render `⁠[{ id: 1, name: "one"}, {id: 3, name: "three"}]`. However, for the item with ⁠id = 3, the store tries to access the value using the name `⁠"items[2].value"` instead of `⁠"items[1].value"`, which causes unexpected effects (see the example).

This bug doesn't occur when you use the index as a key (which is not correct). In this scenario, React behaves a little differently and doesn't cause a re-render.

I'm not familiar with the entire codebase, so I don't know if this change could have any side effects.

### DIFF
--- a/docs/reference/classes/fieldapi.md
+++ b/docs/reference/classes/fieldapi.md
@@ -163,7 +163,7 @@ The current field state.
 getInfo(): FieldInfo<TParentData>
 ```
 
-Defined in: [packages/form-core/src/FieldApi.ts:1217](https://github.com/TanStack/form/blob/main/packages/form-core/src/FieldApi.ts#L1217)
+Defined in: [packages/form-core/src/FieldApi.ts:1222](https://github.com/TanStack/form/blob/main/packages/form-core/src/FieldApi.ts#L1222)
 
 Gets the field information object.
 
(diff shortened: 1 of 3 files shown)


crutchcorn:
Can we add tests to verify that we don't regress problems?

niba:
@crutchcorn  I added a test and found another problem that I resolved.

After my first change, the array fields started rendering correctly, but the displayed values were wrong.

The reason was that `defaultValue` was not updated after removing any item in the array. In the next render, `FieldApi` calls `setValue` with a `defaultValue` that is not valid because it belongs to the removed field.

The flow:
```
defaultValues: ["a", "b", "c"]

render: 
"a",
"b",
"c",

remove "b"

render:
"a"
"b" WRONG_VALUE ->  this.state.value is undefined, so it takes value from ⁠defaultValues where ⁠defaultValue[1] === "b". This one wins because the updater is more important than store calculation
```

To fix this I added new code that modifies how ⁠setValue is called when a name change is detected.

Balastrong:
Can you please also add some tests to validate the new default logic?

niba:
The line not covered by tests is an old logic line. I needed to add a new 'if' branch, so I refactored the code slightly to make it easier to understand.

The default value logic remains the same and should be covered by the tests introduced here: [GitHub Pull Request #445](https://github.com/TanStack/form/pull/445).

If you want, I can add some tests, but they would duplicate existing ones (setting the default value on mount).

Thanks!

Balastrong (approved):
LGTM!

lpkobamn:
Thanks for your work, but it didn't solve the problem, the same situation with version 1.3.0

niba:
@lpkobamn, here is the minimal reproduction: https://stackblitz.com/edit/tanstack-form-do5av6rd?file=README.md

As you can see, it works on the latest version. Maybe your case is slightly different. Please create a bug with a link to the reproduction of your problem.

PR #1324, opened 2025-03-23
fix(form-core): fix broken sync/async validation state not shifting with array actions (remove,swap,shift,move...)

### LINKED ISSUE #1323: "Cannot read properties of undefined" when removing array field
### Describe the bug

When removing elements from array using the removeFieldValue api, we sometimes receive:
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'errorMap')
```

I am able to reproduce this when removing an element that contains a field inside that has an error. 
At some point in the validation logic triggered within removeFieldValue, we're reaching out to get the errorMap for a field that's undefined at that point. 

https://stackblitz.com/edit/vitejs-vite-ypgebmzr?file=src%2Fcomponents%2FForm.tsx

### Steps to reproduce

Simply add a few rows, then delete at least 2. Take a look at the console and you will see:
```
@tanstack_react-form.js?v=b58a52f4:970 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'errorMap')
    at @tanstack_react-form.js?v=b58a52f4:970:27
    at functionalUpdate (@tanstack_react-form.js?v=b58a52f4:410:42)
    at @tanstack_react-form.js?v=b58a52f4:1132:22
    at Store.setState (@tanstack_react-form.js?v=b58a52f4:283:124)
    at FormApi.setFieldMeta (@tanstack_react-form.js?v=b58a52f4:1127:22)
    at @tanstack_react-form.js?v=b58a52f4:967:20
    at batch (@tanstack_react-form.js?v=b58a52f4:255:5)
    at FormApi.validateSync (@tanstack_react-form.js?v=b58a52f4:924:7)
    at FieldApi.validate (@tanstack_react-form.js?v=b58a52f4:1894:123)
    at FormApi.validateField (@tanstack_react-form.js?v=b58a52f4:918:28)
```

Notice how when we push rows we are pushing a value that is not valid (per the form level schema) to the interestLevel field. If we were to push a valid value (greater than 5) the issue goes away.

### Expected behavior

When I remove an array element, regardless of subfields containing errors or not, we should never be attempting to pull errorMap from undefined (resulting in the error message).

Every time

### Screenshots or Videos

_No response_

### Platform

React, TypeScript, 

### TanStack Form adapter

None

### TanStack Form version

v1.1.2

### TypeScript version

v5.7.2

### Additional context

This will affect any users that attempt to use a standard schema validator at form level and attempt to delete a row with subfields containing errors. I'm looking into fixing it already, please assign me the issue if you have access! Thanks!

juanvilladev:
Previously we were using a cumulative field errors map maintained at the form api level. 
This PR removes that data structure in favor of a errorSourceMap data structure that lives within the FieldMeta for each field. This allows shifting/removing/deleting fields to automatically keep the errorSourceMap in sync. 

This will fix #1323 and prevent other bugs from being raised around validation not working after other array actions like removing, shifting, swapping, moving, etc...

### DIFF
--- a/examples/angular/array/package.json
+++ b/examples/angular/array/package.json
@@ -18,7 +18,7 @@
     "@angular/platform-browser": "^19.2.4",
     "@angular/platform-browser-dynamic": "^19.2.4",
     "@angular/router": "^19.2.4",
-    "@tanstack/angular-form": "^1.2.2",
+    "@tanstack/angular-form": "^1.2.3",
     "rxjs": "^7.8.2",
     "tslib": "^2.8.1",
     "zone.js": "^0.15.0"
--- a/examples/angular/simple/package.json
+++ b/examples/angular/simple/package.json
@@ -18,7 +18,7 @@
     "@angular/platform-browser": "^19.2.4",
     "@angular/platform-browser-dynamic": "^19.2.4",
     "@angular/router": "^19.2.4",
-    "@tanstack/angular-form": "^1.2.2",
+    "@tanstack/angular-form": "^1.2.3",
     "rxjs": "^7.8.2",
     "tslib": "^2.8.1",
     "zone.js": "^0.15.0"
--- a/examples/lit/simple/package.json
+++ b/examples/lit/simple/package.json
@@ -9,7 +9,7 @@
     "test:types": "tsc"
   },
   "dependencies": {
-    "@tanstack/lit-form": "^1.2.2",
+    "@tanstack/lit-form": "^1.2.3",
     "lit": "^3.2.1"
   },
   "devDependencies": {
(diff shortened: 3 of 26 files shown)


juanvilladev:
Please hold off on reviewing this, I'm testing a way to remove the cumulative error map entirely since it's introducing quite a bit of complexity.

juanvilladev:
My goal with this PR is to fix the main issue within #1323, this made me realize the approach with the cumulative map is quite complex when it doesn't need to be. I have an idea to make this a lot simpler. Will move to ready for review once implemented.

juanvilladev:
Async validation & sync validation now is fully synced and thoroughly tested with the new errorSourceMap. This reduces the complexity of using the cumulative errors map significantly and fixes all the issues with validations running on the wrong fields when removing/inserting/moving/shifting array fields. 

Please review when you have time!

Thanks!

juanvilladev:
Whoops messed up the rebase

juanvilladev:
Updated! Please review when you have time!

PR #1333, opened 2025-03-24
chore(deps): remove form-core package as it is unused and re-exported from react-form

yjoer:
This PR eliminates the @tanstack/form-core package from the example, as it’s currently unused. If its functionality is needed, it's already [re-exported](https://github.com/TanStack/form/blob/v1.1.2/packages/react-form/src/index.ts#L1) from @tanstack/react-form. By dropping it, we avoid the hassle of keeping an extra package in sync with the adapter.


crutchcorn (approved):
LGTM!

PR #1334, opened 2025-03-25
Allow withForm components to accept extending properties

s-hoff:
Replaces https://github.com/TanStack/form/pull/1268.

This fails:
```ts
type Test<T> = {
  defaultValues: T
  fun: (props: { value: T }) => {}
}
const t: Test<{ firstName: string }> =
  {} as Test<{ firstName: string, lastName: string }>
```
But this works:
```ts
type Test2<T> = {
  defaultValues: T,
  fun: <TF extends T = T>(props: { value: TF }) => void
}

const t2: Test2<{ firstName: string }> =
  {} as Test2<{ firstName: string, lastName: string }>
```

Its probably missing some functions and certainly the other frameworks.

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -193,43 +193,40 @@ export interface FormValidators<
  */
 export interface FormTransform<
   TFormData,
+  TOnMount extends undefined | FormValidateOrFn<TFormData>,
+  TOnChange extends undefined | FormValidateOrFn<TFormData>,
+  TOnChangeAsync extends undefined | FormAsyncValidateOrFn<TFormData>,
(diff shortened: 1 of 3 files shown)


LeCarbonator:
I've mentioned it on Discord previously, but I should list it here in case a maintainer has comments on it:

Some methods can create behaviour that may be unintended if `withForm` is extendable. One such example is `form.reset()`.

Even when providing all values in the `withForm`, since it is a subsection of a form, it will reset fields outside of its scope.

LeCarbonator:
This feature will be implemented in #1469 . `withForm` will remain intended for splitting single forms into multiple bits.

PR #1337, opened 2025-03-26
docs(form): document relation of isSubmitted to onSubmit

Pascalmh:
Adding a more explicit explanation for `form.onSubmit`

Triggered by the following question:

> Does `isSubmitted = true` imply that the `onSubmit` was successful?


gioboa (approved):
Thanks, nice :clap:

Balastrong (commented):
Thanks @Pascalmh for the PR!

Since we're making it more explicit (I like it!) do you think we should also mention that the flag goes back to false at each new submission attempt?

Or simply put, goes true if the last execution of onSubmit completed succesfully

Pascalmh:
@Balastrong I've added a note about `isSubmitting` as well as it indicates the time between the _attempt_ and the _success_.

Balastrong (approved):
Thanks!

@harry-whorlow feel free to merge when you're happy with it :)

PR #1353, opened 2025-03-28
docs: add form button note

### LINKED ISSUE #1344: Issue with form.reset() and <select>
### Describe the bug

When resetting a form with a `<select>` field, the field value resets to the first option and not the `defaultValue` specified.

https://codesandbox.io/p/sandbox/nifty-hill-83zxdt

### Steps to reproduce

1. Go on the sandbox
2. Click on reset

### Expected behavior

The value changes to A no matter what, I'd expect it to reset to D.

Every time

### Screenshots or Videos

_No response_

### Platform

Safari 18.3

### TanStack Form adapter

react-form

### TanStack Form version

v1.1.2

### TypeScript version

v5.7.3

### Additional context

_No response_

gioboa:
Fixes #1344 

With form `<button type="reset" ... />` the native form.reset will be triggered.
To prevent this behavior we can use event `preventDefault`.

image


docs/framework/react/guides/basic-concepts.md
```diff
@@ -112,6 +112,11 @@ Example:
 />
 ```
 
+## Form button with type="reset"
+
+> When using `<button type="reset">` in conjunction with TanStack Form's `form.reset()`, you need to prevent the default HTML reset behavior to avoid unexpected resets of form elements (especially `<select>` elements) to their initial HTML values.
+Use `event.preventDefault()` inside the button's `onClick` handler to prevent the native form reset. Alternatively, using `<button type="button">` will prevent the native HTML reset.
```
Pascalmh, line 118: Maybe you could state the full code to use as an alternative in the last sentence instead. 

Might even be nice to have in the first sentence as well, full example code of how to do it instead of "only" an explanation.

Pascalmh:
The commit should start with `docs:` (instead of `feat(docs)`, that makes a difference on what release type it will trigger.

gioboa:
I see, let me fix it

docs/framework/react/guides/basic-concepts.md
gioboa: Fixed [760e8f4](https://github.com/TanStack/form/pull/1353/commits/760e8f47b680b986cc0f4c1a5ce1498da71b08ab) ✅

harry-whorlow:
I'm in agreement with @Pascalmh, the buttons are wrapped in Subscribe, but it's not used for anything personally I would slim it down. I'm also not certain we need the information duplicated, what do you guys think? 

Other than that thanks for the PR! 🤟

docs/framework/react/guides/basic-concepts.md
```diff
@@ -112,6 +112,94 @@ Example:
...
+                  onChange={(e) => field.handleChange(e.target.value)}
+                >
+                  <option value={"A"}>A</option>
+                  <option value={"B"}>B</option>
+                  <option value={"C"}>C</option>
+                </select>
+              </>
+            );
+          }}
+        />
```
harry-whorlow, line 156: Is this needed? its not used and it only adds to the line count of the page 😊

gioboa: Fixed.

docs/framework/react/quick-start.md
```diff
@@ -134,3 +134,35 @@ ReactDOM.createRoot(rootElement).render(<PeoplePage />)
...
+```tsx
+<button
+  type="button"
+  onClick={() => {
+    form.reset()
+  }}
+>
+  Reset
+</button>
+```
```
harry-whorlow, line 168: After reading through the quick start guide, feel this is a little out of place. Perhaps, keeping it to only basic-concepts is best, what do you think? 🤔

gioboa: Yep, it's ok to me. Let me fix it

harry-whorlow:
LGTM 🚀  thanks for the PR!!

PR #1357, opened 2025-03-29
docs(react-form): Update docs to reflect a change that was made pre-1.0.

craigmcc:
Previously, the result of field.state.errors was an array of strings. Now, it is an array of objects, from which you need to extract the "message" values. This commit updates the TanStack Form docs for React to reflect this. I am not sure of the correct syntax for Angular, Solid, or Vue, but they will have the same issues.  Just search for "errors.join" and this will locate the things that need to be changed.

Authoritative source for this change was a comment on my discussion post: https://github.com/TanStack/form/discussions/1284#discussioncomment-12510456

### DIFF
--- a/docs/framework/react/guides/validation.md
+++ b/docs/framework/react/guides/validation.md
@@ -34,7 +34,7 @@ Here is an example:
         onChange={(e) => field.handleChange(e.target.valueAsNumber)}
       />
       {field.state.meta.errors ? (
-        <em role="alert">{field.state.meta.errors.join(', ')}</em>
+        <em role="alert">{field.state.meta.errors.map(error => error.message).join(', ')}</em>
       ) : null}
     </>
   )}
@@ -65,7 +65,7 @@ In the example above, the validation is done at each keystroke (`onChange`). If,
         onChange={(e) => field.handleChange(e.target.valueAsNumber)}
       />
       {field.state.meta.errors ? (
-        <em role="alert">{field.state.meta.errors.join(', ')}</em>
+        <em role="alert">{field.state.meta.errors.map(error => error.message).join(', ')}</em>
       ) : null}
     </>
   )}
@@ -97,7 +97,7 @@ So you can control when the validation is done by implementing the desired callb
         onChange={(e) => field.handleChange(e.target.valueAsNumber)}
       />
       {field.state.meta.errors ? (
-        <em role="alert">{field.state.meta.errors.join(', ')}</em>
+        <em role="alert">{field.state.meta.errors.map(error => error.message).join(', ')}</em>
       ) : null}
     </>
   )}
@@ -267,7 +267,7 @@ export default function App() {
                 onChange={(e) => field.handleChange(e.target.valueAsNumber)}
               />
               {field.state.meta.errors ? (
-                <em role="alert">{field.state.meta.errors.join(', ')}</em>
+                <em role="alert">{field.state.meta.errors.map(error => error.message).join(', ')}</em>
               ) : null}
             </>
           )}
@@ -350,7 +350,7 @@ To do this, we have dedicated `onChangeAsync`, `onBlurAsync`, and other methods 
         onChange={(e) => field.handleChange(e.target.valueAsNumber)}
       />
       {field.state.meta.errors ? (
-        <em role="alert">{field.state.meta.errors.join(', ')}</em>
+        <em role="alert">{field.state.meta.errors.map(error => error.message).join(', ')}</em>
       ) : null}
     </>
   )}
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Thanks for your help @craigmcc 💯

Balastrong:
Those examples seem already good to me, they don't need the object handling.

The `errors` array has whatever you return from the validators. In those examples the validators are returning strings so you can already do `errors.join` and it displays the error nicely.

The other discussion was about Standard Schema validators where the return is not a string but the standard issue object. Only in those cases you need to handle the objects.

If there are examples in the docs returning objects/schemas and not handling the errors properly they need to be updated but if there aren't any... we're good :D

Balastrong:
Yes you're 100% correct! zod needs that extra step and if you find some usages of zod in the docs that do not handle the objects, those need to be updated.

However, the changes in this PR aren't on zod examples as far as I can see. Let's take [this one](https://github.com/TanStack/form/pull/1357/files#diff-455fab23b8880cb58c7a875f5666ab15daf497b6d6fc17fba4c83466b5aafaa4R37):
```tsx
<form.Field
  name="age"
  validators={{
    onChange: ({ value }) =>
      value < 13 ? 'You must be 13 to make an account' : undefined,
  }}
>
  {(field) => (
    <>
      <label htmlFor={field.name}>Age:</label>
      <input
        id={field.name}
        name={field.name}
        value={field.state.value}
        type="number"
        onChange={(e) => field.handleChange(e.target.valueAsNumber)}
      />
      {field.state.meta.errors ? (
        <em role="alert">{field.state.meta.errors.map(error => error.message).join(', ')}</em>
      ) : null}
    </>
```

Here you don't need `.map(error => error.message)` as the validator is returning the string `'You must be 13 to make an account'`.

If you paste this on stackblitz you can verify that.

---

The underlying issue I see is that we were previously forcing errors to be strings, it had limitation but was kinda easier to manage. Now that we opened up to _whatever you want_ it's infinitely more flexible it also means if you return an object, array or whatever you have to handle that yourself, hence the confusion.

craigmcc:
How is someone suppose to know that these examples were not "Zod specific"?

How is it that, if you *are* using Zod, you run into this problem when you cut-and-paste the existing syntax out of the examples (which pretty much every newbie to the library will do, since there are so many awesome examples available here) and then run into this?

craigmcc:
OK, a bit of background here: 
* I **totally** agree with the idea that the return from _field.state.meta.errors_ should be an array of objects, rather than array of strings.  That gives you folks the opportunity to add new fields as they might be useful, without creating backwards compatibility problems for people who adapt future versions.
* I **absotively posolutely** agree with having done that before your 1.0 release, so that you don't break backwards compatibility by changes like this later on.

What I don't agree with is your implied answer that "it's OK for _field.state.meta.errors_ to return something different based on what validator you are using.  What?????

So, tell me how this behavior is Zod specific?  I searched through the docs directory for Tanstack Forms, and found only a few references to Zod.  All of those were in the intro sections of the validation.md pages for various frameworks, plus one similar reference in the philosophy page.  Nothing at all in the nitty gritty details for whatever framework you choose.

By the way, can you please point me at any documentation about what _field.state.meta.errors_ actually returns?  I can get a small amount of info by what IntelliSense gives me when I reference it, but no real details.  (For example, it seems I can use onBlur there and in the form meta stuff, but it seems to do nothing in either case?  Yah, that's a separate issue). 
Documenting the Typescript types for everything is really nice -- I could not have done what you folks did to make inference work as well as it does.  But it would be really helpful to document the actual objects that get returned by the various APIs.

In the mean time, I have a bunch of apps that I have built for nonprofits since I retired a few years ago.  The web ones are using React Hook Form -- great in it's day, but I cannot get through a week now without seeing issues where a dot-dot release broke Typescript things, and people have to back up a bit.  They came late to the Typescript party, and I'm really looking forward to TSF as a path forwards, and migrating them to something that will last a bit longer that RHF.

In the short term, though, I'm not yet convinced -- both over this and a couple of other things that really are separate discussions.

Balastrong:
I think we're running in circles now.

`errors` is a union type of all the return types of all your validators.

You set onChange to return a number and onBlur to return a string, errors is an array of numbers or strings
You set a schema on onChange (that returns an object) then you'll have errors that is an array of objects

There's no "zod specific" logic or any abstraction made from the library. You return objects, you have objects. You return strings, you have strings.

I'm not sure how can I explain it in other words, that's plain javascript at the end of the day. `field.meta.errors` is an array containing the return values of your validator functions.

You can read more on the [validation](https://tanstack.com/form/latest/docs/framework/react/guides/validation) page but I'm sure you already read it.

craigmcc:
Ah ... I see what you are saying.

The answer to my question about whether the types of array elements _field.state.meta.errors_ depends on which kind of validation you are doing is actually "yes".  In my case, I'm passing in a Zod schema (at the form level), so I get back the objects that Zod returns.  Nearly all the examples on the validation page do their own validations and return strings, so strings is the right answer.

It would be useful if the "Validation through Schema Libraries" example mentions this, and even illustrates what you get back.  That's the scenario where iterating over the objects and using _message_ to get the text of the error makes sense.

PR #1359, opened 2025-03-30
fix(react-form): cookie deletion in getFormData

brhx:
Removing the response header didn't do anything as the cookie had already been set on the client. This replaces it with the h3 deleteCookie method to ensure it actually gets deleted. Before this the error message would persist whenever the page was loaded.


gioboa (approved):
Thanks for this great contribution @brhx 👏

PR #1367, opened 2025-04-02
fix(form-core): touch all fields regardless of canSubmit in `<form>.handleSubmit()`

### LINKED ISSUE #1360: `handleSubmit` does not always mark all fields as touched
### Describe the bug

When calling `<useForm>.handleSubmit()` or `<useAppForm>.handleSubmit()`, all fields being touched is not consistent when used with form-level validators.

That is because [of an early return statement](https://github.com/TanStack/form/blob/v1.2.1/packages/form-core/src/FormApi.ts#L1599) that is sensible in theory, but has problems in edge cases.

### Example
I have two fields, `requiredName` and `optional`. When using a standard schema and passing it to `<useForm>` as `onChange` validator, it will validate `requiredName` when changing `optional`. That makes complete sense.

However, that means that `canSubmit` can become false even if no required field has ever been touched. This results in submission handling returning early without marking all the fields as touched.

The Stackblitz contains examples of this behaviour.

### Use case
My use case is that I only show errors on touched fields. This prevents multiple errors to show up when modifying an optional field (due to schema validation). With `canSubmit = false`, they will now be stuck in an error state but cannot show it due to being untouched.

Is this intended behaviour?

https://stackblitz.com/edit/vitejs-vite-od8bfy6y?file=src%2FApp.tsx

### Steps to reproduce

1. Go to the stackblitz repro
2. For each form variant, change the value of the second field
3. Click the submit button (will always fail)

Expected:
* Both fields should be marked as `isTouched`
* Errors show up for field 1

Actual:
* `onChange` or `onBlur` don't mark field 1 as `isTouched`
* `onSubmit` marks field 1 as `isTouched`
* having both `onSubmit` and any of the other two, field 1 will not be `isTouched`

### Expected behavior

As user, I expect `isTouched` to be true for all fields when handling the submission, independent from which validator you used.

None

### Screenshots or Videos

_No response_

### Platform

- Windows 11
- Firefox v136.0.4

### TanStack Form adapter

react-form

### TanStack Form version

v1.2.1

### TypeScript version

v5.7.2

### Additional context

For anyone with the same problem, you can access `submissionAttempts` with
`useStore(field.form.store, state => state.submissionAttempts)`. This allows you to check if a submission has happened before, overriding any touched condition in your logic.

This state is accessible in both the callback as well as field context.

LeCarbonator:
The [early return on this line](https://github.com/TanStack/form/blob/v1.2.1/packages/form-core/src/FormApi.ts#L1599) prevents fields from being marked as touched. 

Test cases have been added to address this inconsistency. It's assumed that the following statements must be true:

* `handleSubmit()` touches all fields, even if it `canSubmit === false`
* `handleSubmit()` does not call form validators if field validators errored
* `handleSubmit()` calls form validators if field validators succeeded

### Reasoning for the first assumption
`validateAllFields` is called during `handleSubmit`, which touches all fields. However, that code is not reached if validation errored beforehand.

This can lead to results where an optional field is edited, causing an `onChange` error in a required field that is not yet touched (a form-level validation schema, for example).

The form is now invalid and will not validate all fields, despite no obvious error appearing as no required fields were touched.

This PR closes #1360 if merged.


juanvilladev:
Looks like you're completely removing the canSubmit check. Instead I think something safer would be to move the validateAllFields line above canSubmit. If we simply remove the canSubmit flag, I don't see the point in having it?

validateAllFields is what marks them all as touched.

Even then I'm not sure if maintainers would be okay with always allowing validateAllFields to run despite canSubmit being false?

LeCarbonator:
> > > Looks like you're completely removing the canSubmit check. Instead I think something safer would be to move the validateAllFields line above canSubmit. If we simply remove the canSubmit flag, I don't see the point in having it?
> > > validateAllFields is what marks them all as touched.
> > > Even then I'm not sure if maintainers would be okay with always allowing validateAllFields to run despite canSubmit being false?
> > 
> > 
> > Indeed. It's the main reason I marked it as draft, as the implemented solution isn't a solution at all. However, some of the test cases are set up, and there's still some questions left up for discussion about how the behaviour should work.
> > I think the main point of the canSubmit flag is that the user can access it in the frontend and, say, disable the submit button. Once handleSubmit is called, the question is what should always happen and what should be conditional.
> 
> Oh! Completely missed this was a draft. My bad! Nice, I understand what you mean. Perhaps some might find it weird if canSubmit is false, yet user CAN technically submit without users of the library explicitly disabling the button.

Along with the test cases, the canSubmit check should now be restored. It simply marks all fields as touched before doing so.

That leaves the other questions up for debate, but this is a start.

danharper:
Would love to see a fix landed for this - first time using tanstack form and hit this exact issue! Landed on the same solution myself - either removing the "canSubmit" early return, or explicitly touching all fields :)

---

The ["standard schema" example](https://tanstack.com/form/latest/docs/framework/react/examples/standard-schema?panel=sandbox) shows this bug pretty well, plus oddities with "canSubmit".

Type into first name, and you won't see any errors for last name until you focus & blur that field - "Submit" is disabled so you'll only know _why_ it can't be submitted until you click in & out of each field.

https://github.com/user-attachments/assets/db766533-608b-4e37-a505-1f42200d4f09

Along with removing the `disabled={!canSubmit}` logic on the button in the example, this PR fixes everything -- the other field errors are shown after clicking the button.

LeCarbonator:
> Would love to see a fix landed for this - first time using tanstack form and hit this exact issue! Landed on the same solution myself - either removing the "canSubmit" early return, or explicitly touching all fields :)
> 
> The ["standard schema" example](https://tanstack.com/form/latest/docs/framework/react/examples/standard-schema?panel=sandbox) shows this bug pretty well, plus oddities with "canSubmit".
> 
> Type into first name, and you won't see any errors for last name until you focus & blur that field - "Submit" is disabled so you'll only know _why_ it can't be submitted until you click in & out of each field.
> riewj.mov
> 
> Along with removing the `disabled={!canSubmit}` logic on the button in the example, this PR fixes everything -- the other field errors are shown after clicking the button.

I initially waited because of the quick and dirty fix I implemented and to hear more of the contributors, but since there are genuine issues coming up with this bug, I'll mark the PR ready for review. I can always make a future PR if needed :)

ibnumusyaffa:
i think its solved with the new `canSubmitWhenInvalid` form option in this PR

https://github.com/TanStack/form/pull/1234

LeCarbonator:
> i think its solved with the new `canSubmitWhenInvalid` form option in this PR
> 
> 
> 
> https://github.com/TanStack/form/pull/1234

Not quite. The linked PR addresses validation to still run even if canSubmit is false (using the provided option). This PR addresses that it should always touch all fields, regardless of canSubmit.

Balastrong (approved):
We're saying that clicking the submit button should always touch all fields?

To be honest I don't like it. To me a field is touched only if the user has actually interacted with it, which is not what is happening here.
However, `validateAllFields` is already touching everything anyway.

I believe checking on `validationAttempts > 0 || isTouched` should be the right condition to show errors without having to touch everything but until we change the isTouched behaviour (I'm not saying it's gonna happen) I believe this PR goes into the right direction making for consistency.

Thank you, great work!

LeCarbonator:
> We're saying that clicking the submit button should always touch all fields?
> 
> 
> 
> To be honest I don't like it. To me a field is touched only if the user has actually interacted with it, which is not what is happening here.
> 
> However, `validateAllFields` is already touching everything anyway.
> 
> 
> 
> I believe checking on `validationAttempts > 0 || isTouched` should be the right condition to show errors without having to touch everything but until we change the isTouched behaviour (I'm not saying it's gonna happen) I believe this PR goes into the right direction making for consistency.
> 
> 
> 
> Thank you, great work!

True! However, I see the submit button as "I am sending what I have seen and written your way". If a field is untouched, submitting would therefore imply that the user saw no reason to touch it and is okay with the given default.

Issue #1336, opened 2025-03-26
errorMap can be undefined when using setFieldMeta updater function after a pushFieldValue


Ganbin:
### Describe the bug

I have an array field that can be populated dynamically based on some code. 

I use the helpful `form.pushFieldValue` followed by a `form.setFieldMeta` in order to tell the form that those fields are touched and dirty.

Something like that:

```ts
form.pushFieldValue('people', { name: 'person1', age: 18 });
form.setFieldMeta('people[0].name', (prev) => ({
  ...prev,
  isDirty: true,
  isTouched: true,
}));
form.pushFieldValue('people', { name: 'person2', age: 20 });
form.setFieldMeta('people[1].name', (prev) => ({
  ...prev,
  isDirty: true,
  isTouched: true,
}));
```

When doing so, I have an runtime error when I submit my form : 
```
FieldApi.ts:1382 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'onChange') at validateFieldFn (FieldApi.ts:1382:30)
```

After debuggin i realized that the updater of the `setFieldMeta` doesn't always have the `errorMap` property defined.

I have found a workaround by doing something like that: 

```ts
(prev) => ({
  ...prev,
  errorMap: prev?.errorMap || {},
  isDirty: true,
  isTouched: true,
}),
```

But of course this gives me some TS errors because it consider that `errorMap` is defined and that the `?` and `||` are not needed. And it definitely feels more like a "hack-around"

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-form-xuuktcwd?file=src%2Findex.tsx

### Steps to reproduce

1. Go to the given stackblitz.
2. open the console.
3. Click the Add People button.
4. Submit the form.
5. See the console error.

### Expected behavior

As a user I expect that I can set the fieldMeta or directly when pushing the data with `pushFieldValue` or directly after with `setFieldMeta`.

This is important because the fields are populated by data from the user and it must be set a touched and dirty else other mechanism could override those values.

### How often does this bug happen?

Every time

### Platform

- OS: MacOS
- Brave Browser
- Version: 1.1.0

### TanStack Form adapter

react-form

### TanStack Form version

1.1.0

### TypeScript version

5.7.0

Ganbin:
I just see #1323 and the #1324 PR that seems to be related and will surely fix this bug too.

juanvilladev:
@Ganbin confirmed this will be resolved in my PR.

[...]

PR #1370, opened 2025-04-03
fix(form-core): fix broken sync/async validation logic

juanvilladev:
Previously we were using a cumulative field errors map maintained at the form api level.
This PR removes that data structure in favor of a errorSourceMap data structure that lives within the FieldMeta for each field. This allows shifting/removing/deleting fields to automatically keep the errorSourceMap in sync.

This will fix https://github.com/TanStack/form/issues/1323 and prevent other bugs from being raised around validation not working after other array actions like removing, shifting, swapping, moving, etc...

Async validation & sync validation now is fully synced and thoroughly tested with the new errorSourceMap. This reduces the complexity of using the cumulative errors map significantly and fixes all the issues with validations running on the wrong fields when removing/inserting/moving/shifting array fields.

Please review when you have time!

Thanks!

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -4,7 +4,12 @@ import {
   standardSchemaValidators,
 } from './standardSchemaValidator'
 import { defaultFieldMeta } from './metaHelper'
-import { getAsyncValidatorArray, getBy, getSyncValidatorArray } from './utils'
+import {
+  determineFieldLevelErrorSourceAndValue,
+  getAsyncValidatorArray,
+  getBy,
+  getSyncValidatorArray,
+} from './utils'
 import type { DeepKeys, DeepValue, UnwrapOneLevelOfArray } from './util-types'
 import type {
   StandardSchemaV1,
@@ -25,6 +30,7 @@ import type {
   ValidationCause,
   ValidationError,
   ValidationErrorMap,
+  ValidationErrorMapSource,
 } from './types'
 import type { AsyncValidator, SyncValidator, Updater } from './utils'
 
@@ -561,6 +567,10 @@ export type FieldMetaBase<
     UnwrapFieldValidateOrFn<TName, TOnSubmit, TFormOnSubmit>,
     UnwrapFieldAsyncValidateOrFn<TName, TOnSubmitAsync, TFormOnSubmitAsync>
   >
+  /**
+   * @private allows tracking the source of the errors in the error map
+   */
+  errorSourceMap: ValidationErrorMapSource
   /**
    * A flag indicating whether the field is currently being validated.
    */
@@ -1101,6 +1111,11 @@ export class FieldApi<
               ...prev,
               // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
               errorMap: { ...prev?.errorMap, onMount: error },
+              errorSourceMap: {
+                // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+                ...prev?.errorSourceMap,
+                onMount: 'field',
+              },
             }) as never,
         )
       }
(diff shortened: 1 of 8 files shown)


juanvilladev:
> It would be great to stop saving the state of fields at the form level. This leads to duplication and divergence of states.
> 
> 
> 
> For example, how does it do
> 
> [modular-forms](https://github.com/fabian-hiller/modular-forms/blob/9bdade59fbf584ada0289a8b3734c5117ab7f0a2/packages/solid/src/methods/getErrors.ts#L116)

Correct, that's why this PR removes the cumulativeErrorsMap which is at the form level and moves the error source to the fieldMeta. This way when the fieldMeta is shifted, moved, deleted, etc... the source remains valid.

teleskop150750:
> > It would be great to stop saving the state of fields at the form level. This leads to duplication and divergence of states.
> > For example, how does it do
> > [modular-forms](https://github.com/fabian-hiller/modular-forms/blob/9bdade59fbf584ada0289a8b3734c5117ab7f0a2/packages/solid/src/methods/getErrors.ts#L116)
> 
> Correct, that's why this PR removes the cumulativeErrorsMap which is at the form level and moves the error source to the fieldMeta. This way when the fieldMeta is shifted, moved, deleted, etc... the source remains valid.

the same approach could be applied for the rest of the state that relates to the field, including meta. The form could have excluded the other information from the fields. In extreme cases, it would be possible to duplicate it at the field and form levels (write it in two places at the same time).

packages/form-core/src/utils.ts
```diff
@@ -376,3 +380,30 @@ export function shallow<T>(objA: T, objB: T) {
...
+ */
+export const determineErrorValue = ({
+  newFormValidatorError,
+  isPreviousErrorFromFormValidator,
+  previousErrorValue,
+}: {
+  newFormValidatorError: ValidationError
+  isPreviousErrorFromFormValidator: boolean
+  previousErrorValue: ValidationError
+}) => {
```
Balastrong, line 396: Return type could be an explicit `{ newErrorValue: ValidationError, newSource: ValidationSource | undefined }`

packages/form-core/src/FieldApi.ts
```diff
@@ -1387,6 +1397,14 @@ export class FieldApi<
                 // Prefer the error message from the field validators if they exist
                 error ? error : errorFromForm[errorMapKey],
             },
+            errorSourceMap: {
+              ...prev.errorSourceMap,
+              [getErrorMapKey(validateObj.cause)]: error
```
Balastrong, line 1402: Can we do that here and a couple of lines above?

```suggestion
              [errorMapKey]: error
```

packages/form-core/tests/FieldApi.spec.ts
```diff
@@ -362,7 +364,7 @@ describe('field api', () => {
     ])
   })
 
-  it('should remove a subfield from an array field correctly', async () => {
+  it('should remove a subfield from an errorMap: {}array field correctly', async () => {
```
Balastrong, line 367: Pasted this thing here by mistake?

packages/form-core/src/utils.ts
```diff
@@ -376,3 +380,30 @@ export function shallow<T>(objA: T, objB: T) {
...
+    return { newErrorValue: newFormValidatorError, newSource: 'form' }
+  }
+
+  // Clears form level error since it's now stale
+  if (isPreviousErrorFromFormValidator) {
+    return { newErrorValue: undefined, newSource: undefined }
+  }
+
+  // Do not clear field level validators since they're handled by the field validator
+  return { newErrorValue: previousErrorValue, newSource: 'field' }
```
Balastrong, line 408: I think there's an extra case here, if `previousErrorValue` is `undefined` we should probably also return `newSource: undefined`

packages/form-core/src/utils.ts
```diff
@@ -376,3 +380,30 @@ export function shallow<T>(objA: T, objB: T) {
   }
   return true
 }
+
+/**
+ * Determines the logic for determining the error value to set on the field meta within the form sync validation.
```
Balastrong, line 385: I believe it's not only for sync validation

Balastrong (commented):
Good call on moving this to the field meta, I like it!

packages/form-core/src/FieldApi.ts
juanvilladev: Nice! Can't believe I didn't see that! Updated.

packages/form-core/tests/FieldApi.spec.ts
juanvilladev: Whoops, yes, updated!

packages/form-core/src/utils.ts
juanvilladev: Correct! Updated description to mention both sync/async for both utility functions.

juanvilladev:
@Balastrong thanks for the review! I believe I've resolved all of the raised comments. Additionally, I created an equivalent function for determining the new error/error source at the field level validator functions (sync/async). This way its consistent for both form level/field level!

Additionally added tests for the new function and updated existing tests to ensure we cover the missed case you mentioned.

juanvilladev:
Was manually testing #1336 and realized that for that issue the error comes from no optional chain operator being used when checking for the errorMapKey inside the errorMap dynamically:
```ts
 if (field.state.meta.errorMap[errorMapKey] !== newErrorValue)
```

Updated it to:
```ts
 if (field.state.meta.errorMap?.[errorMapKey] !== newErrorValue)
```

Had to add for both:
        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition

Not sure why TS doesn't infer that the value could be undefined ¯\(ツ)/¯

packages/form-core/src/types.ts
```diff
@@ -35,6 +35,23 @@ export type ValidationErrorMap<
...
+  onBlur?: ValidationSource
+  onSubmit?: ValidationSource
+  onServer?: ValidationSource
+}
+
+const a: ValidationErrorMapSource = {
+  onChange: 'field',
+}
+
+const test = a.onBlur
```
Balastrong, line 53: 🧹

Balastrong (approved):
There's just one tiny leftover from your tests probably, everything else looks good! :D

Speaking of the optional errorMap that would be an interesting investigation if you want, to validate if we should fix it on a type level (if we accept it might be undefined) or actually on the code to prevent it being undefined at all times. That's for a new PR anyway :)

packages/form-core/src/types.ts
juanvilladev: @Balastrong done! Sorry was out for a couple of weeks for work. Just cleaned up the unused code above. Thanks!

juanvilladev:
Squashed commits to a single one, unsure why autofix didn't append the documentation changes?

juanvilladev:
> There's just one tiny leftover from your tests probably, everything else looks good! :D
> 
> Speaking of the optional errorMap that would be an interesting investigation if you want, to validate if we should fix it on a type level (if we accept it might be undefined) or actually on the code to prevent it being undefined at all times. That's for a new PR anyway :)

Updated! Thanks!

Balastrong (approved):
LGTM, thanks!

PR #1371, opened 2025-04-03
fix(vue-form): losing reactivity

### LINKED ISSUE #1276: Please fix the implementation in vue
Please provide the ability to use [`toValue`](https://vuejs.org/api/reactivity-utilities.html#tovalue) to pass reactive parameters.

Why in solidjs parameters are passed in callback, at the same time Vue, which is also built on the signal system, does not have such a possibility. This requires writing additional abstractions that worsen dx and affect performance.

Example
[Vue js](https://tanstack.com/form/latest/docs/framework/vue/guides/async-initial-values#basic-usage)

```vue
<script setup lang="ts">
import { useForm } from '@tanstack/vue-form'
import { useQuery } from '@tanstack/vue-query'
import { watchEffect, reactive } from 'vue'

const { data, isLoading } = useQuery({
  queryKey: ['data'],
  queryFn: async () => {
    await new Promise((resolve) => setTimeout(resolve, 1000))
    return { firstName: 'FirstName', lastName: 'LastName' }
  },
})

const firstName = computed(() => data.value?.firstName || '')
const lastName = computed(() => data.value?.lastName || '')

const defaultValues = reactive({
  firstName,
  lastName,
})

const form = useForm({
  defaultValues,
  onSubmit: async ({ value }) => {
    console.log(value)
  },
})
</script>

<template>
  <p v-if="isLoading">Loading..</p>
  <form v-else @submit.prevent.stop="form.handleSubmit">
    <!-- ... -->
  </form>
</template>
```

[SolidJS](https://tanstack.com/form/latest/docs/framework/solid/guides/async-initial-values#basic-usage) 

```ts
import { createForm } from '@tanstack/solid-form'
import { createQuery } from '@tanstack/solid-query'

  const { data, isLoading } = createQuery(() => ({
    queryKey: ['data'],
    queryFn: async () => {
      await new Promise((resolve) => setTimeout(resolve, 1000))
      return { firstName: 'FirstName', lastName: 'LastName' }
    },
  }))

  const form = createForm(() => ({
    defaultValues: {
      firstName: data?.firstName ?? '',
      lastName: data?.lastName ?? '',
    },
    onSubmit: async ({ value }) => {
      console.log(value)
    },
  }))

  if (isLoading) return <p>Loading..</p>

  return null
}
```

The same problem exists in other tanstack libraries

An example of how [pinia-colada](https://pinia-colada.esm.dev/cookbook/migration-tvq.html#Using-ref-and-computed-in-queryKey) fixes the shortcomings of vue-query

[Discussion about signal api design](https://github.com/vuejs/rfcs/discussions/747) in Vue

### LINKED ISSUE #1503: fix(vue-form): losing reactivity
### Describe the bug

The Vue useForm composable requires a workaround to handle asynchronous default values, which is functional but suboptimal for Vue’s reactivity system. The documentation recommends creating computed properties for each field and wrapping them in a reactive object to maintain reactivity, see https://tanstack.com/form/latest/docs/framework/vue/guides/async-initial-values. For example:
```ts
const firstName = computed(() => data.value?.firstName || '');
const lastName = computed(() => data.value?.lastName || '');
const defaultValues = reactive({ firstName, lastName });

const form = useForm({
  defaultValues,
  onSubmit: async ({ value }) => {
    console.log(value);
  },
});
```

While this works, it is not ideal, especially for forms with many fields, as it requires creating a computed property per field and a reactive wrapper. This may impact performance due to the overhead of multiple reactive proxies.
In contrast, the Solid and Svelte adapters use a getter-based API for createForm, which preserves reactivity without requiring individual reactive wrappers for each field. For example:
Solid and Svelte Adapter:
```ts
const form = createForm(() => ({
  defaultValues: {
    firstName: data?.firstName ?? '',
    lastName: data?.lastName ?? '',
  },
  onSubmit: async ({ value }) => {
    console.log(value);
  },
}));
```

https://tanstack.com/form/latest/docs/framework/vue/guides/async-initial-values

### Steps to reproduce

Use the same code from the docs

### Expected behavior

useForm should accept a callback function for its entire configuration, similar to Solid’s createForm. For example:
```ts
const form = useForm(() => ({
  defaultValues: {
    firstName: data?.firstName ?? '',
    lastName: data?.lastName ?? '',
    email: data?.email ?? '',
    phone: data?.phone ?? '',
  },
  onSubmit: async ({ value }) => {
    console.log(value);
  },
```

A pending PR [TanStack/form#774](https://github.com/TanStack/form/pull/1371) addresses this.

Every time

### Screenshots or Videos

_No response_

### Platform

All platforms

### TanStack Form adapter

vue-form

### TanStack Form version

1.11.1

### TypeScript version

_No response_

### Additional context

Related discussion: https://github.com/TanStack/form/discussions/1278

teleskop150750:
close #1503
close #1515
close #1276

Fix of loss of reactivity
Refactor Field: improve granular updates and lazy slot props

### DIFF
--- a/docs/framework/vue/quick-start.md
+++ b/docs/framework/vue/quick-start.md
@@ -29,7 +29,7 @@ const form = useForm(() => ({
           <template v-slot="{ field, value }">
             <input
               :name="field.name"
-              :value="value()"
+              :value="field.state.value"
               @blur="field.handleBlur"
               @input="(e) => field.handleChange((e.target as HTMLInputElement).value)"
             />


teleskop150750:
It might be necessary to put the entire api into signal

[Link](
https://github.com/TanStack/form/blob/a1568ff667ef6afe8aff576bdc1079327701a6de/packages/solid-form/src/createField.tsx#L217)

j0rgedev:
Would it be possible to watch default values key so it can update when it changes? This could be useful when fetching data, instead of creating a lot of computed properties like the example

j0rgedev:
Hey @teleskop150750. Do you have any workaround until this is merged? I think it hasn't been reviewed yet since the main issue was closed

j0rgedev:
Created the issue #1503 to track this PR 👍🏻

LeCarbonator:
It's great to have the explanation listed in the linked issue, but even with a perfect PR, this is a breaking change. It would be delayed until a v2 release.

This also appears to be a (major) improvement on Vue's current reactivity system rather than 'fixing' a loss of reactivity. Please correct me if I'm wrong.

teleskop150750:
> It's great to have the explanation listed in the linked issue, but even with a perfect PR, this is a breaking change. It would be delayed until a v2 release.
> 
> This also appears to be a (major) improvement on Vue's current reactivity system rather than 'fixing' a loss of reactivity. Please correct me if I'm wrong.

The implementation is synchronized with the solidjs version. In signal-based frameworks, it is necessary to pass reactive data in callbacks. There are other workarounds, such as a getter object field or proxy. But they are more complex and inconvenient.

For some reason, SolidJs makes such mistakes less often and passes parameters correctly. In Vue, this causes difficulties every time because Ref is confusing

https://github.com/vuejs/rfcs/discussions/747

teleskop150750:
The current implementation of Vue lacks wathEffect for parameters. So updating parameters does not update the state.

UPD:

teleskop150750:
You can try to rewrite the PR to keep the API. I wanted to bring it closer to SolidJs in advance because it is more flexible and less likely to be subject to changes in the future.

LeCarbonator:
> The current implementation of Vue lacks withEffect for parameters. So updating parameters does not update the state.

Can you elaborate on this? Can that be fixed with the current API until v2 (maybe) migrates to this signal-based one?

teleskop150750:
Solif js [Update FormApi](https://github.com/TanStack/form/blob/117b7432d57a9100b1780a63604b4297adaa7e02/packages/solid-form/src/createForm.tsx#L216-L221)

VueJS [No dependency tracking](https://github.com/TanStack/form/blob/117b7432d57a9100b1780a63604b4297adaa7e02/packages/vue-form/src/useForm.tsx#L285-L287)

example fix

https://vuejs.org/api/reactivity-utilities.html#tovalue

```ts
function useForm(opts?: MaybeRefOrGetter<Options>) {
  const api = new FormApi(opts?.()) 
  let isMouned = false

  watchEffect(() => {
     const optsVal = toValue(opts)
     if (isMouned) {
       api.update(optsVal)
     }
  })
  
  onMounted(() => {
    isMouned = true
    formApi.mount()
  })
}
```

teleskop150750:
I made additional edits to FormField. I don't remember the exact reason. I think it was to reduce the number of updates.

crutchcorn:
Closing as this would be a v2 breaking change and one I'm not entirely convinced we need. Issue will remain open and we have a back reference to this fix if needed.

PR #1375, opened 2025-04-04
feat(FieldApi): allow debounce of onChange and onBlur listener

Pascalmh:
This allows you to listen to a `change`/`blur` event of a Field after a denounce you can set in millisecond.

It makes situations possible where you want to introduce a side effect after the user stopped typing or left the field for some time.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -349,7 +349,9 @@ export interface FieldListeners<
   TData extends DeepValue<TParentData, TName> = DeepValue<TParentData, TName>,
 > {
   onChange?: FieldListenerFn<TParentData, TName, TData>
+  onChangeDebounceMs?: number
   onBlur?: FieldListenerFn<TParentData, TName, TData>
+  onBlurDebounceMs?: number
   onMount?: FieldListenerFn<TParentData, TName, TData>
   onSubmit?: FieldListenerFn<TParentData, TName, TData>
 }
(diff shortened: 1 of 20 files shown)


packages/form-core/src/FieldApi.ts
```diff
@@ -349,7 +349,9 @@ export interface FieldListeners<
   TData extends DeepValue<TParentData, TName> = DeepValue<TParentData, TName>,
 > {
   onChange?: FieldListenerFn<TParentData, TName, TData>
+  onChangeDebounceMs?: number // Added debounce option for onChange
   onBlur?: FieldListenerFn<TParentData, TName, TData>
+  onBlurAsyncDebounceMs?: number // Added debounce option for onBlur
```
Pascalmh, line 354:
```suggestion
  onChangeDebounceMs?: number
  onBlur?: FieldListenerFn<TParentData, TName, TData>
  onBlurAsyncDebounceMs?: number
```

LeCarbonator:
The naming between the debounce properties seems inconsistent. `onChangeDebounceMs` vs. `onBlurAsyncDebounceMs`.
Since the debounce should ignore async vs. sync, the blur should be renamed to `onBlurDebounceMs`.

Balastrong (commented):
I see you're using the existing `timeoutIds` but isn't that dangerous?

I mean, wouldn't it break if you have both a debounce on the validator and on the listener since they're sharing the same timeoutId?

Pascalmh:
> wouldn't it break if you have both a debounce on the validator and on the listener since they're sharing the same timeoutId?

Good catch - I've separated the validation and listener timeoutIds

Balastrong (approved):
LGTM, also thanks for adding the docs!

Issue #1363, opened 2025-04-01
Non-index keys in React causing inputs to be uncontrolled when removing array fields


juanvilladev:
### Describe the bug

When mapping through array fields, using static keys (non-changing) such as indexes, results in:

"A component is changing a controlled input to be uncontrolled. This is likely caused by the value changing from a defined to undefined, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://react.dev/link/controlled-components"

### Your minimal, reproducible example

https://stackblitz.com/edit/vitejs-vite-tayuhyi2?file=package.json

### Steps to reproduce

Add rows > open console > remove any row

### Expected behavior

We should not receive any error messages from React regarding controlled/uncontrolled components.

### How often does this bug happen?

Every time

### Platform

TS 5.7, Tanstack Form 1.2.3.

### TanStack Form adapter

react-form

### TanStack Form version

1.2.3

### TypeScript version

5.7.2

### Additional context

This PR #1319, I believe was supposed to fix this issue, however we can still replicate it on version 1.2.3 which is supposed to have the fix.
Edit: Actually looks like #1319 was not targeting this issue so this is still valid. Other users have also faced this issue on the discord: 
https://discord.com/channels/719702312431386674/1100437019857014895/1353165162953314319

juanvilladev:
Note, this issue is actually still leading to reading "errorMap" when its undefined when setting field meta. The underlying issue is that at some points when we call setFieldMeta, prev is empty! Not sure why this is the case, I'm pretty sure its related to this issue.

PR #1378, opened 2025-04-05
fix(react-form): fallback array field value during index change to prevent uncontrolled error

MVaik:
Fixes https://github.com/TanStack/form/issues/1363

More of a draft, while this seems to fix the issue it could have unintended side-effects.
Perhaps it's useful in finding a better solution.

What seems to happen:
Have at least 2 array elements that are not keyed by index.
Remove the first element.
This will cause the second element to be moved up, but it attempts to retrieve a value with the old index, resulting in an undefined value for a short bit.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -993,11 +993,16 @@ export class FieldApi<
     this.form = opts.form as never
     this.name = opts.name as never
     this.timeoutIds = {} as Record<ValidationCause, never>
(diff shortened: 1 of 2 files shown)


juanvilladev:
The problem with this is that, we're only addressing the specific value field. This is also a problem with the field meta, and all other values. :/

juanvilladev:
Hey @MVaik I just opened a PR with a proper fix for this. Feel free to take a look at it!

MVaik:
Closing in favor of https://github.com/TanStack/form/pull/1440

PR #1401, opened 2025-04-10
docs: suggest using `satisfies` instead of `as`

cxa:
At least `satisfies` do type checking for us.

### DIFF
--- a/docs/philosophy.md
+++ b/docs/philosophy.md
@@ -56,7 +56,7 @@ useForm({
   defaultValues: {
     name: 'Bill Luo',
     age: 24,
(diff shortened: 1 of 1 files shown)


gioboa (approved):
It looks great to me :clap:

LeCarbonator:
Related PR: #1417

cxa:
Close as the docs will be changed to use explicit type without casting.

harry-whorlow:
@cxa thanks for the PR! we decided to go with type annotations as using satisfies fails with optional types. But the suggestion is appreciated!

Issue #1229, opened 2025-03-06
The installation size is large due to the dependency on `@remix-run/node`.


L4Ph:
The current install size of `@tanstack/react-form` is 24MB.
https://packagephobia.com/result?p=@tanstack/react-form

![Image](https://github.com/user-attachments/assets/63ec8620-3027-4a93-8c15-2575a85ad563)

Someone has already asked me about this on my behalf and I understand that you have also responded about it.
https://x.com/tannerlinsley/status/1897323043510431970
> It's 11.8kb min. Don't use packagephobia or bundlephobia.

## What's the problem?
I see the problem as being that it is just a form and a package of wrappers for it, which is such a large installation size.
Just doing the `npm i @tanstack/react-form` generates about 30 MB of communication.
This is not very desirable for CI or for developers who have pipelines in place for builds.

In fact, I have not been able to employ this in a new product I am creating because I am afraid of it.

## What is this package used for?
That's it. Just those three lines.
https://packagephobia.com/result?p=%40remix-run%2Fnode

```ts
import { createCookie } from '@remix-run/node'

export const _tanstackInternalsCookie = createCookie('_tanstack_form_internals')
```

https://github.com/TanStack/form/blob/main/packages%2Freact-form%2Fsrc%2Fstart%2Futils.ts

This improvement alone will reduce the package installation size to about 1 MB.
Perhaps hopefully the post-bundle size can be further reduced as well.

crutchcorn:
This is great research, we'll remove that dep in favor of something else.

At the same time please do not tag specific maintainers - it's in poor taste.

L4Ph:
@crutchcorn 
Tanner was the one who answered this in X, and since I was new to sending an issue to a repository of this size and organization, I felt it would be quicker to ask Tanner, so I was menshoning...
I am so sorry...
I will turn it off.

ryoppippi:
![Image](https://github.com/user-attachments/assets/e2787528-2317-46fb-84cc-7ee7023ce858)

https://packagephobia.com/result?p=@tanstack/react-form

For everyone in this thread, this is a great visualization!

PR #1402, opened 2025-04-10
fix: remove `@remix-run` as a dependency to reduce the install size 20x smaller!

ryoppippi:
Hi!
This PR, I gonna remove `@remix-run/node` to reduce the install size.
I reimplement the cookie function based on remix's implementation.
Edit: instead and copy-n-paste the algorighm from remix-run, I use `devalue` for serializeation

fixes: https://github.com/TanStack/form/issues/1229

Because there is no test for this code, I cannot guarantee that it works. So please check it out


ryoppippi:
oh wait I can simplify it more!

ryoppippi:
fixed with devalue!

packages/react-form/src/start/utils.ts
```diff
@@ -1,3 +1,15 @@
-import { createCookie } from '@remix-run/node'
+import { getCookie, setCookie } from '@tanstack/react-start/server'
+import * as devalue from 'devalue'
 
-export const _tanstackInternalsCookie = createCookie('_tanstack_form_internals')
+export const _tanstackInternalsCookie = {
+  name: 'tanstack-internals',
```
Balastrong, line 5: Let's keep `_tanstack_form_internals` as name

packages/react-form/src/start/utils.ts
```diff
@@ -8,5 +8,8 @@ export const _tanstackInternalsCookie = {
     if (!cookie) return undefined
     return devalue.parse(cookie)
   },
-  serialize: (data: any) => devalue.stringify(data),
+  save: (data: any) => {
+    const cookie = devalue.stringify(data)
```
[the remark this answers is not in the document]
Balastrong: devalue seems like a good idea, I'd keep it

packages/react-form/src/start/utils.ts
```diff
@@ -1,3 +1,15 @@
-import { createCookie } from '@remix-run/node'
+import { getCookie, setCookie } from '@tanstack/react-start/server'
+import * as devalue from 'devalue'
 
-export const _tanstackInternalsCookie = createCookie('_tanstack_form_internals')
+export const _tanstackInternalsCookie = {
```
Balastrong, line 4: Methods here should be `get` and `set`, right?

packages/react-form/src/start/utils.ts
Balastrong, line 5: An extra consideration on the name declared this way in the object, is it because we want it to be accessible from the outside? The only line we need it is where we delete the cookie.

Since we're wrapping the cookie handling here it might make sense to also add the `delete` method on this object and have a non exported `const INTERNALS_COOKIE_NAME` on this file, so that then internal cookie is only handled through these methods.

What do you think?

ryoppippi: done on f443ca3a04d2 !

ryoppippi: Thank you for your review.
I implemented this way to keep the original class structure. 
So, for improvements
- declare cookie name in `utils.ts` and we don't export it.
- create independent functions like `setTanstackInternalCookie` `getInternalCookie` and `deleteInternalCookie` 

This is my plan. If it looks good to you, I gonna start modifying the codebase

packages/react-form/src/start/utils.ts
ryoppippi: idk coz getter and setter has performance drawback when we use in V8. getter/setter is really ergonomic but it is internal function so I avoided it

Balastrong: I'm also fine with parse/save but right now you defined parse/save in this file but in the other files you're using get/set which obviously makes the compiler unhappy :)

packages/react-form/src/start/utils.ts
Balastrong: `_tanstack_form_internals` with `form` in it please, just in case other tanstack libraries want to use a similar name

Balastrong: `[set|get|delete]TanStackInternalCookie` sounds good, thank you 👍

ryoppippi: done on fb6a47135f23 !

ryoppippi: ok I gonna change it!

ryoppippi: done with bda493dd0da7

packages/react-form/src/start/utils.ts
ryoppippi: yeah off topic! Thanks

packages/react-form/src/start/getFormData.tsx
```diff
@@ -1,4 +1,7 @@
-import * as utils from './utils'
+import {
+  deleteInternalTanStackCookie,
+  getInternalTanStackCookie,
+} from './utils'
```
ryoppippi, line 4: ok prefers name import. I remind it when I send a PR next time

Balastrong (approved):
Looks great, thank you so much!

gioboa (approved):
Thanks for this great improvement @ryoppippi 👏

PR #1404, opened 2025-04-11
feat(form-core): add array method `field.clearValues` and `form.clearFieldValues`

Pascalmh:
#### Problem

When setting a field back to an empty array 

```ts
const form = new FormApi({
  defaultValues: {
    items: [
        { firstName: "" }
    ],
  },
});

form.setFieldValue('items', []); <-- reset
```

`items[0].firstName` would still exist - when there's a validation on the firstName (on the Field-level) `form.canSubmit` would be false.

#### Solution

Adding array method `field.clearValues` and `form.clearFieldValues`

```diff
- form.setFieldValue('items', []);
+ form.clearFieldValues("items")
```


LeCarbonator:
The library already adds multiple helper functions for arrays (`pushValue()`, `insertValue()`, `swapValues()`) because the meta has to be carefully accounted for during the update.

Following this principle, I suggest there should be one (or optionally two) helper methods added in this PR:

# `clearValues()`
Clear the values from the array.

This technically can be done as edge case check in `setFieldValue()`, but I think this follows the established array methods better.

# `filterValues(callback)`
Filter the values in the array based on this condition

The helper function would keep track of which indeces were removed and update the meta accordingly.

LeCarbonator:
The provided test case will always fail because the `items` field starts with the string `items`. This is an easy fix though, so not that big of a deal.

Pascalmh:
I'll rewrite the tests to use `clearValues` instead and change to assertions to be more explicit

LeCarbonator:
Sounds good. I got an implementation of clearValues ready for testing as well

LeCarbonator:
Since we're adding two new methods, the title should be changed to feat(form-core) so that the bot creates the proper semantic versioning

LeCarbonator:
How should `onChange` validation work in `filterValues`?

Currently:
* `swapValues` calls it on `array`, `array[fromIndex]` and `array[toIndex]`
* `removeValue` calls it on the array and all fields starting from `index`
* `insertValue` calls it on the array and all fields starting from `index`

With this information, I deduced that
* `clearValues` only needs to call it on the array itself

Should `filterValues` do:
* array + validation starting from the first index that has a shifted meta
* array + validation on all fields
* array
?

packages/form-core/tests/FieldApi.spec.ts
```diff
@@ -1346,6 +1346,28 @@ describe('field api', () => {
...
+
+    const field = new FieldApi({
+      form,
+      name: 'name',
+    })
+
+    field.mount()
+
+    expect(() => field.clearValues()).not.toThrow()
   })
```
harry-whorlow, line 1371: small nit... but since this test is testing that the logic of form.clearFieldValue, which field.clearValues is a wrapper of, dose not throw should the throw test be located in FormApi? what do you think?

LeCarbonator: I don't mind either solution. If clearValues moves into a different direction, then code coverage would pick up that it no longer checks the form api's edge case.

harry-whorlow:
LGTM 🚀

PR #1406, opened 2025-04-11
fix: type test regressions & test refactor

Balastrong:
Refactoring the tests to use `expectTypeOf` instead of `assertType` as it now errors when a type is `never`.

Let's do that on util-types first to also fix the regressions, the complete refactor for all type tests will come on another PR.


Balastrong:
Superseded by #1411

PR #1408, opened 2025-04-12
Remove redundant create server validation

theVedanta:
I was working on a server validation errors populated into field errors issue and found that there is some redundant code with minimal opportunity for customization in NextJS and Remix Server validation. Removed them, and replaced with "server" directory and also updated the docs wherever needed.


docs/framework/react/guides/ssr.md
```diff
@@ -255,8 +255,8 @@ Finally, we'll use `someAction` in our client-side form component.
 'use client'
 
 import { useActionState } from 'react'
-import { initialFormState } from '@tanstack/react-form/nextjs'
-// Notice the import is from `react-form`, not `react-form/nextjs`
+import { initialFormState } from '@tanstack/react-form/server'
+// Notice the import is from `react-form`, not `react-form/server`
```
harry-whorlow, line 259: I know that was just an updated import... but the comment doesn’t make sense since it is imported from react-form/server, perhaps there was a mistake somewhere 😄

packages/form-core/src/FormApi.ts
```diff
@@ -298,6 +298,15 @@ export interface FormOptions<
...
 
+  /**
+   * Specifies which validation type to use on the first submission attempt
+   */
+  validationOnFirstAttempt?: ValidationCause
+
+  /**
+   * Specifies which validation type to use on subsequent submission attempts
+   */
+  validationOnConsequentAttempts?: ValidationCause
```
harry-whorlow, line 309: I think you created a new branch from your other pr, I would separate the PR's

packages/form-core/src/FormApi.ts
```diff
@@ -1609,7 +1618,13 @@ export class FormApi<
...
     }
 
-    await this.validateAllFields('submit')
+    //
+    const validationType =
+      this.state.submissionAttempts === 1
+        ? (this.options.validationOnFirstAttempt ?? 'submit')
+        : (this.options.validationOnConsequentAttempts ?? 'submit')
+
+    await this.validateAllFields(validationType)
```
harry-whorlow, line 1627: same as above (separate the PR's)

packages/form-core/src/FormApi.ts
```diff
@@ -1620,7 +1635,7 @@ export class FormApi<
       return
     }
 
-    await this.validate('submit')
+    await this.validate(validationType)
```
harry-whorlow, line 1638: and this 😄

docs/framework/react/guides/ssr.md
theVedanta: I think this refers to the line below

packages/form-core/src/FormApi.ts
theVedanta: oh shit sorry, my mistake. I about the other branches 💀

packages/form-core/src/FormApi.ts
theVedanta: Will do.

theVedanta:
Addressed all the mistakes. Thank you for pointing them out!

harry-whorlow:
Thank you for taking the time for doing this, but unfortunately I'm going to have to close this PR. I talked to the other maintainers and we feel that this is one of those odd cases where its best to have duplicated code. 

Our reasoning: 
- if the api changes in either remix or next then we will have to separate this out again
- it could be confusing with Remix and Next having /server, then /start for TanStack start
- the compartmentalisation of the adapters makes it easier to reason about in context of their framework, i.e. I'm not thing about Next when I'm working on the Remix code

I hope this makes sense.

We really appreciate the effort though! unfortunately its just one of those weird ones where yes it's duplicated but it's "the lesser of two evils". We're looking forward to seeing you in the PR's again soon though 🚀

theVedanta:
Hey, thank you for letting me know!

It's a good insight and taught me something new as well. I appreciate and agree with your reasoning.

Yes of course!

PR #1413, opened 2025-04-14
docs: add ESLint rule exception for react/no-children-prop

KwibooReese:
I ran into an error caused by ESLint and I have found that a few other people did too, I think it would be good to include this in the docs.

### DIFF
--- a/docs/framework/react/guides/basic-concepts.md
+++ b/docs/framework/react/guides/basic-concepts.md
@@ -72,6 +72,18 @@ Example:
   )}
 />
 ```
+If you are using ESLint, make sure to add this to your eslintrc to avoid linting errors.
+```json
+  "rules": {
+    "react/no-children-prop": [
+      true,
+      {
+        "allowFunctions": true
+      }
+    ],
+  }
+```
+
 
 ## Field State
 


Pascalmh:
Thoughts:
- you do not have to write the code the same as in the docs/examples - you can keep the eslint-rule turned on
- it is just one linting tool, would the docs end up listing all?
- [the rule is disabled in the tanstack-form repo](https://github.com/search?q=repo%3ATanStack%2Fform%20no-children-prop&type=code) - i would prefer if the same setting was set for the rule

LeCarbonator:
This is mostly preference and unrelated to tanstack form.

```tsx
// These writing styles are equivalent in React
<Component children={(foo) => <Bar/>} />

<Component>
  {(foo) => <Bar/>}
</Component>
```

I think it's fine to add docs explaining this as I have seen others ask about it too.
However, I think specifically targeting ESLint "erroring" is overkill. It's up to the user to decide if they want to enforce a certain way to write children functions or not.

KwibooReese:
I changed the wording to mention 

> This is mostly preference and unrelated to tanstack form.
> 
> ```tsx
> // These writing styles are equivalent in React
> <Component children={(foo) => <Bar/>} />
> 
> <Component>
>   {(foo) => <Bar/>}
> </Component>
> ```
> 
> I think it's fine to add docs explaining this as I have seen others ask about it too. However, I think specifically targeting ESLint "erroring" is overkill. It's up to the user to decide if they want to enforce a certain way to write children functions or not.

I've reworded the original commit, let me know what you think. It could be good to add what you put in that message into the docs:
```jsx
<Component>
  {(foo) => <Bar/>}
</Component>
```

I don't think it's entirely necessary to these proposed changes, but it would be helpful for newer devs.

PR #1417, opened 2025-04-15
docs(react): use type annotations instead of casting

harry-whorlow:
As discussed in the maintainers channel, updated the react docs to use type annotations instead of casting.


gioboa (approved):
Thanks :clap: @harry-whorlow

PR #1420, opened 2025-04-16
fix(FieldApi): onChangeListenTo should not block onChangeAsync validator on field it is listening to

### LINKED ISSUE #1410: onChangeListenTo will deactivate onChangeAsync of target
### Describe the bug

Having two form fields, field A with onChangeAsync and field B with onChangeListenTo, the onChangeAsync of field A will never be triggered.
Once the target inside onChangeListenTo is renamed or the whole line is commented out, the onChangeAsync will trigger again.

https://stackblitz.com/edit/tanstack-form-h3v7ps9t?file=src%2Findex.tsx

### Steps to reproduce

Open stackblitz reproducible example, type in "error" into the firstname field.
It should trigger an async error after 1 second, but does not.
Comment out the onChangeListenTo in the Last Name form field, then the first name will trigger if "error" is entered.

### Expected behavior

onChangeAsync should still work for all fields.

Every time

### Screenshots or Videos

_No response_

### Platform

Windows 11, Edge Version 135.0.3179.54

### TanStack Form adapter

react-form

### TanStack Form version

v1.3.0

### TypeScript version

v5.7.3

### Additional context

_No response_

Pascalmh:
fixes #1410


LeCarbonator:
I believe this can cause another problem, but I'm not confident I wrote the proper test case for it:

If two fields share a common `onChangeListenTo` of a third field, then it could be that the sync error from one affects the async validator from the other.

lukethacoder:
Running into similar issues with inter-dependant fields.

```ts
new FieldApi({
  name: 'field_1',
  validators: {
    onChangeListenTo: ['field_2', 'field_3'],
    onChangeAsync: async (props) => {
      // magical async stuff
    }
  },
})

new FieldApi({
  name: 'field_2',
  validators: {
    onChangeListenTo: ['field_1'],
    onChangeAsync: async (props) => {
      // magical async stuff
    }
  },
})

new FieldApi({
  name: 'field_2',
  validators: {
    onChangeListenTo: ['field_1'],
    onChangeAsync: async (props) => {
      // magical async stuff
    }
  },
})
```

Shifting from using `onChangeAsync`  to the non-async `onChange` seems to fix the issues... but means we loose `async` which is not ideal. Will have to do for now.

PR #1421, opened 2025-04-16
feat(form-core): add `parseValueWithSchema()` stateless method

LeCarbonator:
As of tanstack form `v1.4`, there are two ways to use standard schemas as validators:

## 1. Pass directly
```ts
validators: {
  onChange: mySchema
}
```
## 2. Create a callback and import `standardSchemaValidators`
```ts
validators: {
  onChange: ({ value }) => {
    return standardSchemaValidators.validate(
      { value, validationSource: 'form' },
      mySchema
    );
  }
}
```

While more verbose, the second function allows you to combine standard schemas easily with other onChange validation a user may want on the same form / field.
There are some things to note about this:
* `value` is redundant, as `formApi | fieldApi` would have access to it
* `validationSource` is redundant, as `formApi | fieldApi` can infer it
* mySchema can be anything, so this **must** be passed
* `standardSchemaValidators` must be imported as a separate 'helper' file.

Given some users may want to add more logic than just a schema to their validation (and based on the discussion in this [discord post](https://discord.com/channels/719702312431386674/1302072288673533972), I propose the following:

### Usage
```ts
validators: {
  onChange: ({ value, formApi }) => {
     const issues = formApi.getValidationIssuesFromSchema(mySchema);
     /*
     readonly StandardSchemaV1Issue[] | {
    form: {
        [k: string]: StandardSchemaV1Issue[];
    };
    fields: {
        [k: string]: StandardSchemaV1Issue[];
    };
} | undefined
     */
     if (issues) return issues;
     
     // other custom validation goes here
  }
}
```
This would allow users to customize their validation more while preserving the parsing that `standardSchemaValidators` does.

# Important
This method should be stateless. Users should have control over whether to return the issues or not.


LeCarbonator:
`parseValuesWithSchema` is probably the closest match for now. It's not a word standard schema uses itself, but it describes the stateless method quite well in my opinion.

packages/form-core/src/FieldApi.ts
```diff
@@ -1643,6 +1643,30 @@ export class FieldApi<
     )
   }
 
+  /**
+   * Parses the field's value with the given schema and returns
+   * issues (if any). This method does NOT set any internal errors.
+   * @param schema The standard schema to parse this field's value with.
+   */
+  parseValueWithSchema(schema: StandardSchemaV1) {
```
Balastrong, line 1651: Let's use the arrow function notation to preserve the scope (and consistency) in the new methods

packages/form-core/tests/FormApi.spec.ts
```diff
@@ -2907,4 +2908,147 @@ describe('form api', () => {
...
+
+    expect(nameSchemaResult?.fields).toHaveProperty('name')
+    expect(Array.isArray(nameSchemaResult?.fields['name'])).toBe(true)
+    expect(nameSchemaResult?.fields['name']).not.toHaveLength(0)
+
+    // First name schema should complain that 'firstName' is too short
+    const firstNameSchemaResult =
+      form.parseFieldValuesWithSchema(firstNameSchema)
+    expect(firstNameSchemaResult).not.toBeUndefined()
+    expect(Array.isArray(firstNameSchemaResult)).not.toBe(true)
```
Balastrong, line 2943: I'm curious, there are quite some `not` in all the new tests, do you see any paritcular advantage than just using `toBeDefined` or `toBe(false)` in some cases?

packages/form-core/src/standardSchemaValidator.ts
Balastrong, line 1: Well done on the types! Do you want to write some type tests to ensure `validate` and `validateAsync` get the right return type from this?

docs/framework/react/guides/validation.md
```diff
@@ -503,6 +503,28 @@ Async validations on form and field level are supported as well:
...
 ```
 
+If you need even more control over your Standard Schema validation, you can combine a Standard Schema with a callback function like so:
+
+```tsx
+<form.Field
+  name="age"
+  asyncDebounceMs={500}
+  validators={{
+    onChange: 1500,
```
Balastrong, line 513: This line can be removed?

Balastrong (commented):
Thanks for the PR, I like how this is turning out!

LeCarbonator:
That should cover the comments. Also added some type safety to not pass schemas that will set errors on non-existent fields.

Balastrong (approved):
That's a perfect PR and a really nice addition, thank you so much!

PR #1422, opened 2025-04-16
feat(form-core): add `<FieldMeta>.isValid`

LeCarbonator:
This meta property is a shortened alternative to checking
`field.state.meta.errors.length === 0`. Since FormApi already has this
property, a FieldApi variant wouldn't hurt to be more clear.

This PR is based on [Balastrong's suggestion on Discord](https://discord.com/channels/719702312431386674/1100437019857014895/1362130684746928299).


LeCarbonator:
For some reason, refactoring this FormApi check fails horribly. Does someone know why?

```diff
- const isFieldsValid = !fieldMetaValues.some(
-           (field) =>
-             field?.errorMap &&
-             isNonEmptyArray(Object.values(field.errorMap).filter(Boolean)),
-         )

+ const isFieldsValid = fieldMetaValues.every(
+          (field) => field === undefined || field.isValid,
+        )
```

LeCarbonator:
No, but it's clear that the meta needs to be adjusted in some places. I'll tinker with it tomorrow.

LeCarbonator:
It's wild. I added checks wherever field errors or errorMap were tested for and it worked no problem. However, trying to rewrite `isFieldsValid` to use this boolean fails at 8 locations. I'll have to spend some more time figuring out how store works or something related to that

LeCarbonator:
> Does this work?
> 
> ```ts
> const isFieldsValid = fieldMetaValues
>   .filter(Boolean)
>   .every((field) => field.isValid)
> ```

It works now! This code wasn't the problem, but my implementation of `isValid`.

packages/form-core/src/FormApi.ts
```diff
@@ -876,11 +879,9 @@ export class FormApi<
...
         )
 
-        const isFieldsValid = !fieldMetaValues.some(
-          (field) =>
-            field?.errorMap &&
-            isNonEmptyArray(Object.values(field.errorMap).filter(Boolean)),
-        )
+        const isFieldsValid = fieldMetaValues
+          .filter(Boolean)
+          .every((field) => field!.isValid)
```
Balastrong, line 884: I believe I found what's wrong here that forced us to handle isValid manually each time we touched field meta. That's definitely a smell as we shouldn't manually update a derived field (`isValid`) so we had to find a different way.

Turns out a few lines here above, this has a wrong cast:
```ts
const fieldMetaValues = Object.values(currBaseStore.fieldMetaBase) as (
  | AnyFieldMeta
  | undefined
)[]
```
That should be `AnyFieldMetaBase`, which does not have derived field properties, hence isValid will always be... `undefined`

Here's my suggestion, let's replace it with
```ts
const currFieldMeta = currDepVals[1]

// Computed state
const fieldMetaValues = Object.values(currFieldMeta).filter(
  Boolean,
) as AnyFieldMeta[]
```

Now that is the real AnyFieldMeta array and since we're filtering first we can also get rid of all the nullish checks below.

Aaand we can remove all the `const fieldIsValid` we just added around :)

LeCarbonator: LGTM! However, now we have a bunch of 'useless' refactors that extracted the `errorMap` outside of the returned object. It was needed to determine `fieldIsValid`, but since that's gone, should I remove all those diffs back to the original?

Balastrong: Ah yes, sorry, I completely missed that. We can remove those changes too

LeCarbonator: There we go, I removed the unnecessary extraction of `errorMap`. Did you notice anything else that needs addressing?

packages/form-core/src/FieldApi.ts
```diff
@@ -652,6 +652,10 @@ export type FieldMetaDerived<
    * A flag that is `true` if the field's value has not been modified by the user. Opposite of `isDirty`.
    */
   isPristine: boolean
+  /**
+   * A boolean indicating if the field is valid.
```
Balastrong, line 656: It might be obvious but I'd probably also add a sentence to say that is valid means not having errors.

packages/form-core/src/FormApi.ts
Balastrong: Thanks! No I think that's it :)

packages/form-core/src/FieldApi.ts
LeCarbonator: Sure! We can also adjust the form variant of it, so let me know what you think of this sentence and I'll add it to both:

"A boolean indicating if the field is valid. `True` if there are no field errors."

LeCarbonator: Sure! It also ensures that the capitalization is the way JavaScript does it.
Avert your eyes though, there's a bunch of capital Ts in other meta property descriptions as well.

I'll amend the jsdoc in a moment

Balastrong (approved):
That's great, thank you!

PR #1426, opened 2025-04-17
feat(form-core): add array method field.filterValues and form.filterFieldValues

Pascalmh:
### Problem

When setting a field back to an empty array 

```ts
const form = new FormApi({
  defaultValues: {
    people: [
        { firstName: "Bob" },
        { firstName: "Marry" },
    ],
  },
});

form.setFieldValue('people', [
  { firstName: "Marry" }, // <-- "Bob" should be removed
]);
```

`items[0].firstName` would still exist - when there's a validation on the firstName (on the Field-level) `form.canSubmit` would be false.

### Solution

adding `field.filterValues` and `form.filterFieldValues` to TSF:

```ts
form.filterFieldValues(
  'people',
  ({firstName}) => firstName === "Marry"
)
```


LeCarbonator:
I think if/when this gets merged, it would be great to rework the Arrays section of the docs to write about swap, insert, push, filter and clear. Thoughts? Should it be somewhere else?

LeCarbonator:
Oof, it did not like that refactor of metaHelper. I'll get to it later today.

Issue #1183, opened 2025-02-25
onSubmit meta docs


crutchcorn:
Document the behavior of https://github.com/TanStack/form/pull/1167 in the React guides

> Imagine a usecase where the user wants two buttons, they both send the item to the backend (after form validation) but one goes back to the list and the other goes to a different page, like a step 2 or a detail page.

```tsx
const form = useForm({
    defaultValues: {
      firstName: '',
      lastName: '',
    },
    // Is 'never' by default unless you provide a type
    // {} is the default value passed to `onSubmit`'s `meta` property
    onSubmitMeta: {} as { someValue: string },
    onSubmit: async ({ value, meta }) => {
      // Do something with form data
      console.log(value)
      // Do something with the values passed via handleSubmit
      console.log(meta)
    },
  })

  return (
    <div>
      <h1>Simple Form Example</h1>
      <form
        onSubmit={(e) => {
          e.preventDefault()
          e.stopPropagation()
          form.handleSubmit({
            someValue: 'value',
          })
        }}
      >
```

MartinCura:
This feature is great for having 2 submit buttons that do slightly different behavior on submit, such as "Save and continue" & "Save and exit", right? Or is there something better for that use case?

crutchcorn:
100% @MartinCura

PR #1428, opened 2025-04-17
docs: Expand submission handling and validation documentation 

### LINKED ISSUE #1430: Validation for nested fields only works with dot-notation keys (e.g. 'person.firstName') - unclear in docs
### Describe the bug

Hi! 👋

I ran into a confusing behavior when trying to validate nested fields using `validators.onChange`.

**Here's what I tried (which didn’t work):**
```jsx
validators: {
  onChange: ({ value }) => ({
    fields: {
      person: { firstName: 'Wrong!' }, // ❌ Doesn't apply error
    },
  }),
}
```

This didn't apply any validation errors. After quite a bit of debugging, I found that the following structure works:
```jsx
validators: {
  onChange: ({ value }) => ({
    fields: {
      'person.firstName': 'Wrong!', // ✅ Works!
    },
  }),
}
```

**✅ Working example (with dot notation):**
https://stackblitz.com/edit/form-validation-using-dot-notation-for-nested-objects?file=src%2Findex.tsx

**❌ Not working example (with object nesting):**
https://stackblitz.com/edit/tanstack-form-validation-using-nested-object-paths?file=src%2Findex.tsx

**Suggestion**
It might be worth clarifying in the documentation that the fields object returned from validators must use dot-notation string keys even for nested object paths.

If this is the intended design, a brief mention would go a long way in preventing confusion. It took me about an hour of debugging to realize that nested object structure wasn't supported in this context.

Thanks a lot for the great work on TanStack Form - really love how powerful and type-safe it is! 🙌

https://stackblitz.com/edit/tanstack-form-validation-using-nested-object-paths?file=src%2Findex.tsx

### Steps to reproduce

1. Open the [non-working example](https://stackblitz.com/edit/tanstack-form-validation-using-nested-object-paths?file=src%2Findex.tsx).
2. Click the "First Name" input.
3. Type any character (e.g., `a`).

### Expected behavior

The text "Wrong!" should appear next to the input as a validation error message.

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/0cdb7597-5743-41ba-bbcd-15a961653a8b

### Platform

- OS: macOS 15.4
- Browser: Chrome
- Browser version: 135.0.7049.96 (Official Build) (arm64)

### TanStack Form adapter

react-form

### TanStack Form version

v1.6.1

### TypeScript version

@types/react-dom@19.1.2 / @types/react@19.1.2

### Additional context

_No response_

LeCarbonator:
This PR aims to expand the submission handling page with the following:

- [x] Rewrite Submission meta to be clearer
- [x] Add nested fields in form -> field validation example
- [x] Add Standard Schema input vs. output scenario
- [x] Address parity problems between frameworks in `validation.md` and `submission-handling.md`
  - [x] Vue
  - [x] Solid
  - [x] React
  - [x] Angular
  
  ~~[ ] Lit~~ There are docs files that have higher priority than these ones for Lit, as the current documentation is lacking.

Due to point 3, this is marked as draft. Let me know if there are improvements to be added into the uploaded react files.
Since I am only familiar with React, I will test run the code snippets of other frameworks, so they might take a bit of a while.

This PR closes #1430 when merged.


LeCarbonator:
Angular is quite something. I'm attempting to work through it, but if someone with more experience in it could port the two files I'd appreciate it

docs/framework/angular/guides/submission-handling.md
```diff
@@ -0,0 +1,113 @@
...
+        <form (submit)="handleSubmit($event, {
+            section: 'paymentInfo',
+          })">
+          <!-- -->
+        <button type="submit">Submit Payment Info</button>
+        </form>
+  </div>
+  `,
+})
+
```
Pascalmh, line 47: The decorator should touch the class directly, no empty line in between
```suggestion
```

docs/framework/angular/guides/submission-handling.md
```diff
@@ -0,0 +1,113 @@
...
+// Tanstack Form uses the input type of Standard Schemas
+const defaultValues: z.input<typeof schema> = {
+  age: '13',
+}
+
+// ...
+
+export class AppComponent {
```
Pascalmh, line 98: Is this missing the `@Component` decorator?

LeCarbonator: Not on its own, but it has been trimmed (as shown with `// ...`) since this is unrelated to any DOM elements and is only about the zod schema and `handleSubmit`

LeCarbonator: I could rewrite it to this if you think it's a problem
```suggestion
@Component({
  // ...
})
export class AppComponent {
```

docs/framework/react/guides/submission-handling.md
```diff
@@ -3,38 +3,104 @@ id: submission-handling
 title: Submission handling
 ---
 
-In a situation where you want to have multiple form submission types, for example, a form that has a button that navigates to a sub-form and another button that handles a standard submission, you can make use of the onSubmitMeta prop and the handleSubmit function overloads.
+## Passing additional data to submission handling
 
-## Basic Usage
+You may want to split your long form into multiple sections, for example, when creating a booking form.
+To receive meta data in your `onSubmit` function, you can specify the `onSubmitMeta` property.
```
harry-whorlow, line 9: My concern here is that the wording would lead the reader to believe that this is a from group, which although the functionality is close, is not it's intended purpose. 

I think the example given [here](https://github.com/TanStack/form/issues/1183) is best... differiecieate between save and continue, save and exit (I originally needed it for, save and exit, save and navigate elsewhere).

What do you think?

LeCarbonator: Sure, it would help differentiate it from form groups.

However, I thought this was a short / simple way to create different meta by having it be two forms. Could you elaborate on how that action meta would be passed in your example? In a normal button? or also using a `type="submit"` button?

harry-whorlow: Well I suppose you would have two buttons, one could be type submit, the other be type button where you pass the relevant meta to the onSubmission form handler and the other to the onClick of the button (if you care about the semantics... otherwise just two buttons would do). 

_super quick example not sure it works...🙃_
```tsx
function MyForm() {
  // I've committed heresy by using react router
  const navigate = useNavigation()

  const form = useForm({
    defaultValues: defaultPeople,
    onSubmitMeta: {} as () => void,
    onSubmit({ value, meta }) {
      console.log(value)

      meta()
    },
  })

  return (
    <form onSubmit={() => form.handleSubmit(navigate('/'))}>
      <button type="submit"></button>
      <button type="button" onClick={() => navigate('/hello')}></button>
    </form>
  )
}
```

Pascalmh: I guess I'd implement it using the primary action as type "submit" and the secondary as "button"

But having both as a type "button" could be better for an example as it's more explicit.

---

It definitely clears up why this feature exists in the first place using the publish/draft example.

LeCarbonator: > I guess I'd implement it using the primary action as type "submit" and the secondary as "button"
> 
> But having both as a type "button" could be better for an example as it's more explicit.
> 
> It definitely clears up why this feature exists in the first place using the publish/draft example.

How about it being `<button type="submit" onClick={() => form.handleSubmit()}/>` and the form simply prevents default and stops propagation?

Pascalmh: Everything that takes the focus away from the "meta" part should be left out if possible.

If it's easier to go with submit+button, let's go with that, otherwise both type=button

harry-whorlow: @Pascalmh I'm 100% onboard with this opinion. I think "publish" "draft" is a better way of wording... 

We could just do two type "button" so you can have have the logic next to each other. Sure its not 100% semantically correct, but I think having the different logic next to each other is nicer for illustration purposes. 🙂

LeCarbonator: > @Pascalmh I'm 100% onboard with this opinion. I think "publish" "draft" is a better way of wording it too...
> 
> We could just do two type "button" so you can have have the logic next to each other. Sure its not 100% semantically correct, but I think having the different logic next to each other is nicer for illustration purposes. 🙂

Would like to hear your opinions on this example. If you like the approach, we can still change the details such as the meta itself

cc @Pascalmh

docs/framework/react/guides/submission-handling.md
```diff
@@ -3,38 +3,91 @@ id: submission-handling
...
+
+type FormMeta = {
+  submitAction: 'continue' | 'backToMenu' | null
+}
+
+// Metadata is not required to call form.handleSubmit().
+// Specify what values to use as default if no meta is passed
+const defaultMeta: FormMeta = {
+  submitAction: null,
+}
```
harry-whorlow, line 24: This is something I entirely forgot about, `what if you don't pass any meta`... I like this solution. Also the avoidance to the casting like, I did in the past is a good catch.😊

docs/framework/react/guides/submission-handling.md
harry-whorlow: I like the change, for me its a big step up. 🚀

harry-whorlow:
I'm not too confident reviewing angular, since I've only had a brief experience with it. I'll message some of the other maintainers to see if they're able to have a look. But, great job on adding to the docs!

docs/framework/angular/guides/submission-handling.md
```diff
@@ -0,0 +1,107 @@
...
+        handleClick({ submitAction: 'continue' })
+      ">Submit and continue</button>
+      <button type="submit" (click)="
+        handleClick({ submitAction: 'backToMenu' })
+      ">Submit and back to menu</button>
+    </form>
+  `,
+})
+export class AppComponent {
+  name = 'Angular';
```
Pascalmh, line 43:
```suggestion
```

this seems unused?

docs/framework/angular/guides/submission-handling.md
```diff
@@ -0,0 +1,107 @@
...
+  age: '13',
+}
+
+// ...
+
+@Component({
+  // ...
+})
+export class AppComponent {
+  name = 'Angular'
```
Pascalmh, line 93:
```suggestion
```

docs/framework/angular/guides/validation.md
```diff
@@ -288,6 +288,134 @@ export class AppComponent {
...
 ```
 
+### Setting field-level errors from the form's validators
+
+You can set errors on the fields from the form's validators. One common use case for this is validating all the fields on submit by calling a single API endpoint in the form's `onSubmitAsync` validator.
+
+```angular-ts
+@Component({
+  selector: 'app-root',
+  standalone: true,
```
Pascalmh, line 298:
```suggestion
```

docs/framework/angular/guides/validation.md
```diff
@@ -288,6 +288,134 @@ export class AppComponent {
...
+          }
+        </ng-container>
+      </div>
+      <button type="submit">Submit</button>
+    </form>
+  `,
+})
+
+export class AppComponent {
+  name = 'Angular';
```
Pascalmh, line 327:
```suggestion
```

docs/framework/angular/guides/validation.md
```diff
@@ -288,6 +288,134 @@ export class AppComponent {
...
+>   />
+>    @if (ageField.api.state.meta.errors.length > 0) {
+>       <em role="alert">{{ ageField.api.state.meta.errors.join(', ') }}</em>
+>     }
+>   </ng-container>
+> </div>
+> `,
+> })
+> export class AppComponent {
+>   name = 'Angular';
```
Pascalmh, line 395:
```suggestion
```

Pascalmh (changes requested):

Pascalmh:
@harry-whorlow @LeCarbonator I used to work with angular until about a year ago - it looks good and seems to follow the Angular Style Guide - but angular evolves quickly - so someone that is actively working with Angular should take another look if possible.

docs/framework/angular/guides/submission-handling.md
```diff
@@ -0,0 +1,107 @@
...
+
+// Metadata is not required to call form.handleSubmit().
+// Specify what values to use as default if no meta is passed
+const defaultMeta: FormMeta = {
+  submitAction: null,
+};
+
+@Component({
+  selector: 'app-root',
+  standalone: true,
```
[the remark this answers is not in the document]
LeCarbonator: Understood.
```suggestion
```

Balastrong (approved):
Very well done :D

harry-whorlow:
!hanks for the submission @LeCarbonator, and @Pascalmh 
LGTM🚀

PR #1432, opened 2025-04-20
Merging Form-level server errors with Field-level errors

theVedanta:
Instead of the previous `serverValidate` function just returning a string, we now return an object:
```ts
const serverValidate = createServerValidate({
  ...formOpts,
  onServerValidate: ({ value }) => {
    if (value.age < 12)
      return { age: 'Server validation: You must be at least 12 to sign up' }
  },
})
```

This then maps to the field-level errors with the `setMeta` options.

The only reason this is called a preliminary approach is because we are setting the `onServer` key of the errorMap differently.

Previously, the error map for the form looked like so:
```ts
{
  onServer: "You need to be 12 years or older to sign up"
}
```

Now however, the onServer map is slightly different:
```ts
{
  onServer: { age: "You need to be 12 years or older to sign up" }
}
```

This causes some degree of inconsistency, which is why I would love some feedback on whether we can use some other property (or maybe create our own!) to manage the errors between fields and forms. This would not be a comprehensive property but just a map to sync them back and forth.

NextJS support has been pushed. Remix and Start coming soon.
Support for Array field types will also be done shortly.

Would love your feedback and comments on this approach before I proceed!

### DIFF
--- a/packages/react-form/src/nextjs/createServerValidate.ts
+++ b/packages/react-form/src/nextjs/createServerValidate.ts
@@ -104,12 +104,25 @@ export const createServerValidate =
         : onServerError
     ) as UnwrapFormAsyncValidateOrFn<TOnServer>
 
+    // Extract string values from errors if they're in object format
+    const errorsArray = onServerErrorVal
+      ? Array.isArray(onServerErrorVal)
+        ? onServerErrorVal.map((err) =>
+            typeof err === 'object' ? Object.values(err)[0] : err,
+          )
+        : [
+            typeof onServerErrorVal === 'object'
+              ? Object.values(onServerErrorVal)[0]
+              : onServerErrorVal,
+          ]
+      : []
+
     const formState: ServerFormState<TFormData, TOnServer> = {
       errorMap: {
         onServer: onServerError,
       },
       values,
-      errors: onServerErrorVal ? [onServerErrorVal] : [],
+      errors: errorsArray,
     }
 
     throw new ServerValidateError({
--- a/examples/react/next-server-actions/src/app/action.ts
+++ b/examples/react/next-server-actions/src/app/action.ts
@@ -9,9 +9,8 @@ import { formOpts } from './shared-code'
 const serverValidate = createServerValidate({
   ...formOpts,
   onServerValidate: ({ value }) => {
-    if (value.age < 12) {
-      return 'Server validation: You must be at least 12 to sign up'
-    }
+    if (value.age < 12)
+      return { age: 'Server validation: You must be at least 12 to sign up' }
   },
 })
 
--- a/examples/react/next-server-actions/src/app/client-component.tsx
+++ b/examples/react/next-server-actions/src/app/client-component.tsx
@@ -3,7 +3,6 @@
 import { useActionState } from 'react'
 import { mergeForm, useForm, useTransform } from '@tanstack/react-form'
 import { initialFormState } from '@tanstack/react-form/nextjs'
-import { useStore } from '@tanstack/react-store'
 import someAction from './action'
 import { formOpts } from './shared-code'
 
@@ -18,14 +17,8 @@ export const ClientComp = () => {
     ),
   })
 
-  const formErrors = useStore(form.store, (formState) => formState.errors)
-
   return (
     <form action={action as never} onSubmit={() => form.handleSubmit()}>
-      {formErrors.map((error) => (
-        <p key={error as unknown as string}>{error}</p>
-      ))}
-
       <form.Field
         name="age"
         validators={{
--- a/examples/react/next-server-actions/src/app/shared-code.ts
+++ b/examples/react/next-server-actions/src/app/shared-code.ts
@@ -2,7 +2,6 @@ import { formOptions } from '@tanstack/react-form/nextjs'
 
 export const formOpts = formOptions({
   defaultValues: {
-    firstName: '',
     age: 0,
   },
 })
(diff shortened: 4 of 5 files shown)


vincehi:
Hello, do you need a hand to move this branch forward? I would love to see it happen :)

theVedanta:
> Hello, do you need a hand to move this branch forward? I would love to see it happen :)

Yeah sure dude, go for it!

theVedanta:
Not sure if this feature is still requested or whether it has been fixed. Don't wanna tag maintainers but this has been dormant for a while, I'd like some insight on how to proceed! @LeCarbonator @harry-whorlow

packages/react-form/src/nextjs/createServerValidate.ts
```diff
@@ -96,20 +96,41 @@ export const createServerValidate =
       validationSource: 'form',
     })) as UnwrapFormAsyncValidateOrFn<TOnServer> | undefined
 
+    console.log('ON SERVER ERROR', onServerError)
```
LeCarbonator: Forgotten `console.log`

packages/react-form/src/nextjs/createServerValidate.ts
```diff
@@ -96,20 +96,41 @@ export const createServerValidate =
...
+      ? Array.isArray(onServerErrorVal)
+        ? onServerErrorVal.map((err) =>
+            typeof err === 'object' ? Object.values(err)[0] : err,
+          )
+        : [
+            typeof onServerErrorVal === 'object'
+              ? Object.values(onServerErrorVal)[0]
+              : onServerErrorVal,
+          ]
+      : []
```
LeCarbonator: This double ternary is a bit hard to follow. Please refactor to `if` statements instead.

packages/react-form/src/nextjs/createServerValidate.ts
```diff
@@ -96,20 +96,41 @@ export const createServerValidate =
...
+        onServerError.fields as UnwrapFormAsyncValidateOrFn<TOnServer>
+    } else {
+      onServerErrorVal = onServerError as UnwrapFormAsyncValidateOrFn<TOnServer>
+    }
+
+    // Extract string values from errors if they're in object format
+    const errorsArray = onServerErrorVal
+      ? Array.isArray(onServerErrorVal)
+        ? onServerErrorVal.map((err) =>
+            typeof err === 'object' ? Object.values(err)[0] : err,
```
LeCarbonator: I'll need to look into this one. Objects don't inherently have an order, so I'm not sure how consistent taking the first element would be. Perhaps we do this in other places?

The object format you're referring to is the server error fields coming in, right?

LeCarbonator:
The merge conflicts appear to be in some of the important diffs, so I'll leave that up to you. I'll see if convenient unit tests could be made for these as we don't want to regress on this.

packages/react-form/src/nextjs/createServerValidate.ts
jasikpark: There is actually an order of sorts nowadays: https://2ality.com/2015/10/property-traversal-order-es6.html

1. **Integer indices** (array-like indices) come first, in ascending numeric order
2. **String keys** (non-integer strings) come next, in creation order
3. **Symbol keys** come last, in creation

theVedanta:
> Sorry for the long delay! I'm not too confident in my SSR knowledge, so I'll make sure to give this a try myself. I'll also go through some issues and the reproductions to confirm things.
> 
> Hopefully, this initial review should help with some refactoring. It's looking promising!
> 
> Since this has been stale for way too long, this will be my focus for now. Feel free to tag me if there's more info you need or a second review. I can also help out with the PR if you'd like.

Your help with the PR would be awesome! I'm pretty new to learning about the tanstack ecosystem.
I'll try to make another commit with the fixes.

packages/react-form/src/nextjs/createServerValidate.ts
theVedanta: > I'll need to look into this one. Objects don't inherently have an order, so I'm not sure how consistent taking the first element would be. Perhaps we do this in other places?
> 
> The object format you're referring to is the server error fields coming in, right?

I've checked and recalled why the Object.values()[0] was in there. It was a part of the "Preliminary approach" of the draft PR, which is no longer relevant. Please check the latest commit. This now contains a simple parse for objects / strings.

The server can now return 2 kinds of things:
```ts
// Type 1 (an object)
return {
    fields:
        { age: "Field Error: Should be 12 to sign up" },
    form: "" // Can be blank or omitted, but populates the form error if filled
}

// Type 2
return "Form Error: Should be 12 to sign up"
```

Let me know if you think anything needs further fixing.

theVedanta:
@LeCarbonator just following up on this :)

crutchcorn:
@theVedanta I'm super sorry about this, but I've realized that our SSR story is DEEPLY broken (my fault, not related to this specifically).

As such, I'm prototyping new ideas here:

https://github.com/TanStack/form/pull/1890

While I don't want the general public to worry about any potential breaking changes yet (we'll have extensive migration guides and such), I'd love you thoughts on the approach I took there to fix things.

LMK!

Thanks a billion for your amazing help. Closing this PR to move discussions there

theVedanta:
> @theVedanta I'm super sorry about this, but I've realized that our SSR story is DEEPLY broken (my fault, not related to this specifically).
> 
> As such, I'm prototyping new ideas here:
> 
> #1890
> 
> While I don't want the general public to worry about any potential breaking changes yet (we'll have extensive migration guides and such), I'd love you thoughts on the approach I took there to fix things.
> 
> LMK!
> 
> Thanks a billion for your amazing help. Closing this PR to move discussions there

haha no worries man. and thanks so much for closing this. I ended up discussing this PR with Luca but totally forgot to close it! I would love to check out the awesome work you're doing to improve the SSR \:)

PR #1440, opened 2025-04-21
fix(form-core, react-form): remove overwritting of falsy values upon array element shifting

### LINKED ISSUE #1439: Falsy values are being turned into undefined upon removing array field
### Describe the bug

When removing array elements, currently subfields with falsy values on the last row are being changed to undefined. Added a condition to specifically check for undefined. 

This is critical as the casting converting booleans or empty strings to undefined leads to uncontrolled inputs in React and overall strange behavior regarding validators that expect particular value types. 

https://stackblitz.com/edit/vitejs-vite-6huqeq8i?file=src%2FApp.tsx,src%2Fcomponents%2FForm.tsx

### Steps to reproduce

1. Click the push value button twice so you have three rows. 
2. Click the Log button and inspect the console.
3. Notice that interests logs:
```ts
[
    {
        "id": "01965a86-4995-768a-85b9-be02512c2208",
        "interestName": ""
    },
    {
        "id": "01965a86-4ab7-751b-9368-8ce7539f2e49"
        "interestName": undefined
    }
]
```
interestName is now missing! It was overwritten as undefined

### Expected behavior

Falsy values remain stable after removing field value. 

None

### Screenshots or Videos

_No response_

### Platform

Replicatable on any code sandbox/React environment 

### TanStack Form adapter

react-form

### TanStack Form version

1.6.3

### TypeScript version

5.7.2

### Additional context

_No response_

juanvilladev:
This PR addresses two issues:

#1363: Temporary undefined field.state.value during React re-renders

#1439: Falsy values being overwritten when field/form stores go out of sync

1. **Fix for #1439**

When array operations like remove, shift, or swap are performed, the form and field stores can temporarily go out of sync. Previously, we were overwriting all falsy values (e.g., 0, false, '') during this sync process.

To fix this, I added a ternary check to preserve the original value when it's falsy but not undefined. This prevents unintended overwrites of valid falsy inputs.

2. **Fix for #1363**

React was re-rendering while the field store was briefly out of sync with the form store, leading to field.state.value being undefined. During this transitional state, we now return null from the field component, which avoids runtime errors and prevents rendering invalid state.

Additionally, we were encountering an error reading errorMap from undefined. This was due to how we retrieved field names within form-level validators.

The fix includes:

Switching to baseStore instead of the derived store when accessing the form state (from validators). This is because baseStore is the true source of truth and reflects updates immediately, even during field deletion. This ensures we always have an up-to-date list of active fields.

Let me know if anything needs more clarification or if you'd like to walk through any part of the solution!


juanvilladev:
This bug is mainly replicatable on React, specifically because of this effect which calls the update function post array shifting.

https://github.com/TanStack/form/blob/3b90f86acd8a7cc6e1bed187e60d5ca2a4814fcc/packages/react-form/src/useField.tsx#L195-L201

For this reason, I wrote this test specifically within the react-form package. This clearly duplicates the issue within the stackblitz on the issue and ensures that falsy values are preserved.

juanvilladev:
The more I think of this... There's something specific wrong with React. I think this has to do with issue #1363 
It's strange that we ever even enter the block where this occurs given this.state.value should be undefined.

Seems like this.state.value !== the previous value inside setValue.

juanvilladev:
Completed all updates, I believe this is ready for review.

- Confirmed we no longer get the issue with uncontrolled components
- Confirmed we no longer get the error regarding reading errorMap from undefined from validators
- Confirmed falsy values are no longer being set to undefined upon deleting array field elements

Balastrong (commented):
The undefined check looks good, but let me challenge on the other fix: are you sure it's not a workaround to something that needs to be fixed on Store instead?

Also thank you for all the examples and details, great work as usual :)

juanvilladev:
Won't be able to look into this for a while

PR #1448, opened 2025-04-24
docs(svelte-form): add docs for svelte-form

dummdidumm:
without reference docs for now because they can't be generated, as the generator doesn't know how to deal with Svelte imports


Balastrong (approved):
Looks great, thank you!

PR #1453, opened 2025-04-25
[Solid.js] feat: add withForm and createAppForm APIs

crutchcorn:
# TODOs

- [ ] Write docs
- [ ] Receive code review from community to validate if `useFieldContext` should return a signal or stay as-is

### DIFF
--- a/packages/solid-form/src/makeFormCreate.tsx
+++ b/packages/solid-form/src/makeFormCreate.tsx
@@ -1,0 +1,362 @@
+import { createContext, useContext } from 'solid-js'
+import { createForm } from './createForm'
+import type { Component, Context, JSX, ParentProps } from 'solid-js'
+import type {
+  AnyFieldApi,
+  AnyFormApi,
+  FieldApi,
+  FormAsyncValidateOrFn,
+  FormOptions,
+  FormValidateOrFn,
+} from '@tanstack/form-core'
+import type { FieldComponent } from './createField'
+import type { SolidFormExtendedApi } from './createForm'
+
+/**
+ * TypeScript inferencing is weird.
+ *
+ * If you have:
+ *
+ * @example
+ *
+ * interface Args<T> {
+ *     arg?: T
+ * }
+ *
+ * function test<T>(arg?: Partial<Args<T>>): T {
+ *     return 0 as any;
+ * }
+ *
+ * const a = test({});
+ *
+ * Then `T` will default to `unknown`.
+ *
+ * However, if we change `test` to be:
+ *
+ * @example
+ *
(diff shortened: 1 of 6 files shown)


packages/solid-form/src/makeFormCreate.tsx
```diff
@@ -0,0 +1,362 @@
...
+    const AppForm = ((props) => {
+      return (
+        <formContext.Provider value={form}>
+          {props.children}
+        </formContext.Provider>
+      )
+    }) as Component<ParentProps>
+
+    const AppField = ((props) => {
+      const { children, ...rest } = props
```
aadito123, line 278: https://docs.solidjs.com/reference/reactive-utilities/split-props
Use this instead. children is a function so it would be an edge case for someone to do `reactive() ? () => Comp1 : () => Comp2` but still.

packages/solid-form/src/makeFormCreate.tsx
```diff
@@ -0,0 +1,362 @@
...
+    TOnChange extends undefined | FormValidateOrFn<TFormData>,
+    TOnChangeAsync extends undefined | FormAsyncValidateOrFn<TFormData>,
+    TOnBlur extends undefined | FormValidateOrFn<TFormData>,
+    TOnBlurAsync extends undefined | FormAsyncValidateOrFn<TFormData>,
+    TOnSubmit extends undefined | FormValidateOrFn<TFormData>,
+    TOnSubmitAsync extends undefined | FormAsyncValidateOrFn<TFormData>,
+    TOnServer extends undefined | FormAsyncValidateOrFn<TFormData>,
+    TSubmitMeta,
+    TRenderProps extends Record<string, unknown> = {},
+  >({
```
aadito123, line 323: would avoid destructure here in case the user wants to make them reactive too. The deeper reactivity within props and render are maintained so idk if we want to even enable that pattern.

packages/solid-form/src/makeFormCreate.tsx
```diff
@@ -0,0 +1,362 @@
...
+    >,
+  ) => JSX.Element
+}
+
+export function makeFormCreate<
+  const TComponents extends Record<string, Component<any>>,
+  const TFormComponents extends Record<string, Component<any>>,
+>({
+  fieldComponents,
+  fieldContext,
```
aadito123, line 225: Same with the destructures here, I would avoid them

MAST1999:
I've done some work on this and got tests passing and addressed the comments in #1597 

Can we consolidate the efforts and get this going?
I'm willing to help or take over.

crutchcorn:
Closing in favor of your work in #1597 @MAST1999 - thanks for taking this on/over! :D

PR #1455, opened 2025-04-25
[WIP] Add useAppForm API to Vue

crutchcorn:
This PR adds the ability to use [Form Composition](https://tanstack.com/form/latest/docs/framework/react/guides/validation) APIs in our Vue adapter.

> This PR is blocked on Vue 3.6's release and cannot be released until that version hits Vue's stable channel.

# TODOs:

- [x] Add docs


daniel100097:
Hi, I was just looking for a way to split a form in Vue (using TanStack Form) into separate fieldset components (using useAppForm).

Have you by any chance already found a way to pass the form context to a fieldset component in Vue? Or maybe a way to work with just a subset of the main form?

Specifically, I was trying to create a reusable fieldset for an address, so I could easily reuse it across different forms.

crutchcorn:
Unfortunately we're blocked by this ATM on Vue's side:

https://github.com/vuejs/core/issues/8553

daniel100097:
Even though it's not ideal with Vue, it might make sense to use the workaround for now — and once the Vue issue is resolved, we wouldn't need the workaround anymore.

I just tested it again with your StackBlitz example, and as soon as I write `<component :is="compMap.one"/>` , it seems to work.



```
    _createVNode($setup["SlotProvider"], null, {
      default: _withCtx(({ compMap }) => [
        (_openBlock(), _createBlock(_resolveDynamicComponent(compMap.one))),
        _cache[0] || (_cache[0] = _createElementVNode("p", null, "Despite it being a valid JS object and such:", -1 /* CACHED */)),
        _createElementVNode("p", null, _toDisplayString(JSON.stringify(compMap, null, 2)), 1 /* TEXT */)
      ]),
      _: 1 /* STABLE */,
      __: [0]
    })

crutchcorn:
No dice; Since we're in v1, introducing an API on our end requires docs, migration patterns, and more. I've considered allowing for `<component :is>` but don't want to have folks have to recant old code per a suggestion once Vue is updated.

The workaround for now - if you'd like to do that yourself until we launch - is to copy+paste the implementation of `useAppForm` and such into your codebase - shouldn't require any additional changes upstream.

mkeyy0:
@crutchcorn 
First of all, thank you, TanStack team, for the great library; it really simplifies form building.

However, Vue is really missing the form composition feature. I know https://github.com/vuejs/core/issues/8553 blocks this, but I agree with @daniel100097 here - I think it makes sense to release it with the current limitation that requires using `<component :is />` syntax. 

IMO, a limitation on the Vue side shouldn't block this feature, as this workaround is a reality of working with the Vue framework. Once the issue is resolved (or not, cause in vue you have to do the same trick for other situations, like when you use a component in computed, etc), existing code using `<component :is />` will continue to work anyway, and users will gain the option to use hook components without it. The TanStack implementation itself won't need to change; it's just the framework-specific usage that will change.

P.S. I found that even Vue Router uses `<component :is />` to create a transition, it exposes the `Component` from the slot, which you have to render with `<component :is />`. See: https://router.vuejs.org/guide/advanced/transitions#Transitions. Or do you have other concerns that stop you from doing that?

P.P.S. Also, I found that `<component :is />` doesn't deal well with TypeScript, so it is also an issue with this approach, so maybe we should push the vue team to ship this faster

crutchcorn:
Great news here y'all:

https://github.com/vuejs/core/issues/8553

Has finally been merged! :D 

This means that Vue 3.6 will support our needs here and unblock the PR from moving forward. I'll be prepping this PR for merge leading up to that point

mkeyy0:
Omg! That's such a great news! Thank you for your amazing work!

mkeyy0:
@crutchcorn Do you know roughly when this will land in main? Would be quite useful

crutchcorn:
@mkeyy0 unfortunately, this is blocked by Vue upstream until they release 3.6 as stable. They have the feature landed in their betas, so it's just a matter of a stable release.

PR #1456, opened 2025-04-25
feat(form-core): field meta isDefaultValue

### LINKED ISSUE #1080: isDirty should flip back to false when default values are equal
### Describe the bug

`isDirty` flips to `true` when you change any form value, but doesn't return to `false` when the state returns back to default values.

For most if not all form libraries I've used, the expected behavior is the form is not dirty when the default values is equal. See formik example here:
https://stackblitz.com/edit/react-hooks-form-validation-example-formik-hh9rdieb

Tanstack Form already exposes the `isTouched` property which can be used to determine if the form has been touched. It appears right now that both `isTouched` and `isDirty` now have the same behavior. Except when you reset the form, dirty returns to `false` as expected.

https://stackblitz.com/edit/tanstack-form-1cspfar1?file=src%2Findex.tsx

### Steps to reproduce

1. Type in to the input
2. See that the form is dirty
3. Remove the value from the input
4. See that the form remains dirty
5. Reset the form
6. See that the form is no longer dirty

### Expected behavior

1. Type into the input
2. See that the form is dirty
3. Remove the value from the input (effectively resetting it)
4. See that the form is no longer dirty

None

### Screenshots or Videos

_No response_

### Platform

- Windows
- Chrome Version 131.0.6778.140

### TanStack Form adapter

None

### TanStack Form version

0.40.2

### TypeScript version

_No response_

### Additional context

_No response_

harry-whorlow:
Closes #1080 & #1081


packages/form-core/tests/FieldApi.spec.ts
```diff
@@ -77,27 +78,53 @@ describe('field api', () => {
       name: 'name',
       defaultMeta: {
         isTouched: true,
-        isDirty: true,
-        isPristine: false,
```
harry-whorlow, line 81: This is derived data that comes from isDirty, the test can be misleading since, you can't set it here it's wiped when the field is mounted. the only reason the tests passes is because is dirty is true.

LeCarbonator:
Sincd isDirty and isPristine are opposites, how about merging the two rows in the visualization? That could reduce the verticality of the image

packages/form-core/src/utils.ts
```diff
@@ -5,7 +5,7 @@ import type {
   ValidationSource,
 } from './types'
 import type { FormValidators } from './FormApi'
-import type { AnyFieldMeta, FieldValidators } from './FieldApi'
```
harry-whorlow, line 8: not used

packages/form-core/src/utils.ts
```diff
@@ -443,3 +443,37 @@ export const determineFieldLevelErrorSourceAndValue = ({
 
   return { newErrorValue: undefined, newSource: undefined }
 }
+
+export const deepEqual = <T>(a: T, b: T): boolean => {
```
harry-whorlow, line 447: if there's an equivalent function in the code base or dependancies let me know, I'll swap it out. 😄

docs/framework/react/guides/basic-concepts.md
```diff
@@ -92,17 +92,42 @@ const {
...
 ```
 
-There are three field states that can be useful to see how the user interacts with a field: A field is _"touched"_ when the user clicks/tabs into it, _"pristine"_ until the user changes value in it, and _"dirty"_ after the value has been changed. You can check these states via the `isTouched`, `isPristine` and `isDirty` flags, as seen below.
+There are three states in the metadata that can be useful to see how the user interacts with a field:
+
+- _"touched"_, after the user clicks/tabs into the field
+- _"pristine"_, until the user changes the field value
+- _"dirty"_, after the fields value has been changed
+
+You can check these states via the `isTouched`, `isPristine`, and `isDirty` flags, as seen below.
```
Balastrong, line 101: Why not directly using the flag names in the bullet point so we can avoid the other sentence?

Also we might mention `isDefaultValue` here

Balastrong (changes requested):
There are a few things to tweak but overall this will be a great addition!

packages/form-core/src/utils.ts
harry-whorlow: ah... man, thats painfull

packages/form-core/tests/FieldApi.spec.ts
```diff
@@ -77,27 +78,98 @@ describe('field api', () => {
...
+      // derived meta data
       isValid: true,
+      isValidating: false,
       errors: [],
       errorMap: {},
       errorSourceMap: {},
     })
   })
 
+  it('should update the fields meta isDefaultValue', () => {
```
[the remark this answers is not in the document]
harry-whorlow: Yeah that makes sense... I was thinking that the tests of deepValue would cover these cases. But, locking it down further also seems like a good idea. 😄

packages/form-core/src/FormApi.ts
```diff
@@ -912,26 +922,32 @@ export class FormApi<
           }
 
           // As primitives, we don't need to aggressively persist the same referential value for performance reasons
-          const isFieldPristine = !currBaseVal.isDirty
           const isFieldValid = !isNonEmptyArray(fieldErrors ?? [])
+          const isFieldPristine = !currBaseMeta.isDirty
+          const isDefaultValue = deepEqual(
+            curFieldVal,
+            this.options.defaultValues?.[fieldName as never],
```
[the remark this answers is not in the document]
harry-whorlow: hmm, good question

I suppose a follow up question is what if the defaultValues from form and field don't match? would it become a union? or shouldn’t this happen?

packages/form-core/src/FormApi.ts
```diff
@@ -981,6 +1004,9 @@ export class FormApi<
 
         const isTouched = fieldMetaValues.some((field) => field.isTouched)
         const isBlurred = fieldMetaValues.some((field) => field.isBlurred)
+        const isDefaultValue = fieldMetaValues.some(
+          (field) => field.isDefaultValue,
+        )
```
Balastrong, line 1009: I think this should be `every` (and we can add a unit test)

packages/form-core/src/utils.ts
Balastrong: Why do we need that extra flag? I mean, maybe it has always been a bug that we didn't check inside arrays :)

harry-whorlow: @Balastrong, I see where you're coming from. However, I would question as to why the function was named shallow, if it was not intended to do a shallow check.

Balastrong: Let's try to ignore the name for a moment and look at where it was used:

The function was used to check if the current state is different from the previous one, if they are the same there's no need to rerender but if they have different values then a rerender is triggered

The old implementation was likely flagging more often than needed if for example you swapped an object with another one having the same values. It caused more rerenders but was fast to evalute.

Always doing the deep check might be even better as it would cause less rerenders or am I missing something? 🤔

harry-whorlow: No, I think you're onto something there... Alright I'll swap it out 😄

Balastrong (approved):
I think we're in a really good state and it's probably enough to merge, if we missed something or there's room for improvement we can collect more feedback and tackle that on another PR :)

Thank you for addressing all the comments!

PR #1469, opened 2025-04-30
feat(react-form): Add `withFieldGroup`

### LINKED ISSUE #1296: withForm types break if defaultValues includes any extra field
### Describe the bug

Using withForm breaks whenever any extra field is added to the parents defaultValues.

```
import './App.css'
import { useAppForm, withForm } from './hooks/form'

function App() {

  const form = useAppForm({
    defaultValues: {
      form1: '',
      form2: ''
    }
  })

  return (
    <div>
      <ChildForm1 form={form} />
    </div>
  )
}

export default App


const ChildForm1 = withForm({
  defaultValues: {
    form1: ''
  },
  render: ({form}) => {
    return (
      <div />
    )
  }
})
```

https://github.com/ljukas/tanstack-withform-multiple

### Steps to reproduce

Create a child form using withForm from the createHookForm api. Use `useAppForm`, add the required defaultValues for the child form, then add any extra field to the parents defaultValues. The child-form form props types will no give a red squiggly line.

This also means you cannot compose multiple child forms together, or you can but you get type errors on the childform form-prop

### Expected behavior

As I user I expect the child-form to only care about the fields that it itself needs. So that I can compose multiple child forms together, which I expect is the use case from the start.

Every time

### Screenshots or Videos

_No response_

### Platform

All

### TanStack Form adapter

react-form

### TanStack Form version

1.1.0

### TypeScript version

5.7.2

### Additional context

_No response_

LeCarbonator:
Closes #1296 

### Todos
- [x] Implementation
- [x] Unit tests
- [x] Documentation
- [x] Fixes

This PR implements a variation of `withForm` that can be used to create form groups. This form group allows extending `defaultValues` and has **no expectations of form level validators**.

This distinguishes it both from `withForm` as well as instantiations of forms.

Here's an extract of the documentation:

## Reusing groups of fields in multiple forms

Sometimes, a pair of fields are so closely related that it makes sense to group and reuse them — like the password example listed in the [linked fields guide](./linked-fields.md). Instead of repeating this logic across multiple forms, you can utilize the `withFieldGroup` higher-order component.

> Unlike `withForm`, validators cannot be specified and could be any value.
> Ensure that your fields can accept unknown error types.

Rewriting the passwords example using `withFieldGroup` would look like this:

```tsx
const { useAppForm, withForm, withFieldGroup } = createFormHook({
  fieldComponents: {
    TextField,
    ErrorInfo,
  },
  formComponents: {
    SubscribeButton,
  },
  fieldContext,
  formContext,
})

type PasswordFields = {
  password: string
  confirm_password: string
}

// These default values are not used at runtime, but the keys are needed for mapping purposes.
// This allows you to spread `formOptions` without needing to redeclare it.
const defaultValues: PasswordFields = {
  password: '',
  confirm_password: '',
}

const FieldGroupPasswordField = withFieldGroup({
  defaultValues,
  // You may also restrict the group to only use forms that implement this submit meta.
  // If none is provided, any form with the right defaultValues may use it.
  // onSubmitMeta: { action: '' }

  // Optional, but adds props to the `render` function in addition to `form`
  props: {
    // These default values are also for type-checking and are not used at runtime
    title: 'Password',
  },
  // Internally, you will have access to a `group` instead of a `form`
  render: function Render({ group, title }) {
    // access reactive values using the group store
    const password = useStore(group.store, (state) => state.values.password)
    // or the form itself
    const isSubmitting = useStore(
      group.form.store,
      (state) => state.isSubmitting,
    )

    return (
      <div>
        <h2>{title}</h2>
        {/* Groups also have access to Field, Subscribe, Field, AppField and AppForm */}
        <group.AppField name="password">
          {(field) => <field.TextField label="Password" />}
        </group.AppField>
        <group.AppField
          name="confirm_password"
          validators={{
            onChangeListenTo: ['password'],
            onChange: ({ value, fieldApi }) => {
              // The form could be any values, so it is typed as 'unknown'
              const values: unknown = fieldApi.form.state.values
              // use the group methods instead
              if (value !== group.getFieldValue('password')) {
                return 'Passwords do not match'
              }
              return undefined
            },
          }}
        >
          {(field) => (
            <div>
              <field.TextField label="Confirm Password" />
              <field.ErrorInfo />
            </div>
          )}
        </group.AppField>
      </div>
    )
  },
})
```

We can now use these grouped fields in any form that implements the default values:

```tsx
// You are allowed to extend the group fields as long as the
// existing properties remain unchanged
type Account = PasswordFields & {
  provider: string
  username: string
}

// You may nest the group fields wherever you want
type FormValues = {
  name: string
  age: number
  account_data: PasswordFields
  linked_accounts: Account[]
}

  name: '',
  age: 0,
  account_data: {
    password: '',
    confirm_password: '',
  },
  linked_accounts: [
    {
      provider: 'TanStack',
      username: '',
      password: '',
      confirm_password: '',
    },
  ],
}

function App() {
  const form = useAppForm({
    defaultValues,
    // If the group didn't specify an `onSubmitMeta` property,
    // the form may implement any meta it wants.
    // Otherwise, the meta must be defined and match.
    onSubmitMeta: { action: '' },
  })

  return (
    <form.AppForm>
      <PasswordFields
        form={form}
        // You must specify where the fields can be found
        fields="account_data"
        title="Passwords"
      />
      <form.Field name="linked_accounts" mode="array">
        {(field) =>
          field.state.value.map((account, i) => (
            <PasswordFields
              key={account.provider}
              form={form}
              // The fields may be in nested fields
              fields={`linked_accounts[${i}]`}
              title={account.provider}
            />
          ))
        }
      </form.Field>
    </form.AppForm>
  )
}
```

### Mapping field group values to a different field

You may want to keep the password fields on the top level of your form, or rename the properties for clarity. You can map field group values
to their true location by changing the `field` property:

> [!IMPORTANT]
> Due to TypeScript limitations, field mapping is only allowed for objects. You can use records or arrays at the top level of a field group,
> but you will not be able to map the fields.

```tsx
// To have an easier form, you can keep the fields on the top level
type FormValues = {
  name: string
  age: number
  password: string
  confirm_password: string
}

  name: '',
  age: 0,
  password: '',
  confirm_password: '',
}

function App() {
  const form = useAppForm({
    defaultValues,
  })

  return (
    <form.AppForm>
      <PasswordFields
        form={form}
        // You can map the fields to their equivalent deep key
        fields={{
          password: 'password',
          confirm_password: 'confirm_password',
          // or map them to differently named keys entirely
          // 'password': 'name'
        }}
        title="Passwords"
      />
    </form.AppForm>
  )
}
```

If you expect your fields to always be at the top level of your form, you can create a quick map
of your field groups using a helper function:

```tsx
const defaultValues: PasswordFields = {
  password: '',
  confirm_password: '',
}

const passwordFields = createFieldMap(defaultValues)
/* This generates the following map:
 {
    'password': 'password',
    'confirm_password': 'confirm_password'
 }
*/

// Usage:
<PasswordFields
  form={form}
  fields={passwordFields}
  title="Passwords"
/>
```


LeCarbonator:
Issues that could be addressed:

- If the instantiated form has `onSubmitMeta`, it's no longer compatible with the form group. Perhaps if onSubmitMeta is unset and/or not called inside form Group, it should allow it outside.

LeCarbonator:
The unit tests should be reusable in case this isn't the desired approach.

LeCarbonator:
Related PR: #1334

LeCarbonator:
Something to consider:

- You cannot nest these form groups (Address form group vs. Address form group in an array)
- They can be conflicting with other groups (two groups sharing the `name` path). This can be averted by assuming a path namespace like `address.name

LeCarbonator:
Strange, the derived state from the form lens isn't updating ...

LeCarbonator:
There's currently a difference between a field's name and its actual field name. Looking into it later

LeCarbonator:
image
reminding myself for tomorrow. Definitely not useable like this.

packages/form-core/src/FormLensApi.ts
```diff
@@ -0,0 +1,709 @@
...
+
+  /**
+   * Validates all fields using the correct handlers for a given validation cause.
+   */
+  validateAllFields: (cause: ValidationCause) => Promise<unknown[]>
+
+  /**
+   * Validates the children of a specified array in the form starting from a given index until the end using the correct handlers for a given validation type.
+   */
+  validateArrayFieldsStartingFrom = async <TField extends DeepKeys<TLensData>>(
```
LeCarbonator, line 411: Use `DeepKeysOfType` to be in line with v1.11.1

See also pushFieldValue, insert, etc.

packages/form-core/src/FormLensApi.ts
```diff
@@ -0,0 +1,709 @@
...
+
+  /**
+   * Returns form, lens and field level errors
+   */
+  getAllErrors = (): {
+    form: {
+      errors: unknown[]
+      errorMap: FormValidationErrorMap
+    }
+    lens: {
```
LeCarbonator, line 619: This is the only place to get lens level errors. It should be added to `store`.

packages/form-core/src/FormLensApi.ts
```diff
@@ -0,0 +1,709 @@
...
+      ),
+    }
+  }
+
+  /**
+   * Parses the lens values with a given standard schema and returns
+   * issues (if any). This method does NOT set any internal errors.
+   * @param schema The standard schema to parse the lens values with.
+   */
+  parseValuesWithSchema = (schema: StandardSchemaV1<TLensData, unknown>) => {
```
LeCarbonator, line 689: Is there a good use case for these in lenses?

packages/form-core/tests/FormApi.spec.ts
```diff
@@ -3639,3 +3639,61 @@ it('should mark sourceMap as undefined when async field error is resolved', asyn
...
+  form.mount()
+
+  form.setFieldValue('shallow', 'Shallow')
+  form.setFieldValue('nested.field.name', 'Nested')
+
+  expect(form.state.values.shallow).toEqual('Shallow')
+  expect(form.state.values.nested.field.name).toEqual('Nested')
+
+  form.resetField('shallow')
+  expect(form.state.values.shallow).toEqual('Shallow')
```
bpinto, line 3695: I couldn't find any docs reference to my question so I'm doing it here, shouldn't `reset` without a value reset the form back to its original form's defaultValue? In this case `''`?

packages/react-form/tests/createFormHook.test.tsx
```diff
@@ -112,4 +113,298 @@ describe('createFormHook', () => {
...
+      return <ChildForm form={form} name="person" />
+    }
+
+    const { getByLabelText } = render(<Parent />)
+
+    const input = getByLabelText('person.firstName')
+    input.focus()
+    expect(input).toHaveFocus()
+
+    await userEvent.clear(input)
```
bpinto, line 347: You may want to call [userEvent.setup()](https://testing-library.com/docs/user-event/setup/) instead of using `userEvent` directly. I believe I've seen this usage on the repository already.

packages/react-form/src/createFormHook.tsx
```diff
@@ -362,8 +630,99 @@ export function createFormHook<
...
+            name: lens.getFormFieldName(innerProps.name),
+            defaultValues,
+          }
+        } else {
+          return {
+            ...innerProps,
+            defaultValues,
+          }
+        }
+      }, [innerProps])
```
bpinto, line 716: Using an object as a dependency isn't great because an object will, unless memoized itself, be itself a new object on every render.

Since you are only interested in tracking `form` and `name` and both are probably stable as `form` is memoized and `name` is a string, I think it would be more efficient if you were to destructure these two properties from `innerProps` and use them as dependencies directly: `[form, name]`. Or else `[innerProps.form, innerProps.name]`.

packages/form-core/tests/FormApi.spec.ts
bpinto: @LeCarbonator was kind enough to explain to me on discord that there is no default value set on this test so the expectation is correct.

packages/react-form/src/createFormHook.tsx
LeCarbonator: The issue is that `innerProps` is spread inside the memo. This could usually be solved with destructuring all variables like you mentioned, but this `innerProps` additionally receives `NoInfer<TRenderProps>` (a record of any string key and any value).

packages/react-form/tests/createFormHook.test.tsx
LeCarbonator: Thanks! Yes, other files appear to call it at the top of the file. I've changed `createFormHook.test.tsx` to use the same setup.

packages/react-form/src/createFormHook.tsx
```diff
@@ -229,6 +385,132 @@ export function createFormHook<
...
+        TSubmitMeta,
+        TComponents,
+        TFormComponents
+      >
+    })
+
+    useIsomorphicLayoutEffect(formLensApi.mount, [formLensApi])
+
+    return formLensApi
+  }
```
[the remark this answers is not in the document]
LeCarbonator: Only if we can somehow access app field types, which are not currently exported. If that change is fine by you, I'll gladly move it to its own file

packages/form-core/src/FormLensApi.ts
```diff
@@ -0,0 +1,758 @@
...
+    // TODO find better solution to chaining names. This is prone to breaking if the syntax ever changes
+
+    // If lensData is an array at top level, the name
+    // would be [i].name
+    // this would be incompatible with the form as it's expected
+    // to receive lens.prefix[i].name and NOT lens.prefix.[i].name
+    if (subfield.charAt(0) === '[') {
+      return this.lensPrefix + subfield
+    }
+    return `${this.lensPrefix}.${subfield}`
```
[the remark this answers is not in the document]
LeCarbonator: Done! `concatenatePaths`

packages/form-core/src/FormLensApi.ts
```diff
@@ -0,0 +1,758 @@
...
+        return {
+          ...rest,
+          lensErrorMap,
+          lensErrors,
+          formErrorMap,
+          formErrors,
+          values: this.form.getFieldValue(this.name) as TLensData,
+        }
+      },
+    })
```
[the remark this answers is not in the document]
LeCarbonator: The thing is, the lens itself should ideally not exist. It should merely be a translator between fields and a form using some form of mapping.

With this in mind, i think I will remove most state values as they are not related to it and can be accessed differently (submissionAttempts getting fetched from `group.form.store` instead of directly)

However, `values` would be convenient to have derived somehow for easier `useStore` and `Subscribe`. I'm just not sure how to go about it.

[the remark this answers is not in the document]
harry-whorlow: I would lean towards 2 myself, it feels odd to have a separate state outside the form. Though it seems we're all in agreement

packages/form-core/src/FormLensApi.ts
```diff
@@ -0,0 +1,758 @@
...
+  FormValidationErrorMap,
+  UpdateMetaOptions,
+  ValidationCause,
+  ValidationError,
+  ValidationErrorMap,
+} from './types'
+import type { StandardSchemaV1 } from './standardSchemaValidator'
+
+// What wasn't implemented from FormApi:
+// - setErrorMap would require type-safe setting of the validators. Don't implement it.
```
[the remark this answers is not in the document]
LucaJakob: It was a footnote so I don't accidentally implement it. Basically, you would need to know validator data for it to make any sense, so it's useless for field groups.

packages/form-core/src/FormLensApi.ts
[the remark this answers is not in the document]
LeCarbonator: The revised version tries to keep as little store as possible in the field group, since it's not its true purpose. Thoughts? @harry-whorlow @crutchcorn

MPiland:
@LeCarbonator  I've been toying this PR (I know it's not done) because it's exactly what I need. I'm not an experienced developer by any means, but I am a product manager by trade and so I'm used to thinking about user experience or in this case developer experience. Can I suggest that instead a fields prop that either takes a string or an object of key value pairs where you map a form property to the group property, could we instead have a group prop and we pass the actual form object to a group prop or map the destination property to the form property directly? That could be more intuitive since we pass the form prop to the withForm hook and this would feel similar in experience. I don't know the feasibility with that and maintaining type safety, though.

LeCarbonator:
@MPiland could you provide a code snippet of what that structure may look like? It doesn't need to be feasible, just a snippet of what you envisioned the API to be

MPiland:
@LeCarbonator  I think there's a couple ways that could maybe work. The prop could be just named group that could either accept an object or an object of properties mapped to form properties. This would probably require a form subscribe.

```
/* As object */
<form.Subscribe select={(state) => state.values)}>
  {(values) => {
    <Auth group={values.auth} form={form}  />
   }
</form.Subscribe>

/* As properties mapped to form properties */
<form.Subscribe select={(state) => state.values)}>
  {(values) => {
    <Auth group={{username: values.username, password: values.password}} form={form}  />
   }
</form.Subscribe>
```

Alternatively, you could add a new option to the createFormHook called groupComponents. They could be called similar to field components, but their name prop could take a string or an object of key values similar to how the withFieldGroup works. Then it could use a useGroupContext. This is all completely brainstorming so it may not be possible. But you could call it like

```
<form.AppGroup name='auth'>
  {({ AuthGroup }) => (
    <AuthGroup {...props} />
  )} />
</form.AppGroup>

/* or */

<form.AppGroup name={{ username: 'username', password: 'password' }}>
  {( group ) => (
    <group.AuthGroup {...props} />
  )} />
</form.AppGroup>

/* Defining it could be like this */

export const AuthGroup = () => {
  const {state, group, AppField} = useGroupContext<AuthGroupType>()

  <state.AppField name='password'>
    {( field) => (
      <field.TextInput label='Password' />
    )}
  </state.AppField>
}

/* or */

<form.Group name='auth'> // Alternatively <form.Group name={{ password: 'password' }}>
  {( group ) => (
    <group.Field name='password'>
      .....
    </group.Field>
  )}
</form.Group>
```

The current approach still works, though, and if that makes more sense to other users, I'm ok being told I'm wrong. I'm just throwing out some ideas. I really like this library. It took me a minute to get it, but once I did it's very intuitive. I just need this last group feature to be able to use it.

LeCarbonator:
@MPiland Alright, here's my thoughts on the snippets:

```tsx
<form.Subscribe select={(state) => state.values)}>
  {(values) => {
    <Auth group={values.auth} form={form}  />
   }
</form.Subscribe>
```
I'm heavily against this implementation. The reason is that field groups aren't just displaying values, they map fields called within to the form outside. This would force it to be a one-way street which goes against a lot of the features other users want.

As for `<form.AppGroup/>`, I can see it working, but the benefits aren't very clear to me.
The reason `AppField` exists is because you have more flexibility for how you implement it:
```tsx
/* Do you want granular control? */
<form.AppField name="firstName">
  {field => <>
     <label>First Name</label>
     <field.TextInput />
     <field.Feedback />
  </>}
</form.AppField>

/* Or wrap it all in a component? */
<form.AppField name="firstName">
  {field => <field.TextInput showFeedback label="First Name" />}
</form.AppField>
```

Field groups don't really fit that category, as they're meant to do the work themselves. Once you have the field group, there's really only one way to use it, and that is like you showed:
```tsx
/* Combinations don't really exist. Either you use the field group or you don't */
<form.AppGroup name='auth'>
  {({ AuthGroup }) => (
    <AuthGroup {...props} />
  )} />
</form.AppGroup>
```

That's why I prefer the HOC implementation. Hopefully it's clear what I'm trying to convey.

MPiland:
@LeCarbonator That all makes sense to me! I will defer to your expertise. I think the original way works  as well. I appreciate you entertaining my thoughts!

TheoJV24:
Will this feature work if multiple forms share the same UI but have different fields?
In our case, each form is divided into tabs, each rendering a form with Section A and Section B.
Section B shares the same UI across forms, but some fields are present only in certain forms.

Example:
Cycling From — Section B: calories, elevationGain, distance
Swimming Form  — Section B: calories, distance
Gym Form  — Section B: calories

LeCarbonator:
> Will this feature work if multiple forms share the same UI but have different fields? In our case, each form is divided into tabs, each rendering a form with Section A and Section B. Section B shares the same UI across forms, but some fields are present only in certain forms.
> 
> Example: Cycling From — Section B: calories, elevationGain, distance Swimming Form — Section B: calories, distance Gym Form — Section B: calories

Not with its current implementation, no. The intent of this feature is to group multiple fields together so **they can all be reused**. The Section B you describe is not consistent across forms, but is instead a collection of a `calories` field, a `elevationGain` field and a `distance` field.
Since it's only single fields that remain consistent across forms, consider using a Field Component instead.

An example usage of this field group would be:

### Section: calories, elevationGain, distance
* Cycling Form - `calories`, `elevationGain`, `distance`
* Swimming Form - `data.calories`, `data.elevationGain`, `data.distance` (namespaced with `data`)
* Gym Form - `gymCalories`, `gymElevationGain`, `gymDistance` (mapped to `calories`, `elevationGain`, `distance`)

drahmedshaheen:
`The inferred type of useAppForm cannot be named without a reference to ../../../node_modules/@tanstack/react-form/dist/esm/createFormHook . This is likely not portable. A type annotation is necessary.`

It goes for all createFormHook extracts.

LeCarbonator:
Good catch! I'll add those annotations tomorrow (alongside fixing the failing unit tests at the moment)

LeCarbonator:
As discussed on Discord, it appears to have been because of pnpm cache instead of a problem with the PR.

AfrazHussain:
@LeCarbonator  Firstly, thanks so much for putting this together. I've been really looking forward to this feature! 🙌

Just wondering if there's anything blocking this from being merged, or if any help is needed to move it forward. Also curious if there's any idea of when it might be released (no pressure, just excited to try it out!).

LeCarbonator:
@AfrazHussain not particularly. Things have slowed down a notch since it is the middle of summer, but apart from reviewing the revised version, there's no blocking part of this PR.

The PR is installable using the scripts provided by [this comment](https://github.com/TanStack/form/pull/1469#issuecomment-2843118451). Of course it's not part of the main branch yet, but it's testable.

I'm sure it will be released soon! If you notice a problem in the current code, feel free to leave a review comment so it can be addressed.

AfrazHussain:
> The PR is installable using the scripts provided by https://github.com/TanStack/form/pull/1469#issuecomment-2843118451.

Ah, that is quite helpful. Thanks a lot! I'll try it out and then report back if there's any feedback/issues.

ArtemFH:
> `The inferred type of useAppForm cannot be named without a reference to ../../../node_modules/@tanstack/react-form/dist/esm/createFormHook . This is likely not portable. A type annotation is necessary.`
> 
> It goes for all createFormHook extracts.

Did you manage to figure it out? I don't understand why you were ignored, just to stumble across the test boxes.

LeCarbonator:
@ArtemFH The discussion was continued on Discord, so it looks like it was lost here.

`tsconfig.json` likely has `declaration: true` enabled, and TypeScript warns you that those generated declaration files will not be able to name the exported functions.

We'll fix it by exporting the types very soon, but in case you need a current workaround, that configuration should be turned off.

zhouzi:
What's the recommended workaround for those that do need to emit declarations?

If there's none, it's ok. I'll wait for the types to be exported.

LeCarbonator:
No pretty ones apart from going into TanStack Form's declaration files and exporting the reported types @zhouzi

Harukisatoh:
@LeCarbonator you mentioned this api is intended to be used when fields are closely related but that's still not so clear to me. Could you give examples of use cases? Also, could that be used for a multi-step form?

LeCarbonator:
@Harukisatoh Here's an example of fields that depend on each other and benefit greatly from field groups:

Example: I have 5 forms that all include intervals or date ranges. For that, I want a start date and an end date, and ensure that the end date cannot be set before the start date. Since the user can set the time, I would also like to format the fields on blur so that `12` becomes `12:00` automatically. If I create fields with that logic, I would have to repeat that logic for all forms that use it, and if I overhaul the UI or behaviour, I have to go and search for all locations where I used it.

Field groups allow me to not only make a group of both fields, but also use 'references' to other fields - Which means I can decide how granular those groups should be

```tsx
// Case 1: my field group creates fields for Start and End times.
<IntervalFields form={form} fields={{
   startDate: 'my.nested.start.date',
   endDate: 'my.nested.end.date'
}} />
```

Notice how I now forced my fields to be together. What if I want to make them stack vertically instead of horizontally? What if something is supposed to go between them?

With field groups,  I can go for this structure instead:
```tsx
// Case 2: Two field groups, one for each field
<IntervalStartField form={form} fields={{
   // will create the field
   startDate: 'my.nested.start.date',
   // won't be a field, but I can reference it for setting / reading / subscribing values 
   endDateReference: 'my.nested.end.date' 
}} />

<AnythingIWant />

<IntervalEndField form={form} fields={{
   // will create the field
   endDate: 'my.nested.end.date',
   // won't be a field, but I can reference it for setting / reading / subscribing values 
   startDateReference: 'my.nested.start.date' 
}} />
```

Does that clear it up?

zhouzi:
Thanks for #1661 @LeCarbonator 🙏

I upgraded to v1.15.3 and the migration from `withForm` to `withFieldGroup` was smooth, as can be seen in https://github.com/zhouzi/akimeo/commit/9c15c853004e9f9100c21281a9eb6254785eea3c

PR #1482, opened 2025-05-06
fix: only allow setting manual errors according to the defined validators

Balastrong:
At the current state the types of `form.setErrorMap` are wrong (asking for a validator function) and `field.setErrorMap` allows everything.

With this PR the allowed types are now exactly the ones coming from the defined validators to ensure type safety is correct when reading the values (e.g. from `field.getMeta().errors`)

### DIFF
--- a/examples/react/field-errors-from-form-validators/src/index.tsx
+++ b/examples/react/field-errors-from-form-validators/src/index.tsx
@@ -112,8 +112,7 @@ export default function App() {
             errorMap.onSubmit ? (
               <div>
                 <em>
-                  There was an error on the form:{' '}
-                  {errorMap.onSubmit?.toString()}
+                  There was an error on the form: {errorMap.onSubmit.toString()}
                 </em>
               </div>
             ) : null
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -1671,17 +1671,24 @@ export class FieldApi<
   /**
    * Updates the field's errorMap
    */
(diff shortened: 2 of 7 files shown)


frenzzy:
How should server errors be set now?
Especially for server errors that are not tied to a specific field:

```ts
form.setErrorMap({ onServer: 'login or password is incorrect' });
// onServer: Type 'string' is not assignable to type 'undefined'.
```

Balastrong:
You shouldn't need to manually set server errors as they're automatically set from your [server validator](https://tanstack.com/form/latest/docs/framework/react/guides/ssr) or in general by returning something from any [other validator](https://tanstack.com/form/latest/docs/framework/react/guides/validation#setting-field-level-errors-from-the-forms-validators).

This also gives lets `form.errors` & co infer the correct error type

MGRatEJOT:
How would I set an error that comes from a react-query mutation that was fired off on submit and failed? Previously this was no problem.

vojtech-dobes:
I apologize upfront if this question is just wasting everybody's time. But could someone confirm, whether return field errors from `onSubmitAsync` is similarly type-checked (TS would complain about returning error for non-existent field)?

Balastrong:
I don't think I'm understanding the question, onSubmitAsync follows the same rules of the other validators and types are fully inferred.
If you think there's something not working right please create an issue with all the details so we can have a look :)

vojtech-dobes:
@Balastrong Apologies, here is proper issue: https://github.com/TanStack/form/issues/1537. Thanks!

Issue #1490, opened 2025-05-07
form.reset(value) is not resetting defaultValues to a provided value


ilialksv:
### Describe the bug

It seems like `form.reset(value)` is updating `defaultValues`, but by the next render `defaultValues `are being reset to the initially provided ones. So, as a result, there is no difference whether to call `form.reset(value)` or just `form.reset()` because they lead to the same result.

### Your minimal, reproducible example

https://stackblitz.com/edit/vitejs-vite-x8k3rfrs?file=src%2FApp.tsx

### Steps to reproduce

1. Go to the Stackblitz url.
2. Change values in input(s).
3. Click "Submit" button.
4. After submit, `defaultValues` are being reset to the new ones (you can see it in console).
5. By the next render, `defaultValues `are back to the initial ones.

### Expected behavior

As a user I expect that `defaultValues `are updated with the values provided to `form.reset` method and not being reset to the initial `defaultValues`.

### How often does this bug happen?

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/bf1f3a75-9436-44a1-9015-1d282185f425

### Platform

- OS: Windows 11
- Browser: Chrome
- Version: 135.0.7049.116

### TanStack Form adapter

react-form

### TanStack Form version

v1.9.1

### TypeScript version

v5.8.3

xargr:
the question here is what's the ideal behaviour and if the default value meant to be always rendered when there isn't value or they have to be rendered only in the first rendering

ilialksv:
Thanks for the explanation! But it's a bit of a confusing behavior. I think it's quite a common use case of this method, when you need to reset a value after updating some data inside an asynchronous function.

It would be great if this method would work correctly in this case too, so that we wouldn't have to invent ways around it.

harry-whorlow:
@ilialksv, no you've miss understood me, I think it's a bug... only, I was stating where its happening. 😄 

I've narrowed it down a little further, form core handles this correctly, the react adaptor is where this appears to be happening, well it might be happening in the other adapters I haven't checked. 

Theres a few issues like this that have opened in the past couple days, let me dig around a bit.

ilialksv:
@harry-whorlow Got it! Thanks for clarifying and for looking into it further.

By the way, I just wanted to express my appreciation for this package. It has been a great help in my projects. Thank you and the team for your hard work!

harry-whorlow:
[...]

Issue #1487, opened 2025-05-07
validateAllFields immediately after form.reset doesn't work


jashworth:
### Describe the bug

form.validateAllFields (line 130) does not appear to revalidate the fields after form.reset({ ... })

...at least not synchronously (see lines 133-135)

### Your minimal, reproducible example

https://codesandbox.io/p/devbox/fervent-ives-32x45k

### Steps to reproduce

1. Observe the initial error ("must not be before 2025-06-02") in the Run Date field of the form
2. Select the "Dead" radio button
3. Select the "Once" radio button
4. Observe the date is the same as before but the error has gone, despite the call to form.validateAllFields (line 130)

### Expected behavior

form.validateAllFields (line 130) revalidates all fields, resulting in the error being displayed

### How often does this bug happen?

Every time

### Platform

Windows
Chrome (latest)

### TanStack Form adapter

react-form

### TanStack Form version

1.90.0

### TypeScript version

5.7.2

### Additional context

The aim is to use form.reset({ ... }) to set multiple co-variant fields together, revealing any constraint violations to the user as soon as the values have been set.

Is there an alternative way to revalidate after reset, preferably one that doesn't rerender the entire component?

harry-whorlow:
@jashworth I think you've left your codeSandbox as private, as clicking the link provided doesn’t open one 😅 

I haven't looked but do you think this could be related to #1485, if you could do a little investigation it would help us out a bunch 😄 if not we will take a look when we get some time!

jashworth:
Oops!

is public now

harry-whorlow:
I'm closing this issue as it been fixed since 😊

Issue #1485, opened 2025-05-07
Bug: form.reset(value) seems not caring about value. It's reseting to the defaultValues anyway.


LumaKernel:
### Describe the bug

I expected form.reset with the first arg, it resets with setting the default values to that arg, as written in JSDoc of the API.

But it seems resetting to the value as well as not supplying the arg.

### Your minimal, reproducible example

https://github.com/LumaKernel/tanstack-form-issue-reset-with-values

### Steps to reproduce

1. Clone repro repo.
2. npm i
3. npm run dev
4. Put values
5. Submit
6. Reset to the blank values.

### Expected behavior

Reset to the value that is specified by arg. So, just the states reset.

### How often does this bug happen?

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/87e0afb2-3751-42db-980d-88b15115bb54

### Platform

  Node: v22.14.0
  npm: 10.5.0

### TanStack Form adapter

react-form

### TanStack Form version

1.9.1

### TypeScript version

5.8.3

harry-whorlow:
@LumaKernel check out this [stackblitz](https://stackblitz.com/github/tanstack/form/tree/main/examples/react/simple?embed=1&theme=dark&preset=node&file=src/index.tsx). I'm not sure what you've done, or perhaps the issue has been resolved in the later patch, but reset dose seem to be working. 😄

harry-whorlow:
relates to https://github.com/TanStack/form/issues/1490

LeCarbonator:
#1494 has been merged. Update your Tanstack Form version to v1.14.1.

PR #1494, opened 2025-05-09
chore: test cases

harry-whorlow:
Relates to #1490, #1485 and potentially #1487.

This appears to stem from the react adapter as Angular, Vue and Core all seem to work as intended.

reset will work correctly initially then on next pass wipe the previous state. Curiously this seems to only be the case in the onSubmit handler.

### DIFF
--- a/packages/angular-form/tests/test.component.spec.ts
+++ b/packages/angular-form/tests/test.component.spec.ts
@@ -363,3 +363,52 @@ describe('TanStackFieldDirective', () => {
     expect(getByText(onBlurError)).toBeInTheDocument()
   })
 })
+
+describe('form should reset when rendered correctly - angular', () => {
+  it('should be able to handle async resets', async () => {
+    @Component({
+      selector: 'test-component',
+      standalone: true,
+      template: `
+        <ng-container [tanstackField]="form" name="name" #f="field">
+          <input
+            data-testid="fieldinput"
+            [value]="f.api.state.value"
+            (input)="f.api.handleChange($any($event).target.value)"
+          />
+        </ng-container>
+        <button
+          type="button"
+          (click)="form.handleSubmit()"
+          data-testid="submit"
+        >
+          submit
+        </button>
+      `,
+      imports: [TanStackField],
+    })
+    class TestComponent {
+      form = injectForm({
+        defaultValues: {
+          name: '',
+        },
+        onSubmit: ({ value }) => {
+          expect(value).toEqual({ name: 'test' })
+          this.form.reset({ name: 'test' })
+        },
+      })
+    }
+
+    const { getByTestId } = await render(TestComponent)
+
+    const input = getByTestId('fieldinput')
+    const submit = getByTestId('submit')
+
+    await user.type(input, 'test')
+    await expect(input).toHaveValue('test')
+
+    await user.click(submit)
+
+    await expect(input).toHaveValue('test')
+  })
+})
--- a/packages/form-core/tests/FormApi.spec.ts
+++ b/packages/form-core/tests/FormApi.spec.ts
@@ -120,6 +120,25 @@ describe('form api', () => {
     })
   })
 
+  it('form should reset when rendered correctly - core', async () => {
+    const defaultValues = {
+      name: '',
+    }
+    const form = new FormApi({
+      defaultValues: defaultValues,
+      onSubmit: ({ value }) => {
+        form.reset(value)
+
+        expect(form.options.defaultValues).toMatchObject({
+          name: 'test',
+        })
+      },
+    })
+    form.mount()
+    form.setFieldValue('name', 'test')
+    form.handleSubmit()
+  })
+
   it('should reset and set the new default values that are restored after an empty reset', () => {
     const form = new FormApi({ defaultValues: { name: 'initial' } })
     form.mount()
--- a/packages/react-form/tests/useForm.test.tsx
+++ b/packages/react-form/tests/useForm.test.tsx
@@ -795,3 +795,53 @@ describe('useForm', () => {
     expect(fn).toHaveBeenCalledTimes(1)
   })
 })
+
+it('form should reset when rendered correctly - react', async () => {
+  function Comp() {
+    const form = useForm({
+      defaultValues: {
+        name: '',
+      },
+      onSubmit: ({ value }) => {
+        expect(value).toEqual({ name: 'test' })
+
+        form.reset({ name: 'test' })
+      },
+    })
+
+    return (
+      <div>
+        <form.Field
+          name="name"
+          children={(field) => (
+            <input
+              data-testid="fieldinput"
+              name={field.name}
+              value={field.state.value}
+              onChange={(e) => field.handleChange(e.target.value)}
+            />
+          )}
+        />
+
+        <button
+          type="button"
+          onClick={() => form.handleSubmit()}
+          data-testid="submit"
+        >
+          submit
+        </button>
+      </div>
+    )
+  }
+
+  const { getByTestId } = render(<Comp />)
+  const input = getByTestId('fieldinput')
+  const submit = getByTestId('submit')
+
+  await user.type(input, 'test')
+  await waitFor(() => expect(input).toHaveValue('test'))
+
+  await user.click(submit)
+
+  await waitFor(() => expect(input).toHaveValue('test'))
+})
--- a/packages/vue-form/tests/useForm.test.tsx
+++ b/packages/vue-form/tests/useForm.test.tsx
@@ -476,4 +476,55 @@ describe('useForm', () => {
     await waitFor(() => getByText(error))
     expect(getByText(error)).toBeInTheDocument()
   })
+
+  it('form should reset when rendered correctly - vue', async () => {
+    const Comp = defineComponent(() => {
+      const form = useForm({
+        defaultValues: {
+          name: '',
+        },
+        onSubmit: ({ value }) => {
+          expect(value).toEqual({ name: 'test' })
+
+          form.reset({ name: 'test' })
+        },
+      })
+
+      return () => (
+        <div>
+          <form.Field name="name">
+            {({ field }: { field: AnyFieldApi }) => (
+              <input
+                data-testid="fieldinput"
+                name={field.name}
+                value={field.state.value}
+                onInput={(e) =>
+                  field.handleChange((e.target as HTMLInputElement).value)
+                }
+              />
+            )}
+          </form.Field>
+
+          <button
+            type="button"
+            onClick={() => form.handleSubmit()}
+            data-testid="submit"
+          >
+            submit
+          </button>
+        </div>
+      )
+    })
+
+    const { getByTestId } = render(<Comp />)
+    const input = getByTestId('fieldinput')
+    const submit = getByTestId('submit')
+
+    await user.type(input, 'test')
+    await waitFor(() => expect(input).toHaveValue('test'))
+
+    await user.click(submit)
+
+    await waitFor(() => expect(input).toHaveValue('test'))
+  })
 })


bpinto:
I'm not sure if this is the proper fix, but this fixes the newly added test without introducing other test failures:

```
diff --git a/packages/react-form/src/useForm.tsx b/packages/react-form/src/useForm.tsx
index 729064ab..dd7a012c 100644
--- a/packages/react-form/src/useForm.tsx
+++ b/packages/react-form/src/useForm.tsx
@@ -210,13 +210,5 @@ export function useForm<
 
   useStore(formApi.store, (state) => state.isSubmitting)
 
-  /**
-   * formApi.update should not have any side effects. Think of it like a `useRef`
-   * that we need to keep updated every render with the most up-to-date information.
-   */
-  useIsomorphicLayoutEffect(() => {
-    formApi.update(opts)
-  })
-
   return formApi
 }
```

harry-whorlow:
@bpinto thanks for taking a look into this, I was rather busy over the weekend, so it's super appreciated!

hmm thats, interesting... I'll shoot crutch corn a message since he authored the code, but you are correct it dose pass the tests, tough to play devils advocate we are lacking framework specific tests so it might be breaking something. 😊

Once again thanks for the investigation!

LeCarbonator:
[these lines changing `this.options`](https://github.com/TanStack/form/blob/2a82baa488fc4bddf985ee94e9bc219c70dd9ce5/packages/form-core/src/FormApi.ts#L1260C2-L1265C6) appear to be the cause. Commenting them out makes the test pass.

LeCarbonator:
Testing on stackblitz, this appears to work as expected in react now!

packages/react-form/src/useForm.tsx
```diff
@@ -216,7 +225,7 @@ export function useForm<
    */
   useIsomorphicLayoutEffect(() => {
     formApi.update(opts)
-  })
+  }, [stableOptsRef.current])
```
bpinto, line 228: Given this value (`stableOptsRef.current`) is calculated on every re-render, I don't think using it as a `[deps]` adds any additional benefit over doing the check inside of the `useIsomorphicLayoutEffect` function.

```js
  useIsomorphicLayoutEffect(() => {
    if (!evaluate(opts, stableOptsRef.current) {
      formApi.update(opts)
    }
  })
```

The reason for why I bring this up, is because I personally would rather avoid using a `toString()` to compare functions inside the `evaluate` function. Therefore, there might be some additional checks we could do here that could make it possible for us to avoid that.

I don't know what the `shouldUpdateReeval` is for but for the other 2 updates, if the form has been touched (`!this.state.isTouched`) then the update is skipped. So maybe there is something we can check to avoid the `update()` call altogether and potentially not need to deal with function `toString()` comparison.

harry-whorlow: hmm... My concern here is that this issue only exists in the react sphere, all the other frameworks are working correctly, and modify core just to service a bug that only exists in react feels off. 

I think I'll shelve this for a short minuet to look into other solutions, I appreciate the input.

jsefiani:
@harry-whorlow I tested out this PR, and I see a UI glitch whenever I reset the form with the new values from the API. The initial values are briefly displayed and then replaced with the new values.

harry-whorlow:
> @harry-whorlow I tested out this PR, and I see a UI glitch whenever I reset the form with the new values from the API. The initial values are briefly displayed and then replaced with the new values.

From discord correct? this was unrelated if I'm not mistaken. 🤘

harry-whorlow:
So this is not an ideal solution, and we've talked about this internally, however we couldn’t come up with an alternative solution. 

I'll leave this up till Friday, so if anyone has any suggestions or other concerns they can post them here and I'll investigate. Otherwise I'll merge 🙂

bpinto:
I am not sure what type of API is meant to be implemented here so I wrote a failing tests below that may or may not be failing correctly.

On this test, one of the form options `asyncDebounceMs` is being updated which causes the `if (!evaluate(opts, stableOptsRef.current)` to evaluate `true`.

This is an alternative way of introducing the bug that this PR is trying to fix as the `defaultValues` of the `useForm()` overrides the latest default value specified via `reset()`.

```js
diff --git a/packages/react-form/tests/useForm.test.tsx b/packages/react-form/tests/useForm.test.tsx
index 7b012db7..d042c095 100644
--- a/packages/react-form/tests/useForm.test.tsx
+++ b/packages/react-form/tests/useForm.test.tsx
@@ -797,7 +797,10 @@ describe('useForm', () => {
 
   it('form should reset default value when resetting in onSubmit', async () => {
     function Comp() {
+      const [asyncDebounceMs, setAsyncDebounceMs] = useState(0)
+
       const form = useForm({
+        asyncDebounceMs,
         defaultValues: {
           name: '',
         },
@@ -835,6 +838,10 @@ describe('useForm', () => {
           <button type="reset" data-testid="reset" onClick={() => form.reset()}>
             Reset
           </button>
+
+          <button type="button" data-testid="update-form-options" onClick={() => setAsyncDebounceMs(10)}>
+            Update form options
+          </button>
         </form>
       )
     }
@@ -843,6 +850,7 @@ describe('useForm', () => {
     const input = getByTestId('fieldinput')
     const submit = getByTestId('submit')
     const reset = getByTestId('reset')
+    const updateFormOptions = getByTestId('update-form-options')
 
     await user.type(input, 'test')
     await waitFor(() => expect(input).toHaveValue('test'))
@@ -853,6 +861,9 @@ describe('useForm', () => {
     await user.type(input, 'another-test')
     await user.click(submit)
     await waitFor(() => expect(input).toHaveValue('another-test'))
+
+    await user.click(updateFormOptions)
+    await waitFor(() => expect(input).toHaveValue('another-test'))
   })
 
   it('form should update when props are changed', async () => {
```

---

A smaller patch simulating this issue:

```js
diff --git a/packages/react-form/tests/useForm.test.tsx b/packages/react-form/tests/useForm.test.tsx
index 7b012db7..d129ad8e 100644
--- a/packages/react-form/tests/useForm.test.tsx
+++ b/packages/react-form/tests/useForm.test.tsx
@@ -797,7 +797,10 @@ describe('useForm', () => {
 
   it('form should reset default value when resetting in onSubmit', async () => {
     function Comp() {
+      const [asyncDebounceMs, setAsyncDebounceMs] = useState(0)
+
       const form = useForm({
+        asyncDebounceMs,
         defaultValues: {
           name: '',
         },
@@ -805,6 +808,7 @@ describe('useForm', () => {
           expect(value).toEqual({ name: 'another-test' })
 
           form.reset(value)
+          setAsyncDebounceMs(10)
         },
       })
```

---

As far as I understand, the core of the issue isn't being addressed on this PR yet since invoking `formApi.update()` with a `defaultValues` that is different to the current form internal `defaultValues` will update the internal value overriding any previous change.

I believe the reason for this is to support [async initial values](https://tanstack.com/form/latest/docs/framework/react/guides/async-initial-values) however it seems that `defaultValues` property on `useForm` should not be reactive if the form internal defaultValues (or values) have ever been updated.

harry-whorlow:
@bpinto, thanks for the input! 😄 

Your first test is failing correctly though isn't it? Essentially your passing in the form's props when the asyncDebounceMs is changed:

```tsx
{ asyncDebounceMs,  defaultValues: { name: '' } }
```
So expecting "name: another-value" would fail.

I see your line of thinking though, but it doesn’t really match up with my model of how I expect state change. Correct me if I'm wrong, but you're suggesting that form should ignore any future changes made to the default values after a form.reset() is called? I'll run it by the other maintainers though, see what they think. 🙂

bpinto:
> I see your line of thinking though, but it doesn’t really match up with my model of how I expect state change. Correct me if I'm wrong, but you're suggesting that form should ignore any future changes made to the default values after a form.reset() is called? I'll run it by the other maintainers though, see what they think. 🙂

👍 Yep, exactly. I think that form should ignore any future changes made to default values after a `form.reset()` (or any form value change) but as I said on the beginning of my comment, I'm not sure if my expectation aligns with the library's.

However, I want to emphasize one thing that you are questioning to confirm we are talking about the same thing:

>  but you're suggesting that form should ignore any future changes made to the default values after a form.reset() is called?

in the tests above, the `defaultValues` passed to `useForm()` is **not** changing, but rather it's only `asyncDebounceMs` that is changing. The form _internal_ `defaultValues` is changing because of `reset()` invocation but the attribute `defaultValues` that is passed to `useForm` does not change in the test.

deshazer:
> @harry-whorlow I tested out this PR, and I see a UI glitch whenever I reset the form with the new values from the API. The initial values are briefly displayed and then replaced with the new values.

I was running into this same issue. I was submitting the form, `await`ing tanstack/query to refetch the data, and then calling `formApi.reset()`. I was seeing a brief flash of the previous data on reset. 

If it helps anyone, calling `setTimeout(() => formApi.reset())`  (instead of just `formApi.reset()`) gets rid of the flash. Feels hacky, but it works great. 👍

santiagazo:
> > @harry-whorlow I tested out this PR, and I see a UI glitch whenever I reset the form with the new values from the API. The initial values are briefly displayed and then replaced

https://github.com/user-attachments/assets/0c849fb8-c1e6-45e7-a9f3-5e403a35723f

 with the new values.
> 
> I was running into this same issue. I was submitting the form, `await`ing tanstack/query to refetch the data, and then calling `formApi.reset()`. I was seeing a brief flash of the previous data on reset.
> 
> If it helps anyone, calling `setTimeout(() => formApi.reset())` (instead of just `formApi.reset()`) gets rid of the flash. Feels hacky, but it works great. 👍

Confirmed the same issue:
See video.

harry-whorlow:
@santiagazo, Hi thanks for the input!

From what I remember jsefiani's flashing bug was unrelated... 
What version of form is that? did you pull down the pr branch, or the main branch? 

[edit]: out of curiosity are you intentionally handling submission in your validators section?

deshazer:
> @santiagazo, Hi thanks for the input!
> 
> From what I remember jsefiani's flashing bug was unrelated... What version of form is that? did you pull down the pr branch, or the main branch?
> 
> [edit]: out of curiosity are you intentionally handling submission in your validators section?

I was running into the same "flashing" behavior on the main branch. I'm currently running @tanstack/react-form v1.12.2.

I also run form submission in my async validator. It's nice because if the backend returns any errors (potentially related to validation), then I can just return them from the validator. That conveniently merges them into the form errors and they all display on the page just like my frontend zod validation does. It probably wasn't the intention when it was designed, but it does work pretty smoothly as long as your backend can return validation errors similar enough to zod (or can be parsed/converted easily enough). 🤷‍♂️

harry-whorlow:
> I was running into the same "flashing" behaviour on the main branch. I'm currently running @tanstack/react-form v1.12.2.

@santiagazo ah, you're on the main branch. okay, I thought for a second that you had installed the pr branch. 🙂
Have you tried out the pr version?

As a side note, in #1489 we added spreadable error maps so the errors can be spread into the form from the onError of the handle submit. Though, if I remember correctly I think we want to iterate on this.

RaffaeleCanale:
I fear this PR might have broken reactivity in React. When going from `1.14.0` to `1.14.1`, I notice the form state is no longer reactive.

I tried to come up with an example here: https://stackblitz.com/edit/vitejs-vite-o1nrlims

If you look at the code, you'd expect the "Submit" button to change text while it's submitting. But now it no longer works, however if in that stackblitz you try to install `v1.14.0`, then it works.

The component tree no longer re-renders when submitting, which I expect might be due to the removal to this store hook call.

LeCarbonator:
@RaffaeleCanale that's because this was never intended to be reactive in the first place. It just happened to be this way until now.

If you want to receive reactive data from form state, you should use `<form.Subscribe>`. The [simple example](https://tanstack.com/form/latest/docs/framework/react/examples/simple) highlights this.

harry-whorlow:
@RaffaeleCanale brother, I'm two days into my holiday 😭  

Jokes aside, thank you for the report and the sandbox it's really appreciated... 

As @LeCarbonator has mentioned it was kind of an unintended side effect, something like this should solve your issues: 
(as per your example)
```tsx
<form.Subscribe
  selector={(state) => state.isSubmitting}
  children={(isSubmitting) => (
    <button
      type="submit"
      className="text-white bg-blue-700 hover:bg-blue-800 focus:ring-4 focus:outline-none focus:ring-blue-300 font-medium rounded-lg text-sm w-auto px-5 py-2.5 text-center cursor-pointer"
      onClick={() => form.handleSubmit()}
    >
      {isSubmitting ? 'Submitting...' : 'Submit'}
    </button>
  )}
/>
```

Complete [sandbox](https://stackblitz.com/edit/vitejs-vite-dt5jqogg?file=src%2FApp.tsx), if you have any more concerns let me know 😄

RaffaeleCanale:
You are both right, our code was indeed not properly subscribing to the state and we were blissfully benefiting from the previous behavior, so it should be something we can fix on our side.

Thanks for the clear answer and sorry if I caused a fright, hope you'll still enjoy your holidays :smile:

harry-whorlow:
@RaffaeleCanale no worries, glad to be of help! 😄

PR #1495, opened 2025-05-09
fix(form-core): Only allow array deep keys in array methods

LeCarbonator:
This PR restricts array methods on the `FormApi` to only allow paths to array keys.

Consider the following example:
```ts
type FormValues = {
    name: string
    age: number
    startDate: Date
    title: string | null | undefined
    relatives: { name: string }[]
    counts: (number | null | undefined)[]
  }

    name: '',
    age: 0,
    startDate: new Date(),
    title: null,
    relatives: [{ name: '' }],
    counts: [5, null, undefined, 3],
  }

  const form = new FormApi({
    defaultValues,
  })
```
### What's the behaviour on v1.10?
With the setup given above, you can write the following commands:
```ts
  const fieldValue = {}
  // The field value is typed as `never`, so you're informed that it's not correct
  form.pushFieldValue('age', fieldValue as never)
  form.insertFieldValue('age', 0, fieldValue as never)
  form.replaceFieldValue('age', 0, fieldValue as never)
  // No error! It's silently ignored in the implementation
  form.removeFieldValue('age', 0)
  form.swapFieldValues('age', 0, 1)
  form.moveFieldValues('age', 0, 1)
```

### What would change?
```ts
  const fieldValue = {}
  // Error: Argument of type '"age"' is not assignable to parameter of type '"relatives" | "counts"'.ts(2345)
  form.pushFieldValue('age', fieldValue)
  form.insertFieldValue('age', 0, fieldValue)
  form.replaceFieldValue('age', 0, fieldValue)
  form.removeFieldValue('age', 0)
  form.swapFieldValues('age', 0, 1)
  form.moveFieldValues('age', 0, 1)
```
### How does it work?
This implementation first determines `DeepKeys<T>` and then extracts the keys that are of the desired type. Therefore, it should not have a big impact on performance or have problems with future fixes to `DeepKeys<T>`


LeCarbonator:
Not sure why it closed? It claims I did it, so it was unintentional.

Balastrong (approved):
That's brilliant, thank you!

PR #1497, opened 2025-05-09
fix(form-core): fix form.resetField() ignoring nested fields

### LINKED ISSUE #1496: form.resetFieldValue() doesn't work for nested fields
### Describe the bug

When calling `form.resetFieldValue()` on a nested field (`'some.nested.field'`), it will not reset.

The reason because it doesn't acknowledge the nested field name as a field path, and instead treats it as the literal key in the form values. See [this line in `FormApi.ts`](https://github.com/TanStack/form/blob/5565233b21889e5c58bfcf3acbae6b9719767284/packages/form-core/src/FormApi.ts#L2135).

https://stackblitz.com/edit/vitejs-vite-3nazz9pj?file=src%2FApp.tsx

### Steps to reproduce

1. Go to the repro
2. Change the shallow field and try to reset it. This works.
3. Change the nested fied and try to reset it. It won't work.

### Expected behavior

As a user, I expect the field name to be translated into the nested field value as that's what the typing implies.

Every time

### Screenshots or Videos

Here's the outcome of what's actually happening behind the scenes:

Image

### Platform

Any browser

### TanStack Form adapter

None

### TanStack Form version

v1.10.0

### TypeScript version

v5.8.3

### Additional context

_No response_

LeCarbonator:
Closes #1496 

See the issue for more details.

The main issue is that the field name was assumed to be a shallow `keyof TFormData` when it is actually `DeepKeys<TFormData>`. TypeScript didn't complain because there was an `as` assertion.

This PR implements the unit test provided in the issue and fixes it.


packages/form-core/tests/FormApi.spec.ts
```diff
@@ -3580,3 +3580,61 @@ it('should mark sourceMap as undefined when async field error is resolved', asyn
...
+
+  form.resetField('shallow')
+  expect(form.state.values.shallow).toEqual('')
+
+  form.resetField('nested.field.name')
+  expect(form.state.values.nested.field.name).toEqual('')
+})
+
+it('should preserve nested fields on resetField if defaultValues is not provided', () => {
+  const defaultValues = {
```
bpinto, line 3613: @LeCarbonator maybe as with https://github.com/TanStack/form/pull/1469#discussion_r2085379967 it would be a good idea to rename this variable?

harry-whorlow:
Thats a good catch, and good job on the PR.

Do you mind if I throw on a type test?

LeCarbonator:
> Thats a good catch, and good job on the PR.
> 
> Do you mind if I throw on a type test?

Go ahead!

harry-whorlow (approved):

PR #1517, opened 2025-05-18
fix(form-core): Fix  `setBy`/`getBy` not handling arrays properly

### LINKED ISSUE #1488: Setting property of object in nested 2D array changes the nested array to object
### Describe the bug

If you create a form for a 2D array of objects (e.g. `{ value: number }[][]`) and call `field.handleChange` for one of the properties of the objects, the nested array will suddenly turn into an object.

For example, if the current form value is `{ myArr: [[{ value: 1} ]] }` and field `myArr[0][0].value` calls `field.handleChange(42)`, then the new form value will be set to `{ myArr: [{ 0: { value: 1} }] }`.

https://stackblitz.com/edit/tanstack-form-w9x55dvm?file=src%2Findex.tsx

### Steps to reproduce

1. Open the Stackblitz
2. Open the terminal
3. Click the "Add subarray" button
4. Click one of the "Add object" buttons
5. Observe the last console log entry shows the desired form shape (see screenshot below)
6. Change one of the values, e.g. using the number input's arrows
7. Observe that the first new console log entry shows the changed shape (see screenshot below)

### Expected behavior

The shape of the nested arrays should stay the same if some element's (property's) value changes.

Every time

### Screenshots or Videos

Image

### Platform

- OS: macOS
- Browser: Arc (Chromium)

### TanStack Form adapter

react-form

### TanStack Form version

^1.9.1 (see stackblitz)

### TypeScript version

^5.4.2 (see stackblitz)

### Additional context

_No response_

LeCarbonator:
Closes #1488 

During parsing of the field name, multi-dimensional arrays (or names starting with an array access) were wrongly parsed in functions using `makePathArray` internally.

The side effects of this are listed in the linked issue, but to summarize, it wrongly interpreted array access as an object key, and that apparently lead to JavaScript silently converting the array into an object (classic ...)

The reason were mostly the regular expressions used to parse the field name, so test cases have been added and the function has been refactored.

## Before
```tsx
makePathArray('field[0][1].value') 
// == ['field', 0, '1', 'value']
```

## With PR
```tsx
makePathArray('field[0][1].value') 
// == ['field', 0, 1, 'value']
```

There is now a lookahead used to avoid mismatches with subsequent numbers. Benchmarks didn't show a significant drop in performance, but let me know if you notice something suspicious.

### DIFF
--- a/packages/form-core/src/utils.ts
+++ b/packages/form-core/src/utils.ts
@@ -135,11 +135,13 @@ export function deleteBy(obj: any, _path: any) {
   return doDelete(obj)
 }
 
-const reFindNumbers0 = /^(\d*)$/gm
-const reFindNumbers1 = /\.(\d*)\./gm
-const reFindNumbers2 = /^(\d*)\./gm
-const reFindNumbers3 = /\.(\d*$)/gm
-const reFindMultiplePeriods = /\.{2,}/gm
+const reLineOfOnlyDigits = /^(\d+)$/gm
+// the second dot must be in a lookahead or the engine
+// will skip subsequent numbers (like foo.0.1.)
+const reDigitsBetweenDots = /\.(\d+)(?=\.)/gm
+const reStartWithDigitThenDot = /^(\d+)\./gm
+const reDotWithDigitsToEnd = /\.(\d+$)/gm
+const reMultipleDots = /\.{2,}/gm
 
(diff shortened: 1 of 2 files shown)


Balastrong (approved):

bompi88:
LGTM 🚀

PR #1521, opened 2025-05-19
docs(react-form): add type=submit in form-composition.md subscribe button

MrGeniusProgrammer:
the docs does not clearly states that form.handleSubmit must be present in the subscribe button in form composition and will confuse many developers (and I did get confused). So with this it is clearly stated.

### DIFF
--- a/docs/framework/react/guides/form-composition.md
+++ b/docs/framework/react/guides/form-composition.md
@@ -124,7 +124,7 @@ function SubscribeButton({ label }: { label: string }) {
   const form = useFormContext()
   return (
     <form.Subscribe selector={(state) => state.isSubmitting}>
-      {(isSubmitting) => <button onClick={() => form.handleSubmit()} disabled={isSubmitting}>{label}</button>}
+      {(isSubmitting) => <button type="submit" disabled={isSubmitting}>{label}</button>}
     </form.Subscribe>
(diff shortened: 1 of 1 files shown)


docs/framework/react/guides/form-composition.md
```diff
@@ -124,7 +124,7 @@ function SubscribeButton({ label }: { label: string }) {
   const form = useFormContext()
   return (
     <form.Subscribe selector={(state) => state.isSubmitting}>
-      {(isSubmitting) => <button disabled={isSubmitting}>{label}</button>}
+      {(isSubmitting) => <button onClick={() => form.handleSubmit()} disabled={isSubmitting}>{label}</button>}
```
LeCarbonator, line 127: The button is called `SubscribeButton` which doesn't really explain the purpose it's used for in the example (it being a `SubmitButton`)

I think it would be good to rename the ccomponent to be a `SubmitButton` instead.
So since it's meant to be a Submit button based on the label and usage, how about making it a submit button?

```suggestion
      {(isSubmitting) => <button type="submit" disabled={isSubmitting}>{label}</button>}
```

MrGeniusProgrammer: I think it is a good idea

PR #1531, opened 2025-05-28
docs: follow guidelines for reset in simple example

zhouzi:
I was playing with TanStack Form, starting from the simple example. I stumbled upon the annoying bug with `<select>` and form reset and found out that it's clearly documented here:

- https://tanstack.com/form/latest/docs/framework/react/guides/basic-concepts#reset-buttons

So this PR updates the simple example to follow the guidelines regarding form reset (which is to `event.preventDefault()`).


LeCarbonator:
Thanks for the fix!

PR #938, opened 2024-09-05
feat: add isBlurred field state

Pascalmh:
closes #937

PR #1540, opened 2025-05-30
docs: corrected field state definitions

### LINKED ISSUE #1169: "Touched" definition in documentation is wrong
The docs say

> **A field is "touched" when the user clicks/tabs into it**, "pristine" until the user changes value in it, and "dirty" after the value has been changed.

https://tanstack.com/form/latest/docs/framework/react/guides/basic-concepts 

But this is not true. `form.Field` doesn't seem to expose anything that would even *allow* it to be true. It exposes `handleChange` and `handleBlur`. Both of them set the touched flag, but blur only accounts for focus being _lost_, not focus being gained.

I think this is correct:

Image

michaelboyles:
Fixes #1169 

"touched" definition in the docs say it's related to the user clicking the field. There isn't even a handler for clicking/focussing the field, so that's not right

The actual definition of touched is blurred/changed

Also added isBlurred and isDefaultValue states to the description, and further clarified that `isPristine` is just the same as `!isDirty`


docs/framework/angular/guides/basic-concepts.md
```diff
@@ -47,10 +47,19 @@ const {
...
+There are four states in the metadata that can be useful to see how the user interacts with a field:
+
+- _"isTouched"_, after the user changes the field or blurs the field
+- _"isDirty"_, after the field's value has been changed, even if it's been reverted to the default. Opposite of _"isPristine"_
+- _"isBlurred"_, after the field has been blurred (exited)
+- _"isDefaultValue"_, whether the field's current value is the default value
+
+There is also one additional derived state:
+
+- _"isPristine"_, until the user changes the field value. Opposite of _"isPristine"_
```
harry-whorlow, line 59: I'm open for debate on these two points:

1- Do we need to separate out derived state from set state? I'm not entirely - convinced the overwhelming majority of developers are interested that is derived. 

2- isBlurred was introduced for validation purposes in #938, is it used for user feedback? Perhaps it might be cleaner to add it to line 46 and leave it out of this section for clarity.

harry-whorlow (changes requested):
Hi, thanks for the PR! 🙌

Quick question - have you had a chance to check out PR #1456 and the related issues #1080 and [#1081]? I ask because those discussions were the main drivers behind the current layout and structure of the field state documentation. (We had a ton of interaction on those issues, so we wanted to emphasize the reasoning behind our decisions.)

The structure I originally went with was intentional and based on three layers:

General field state – covering value, validation, and errors.
Common interaction states – like isTouched, isDirty, and isPristine, which are most relevant for user interaction.
Introduction of isDefaultValue – added to reflect differences in how isDirty values are handled in React compared to Vue or Angular.
(Though I realize most of this was done in the React docs.)

Would you mind if I pushed a commit to your PR? There are a few things I’d like to tweak, and we can discuss whether they make sense. (I’ll also fix the isPristine type while I’m at it!) 😄

docs/framework/angular/guides/basic-concepts.md
michaelboyles: The reason I separated it was because previously it wasn't very clear that pristine is simply the opposite of dirty. But I did additionally clarify that inline, so perhaps both the inline clarification and the separation is overkill 

To be honest if I was designing the API, I'd get rid of isPristine. It's not hard to negate isDirty. The obvious opposite term for dirty is clean, so the fact it's not simply "clean" makes it sound like it's (clean + something). It sounds like there's an extra concept, when there isn't.

Anyway, it's a breaking change to remove it now, so probably too late.

I added isBlurred here because blurring is tied into touched state, so it feels interrelated to me.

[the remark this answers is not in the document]
michaelboyles: Good spot, thanks

michaelboyles:
Sure, make whichever changes you like. I can fix up too, but probably not til Monday.
 
The most important thing is that clicking the field is nothing to do with touched state. That's just simply wrong.

harry-whorlow:
whats your opinion on this? just updated the docs so all the libraries have the "Understanding 'isDirty' in Different Libraries" segment, and reorganised what you wrote a little.

michaelboyles:
There's a typo on the diagrams - isBlured missing an R

I don't personally see the value in having 2 separate diagrams. They seem identical except for the last row, right? I'd use the "extended" one as the only one, with all 5 properties described above, then elaborate on dirty underneath. But not dying on that hill. At least it's accurate now.

Not sure if it was intentional or not, but for Solid you've changed the syntax from `field().state.meta` to `field.state.meta`. I'm not a Solid person so I don't know which is correct, but looks like it might be a copy-paste typo.

harry-whorlow:
@MichaelBoyles Ah, good catch on the typo!

Yeah, I'm not particularly a fan of the double images, but given that the issues I linked were some of the more interacted-with ones—and people were asking for a clear explanation of our decisions—I don't mind using more space here to emphasise it.

harry-whorlow:
So this pr is looking good to go! 

Thanks for the submission, I'll get around to merging it in a bit. 🚀

PR #1544, opened 2025-05-31
fix: cleanup PR

flensrocker:


examples/angular/large-form/src/app/app.component.ts
```diff
@@ -15,22 +15,20 @@ import type {
...
+      (blur)="api.handleBlur()"
+      (input)="api.handleChange($any($event).target.value)"
+    />
   `,
 })
 export class AppTextField {
-  label = input.required<string>()
-  lastName = injectField<string>()
+  readonly label = input.required<string>()
+  protected readonly field = injectField<string>()
```
crutchcorn, line 31: I generally do not like using `readonly` or `protected` in TS, which are not JS features

examples/angular/large-form/src/app/app.component.ts
```diff
@@ -15,22 +15,20 @@ import type {
   selector: 'app-text-field',
   standalone: true,
   template: `
-    @let api = lastName.api();
-    @if (api) {
```
crutchcorn, line 19: `@if` was needed for type safety in my experience

crutchcorn:
Closing this PR, but only because I'm right about to commit the same changes in a moment :) 

Let's chat more on Bsky? Happy to pair!

flensrocker:
Totally ok with me, I just explore and share!

protected/readonly: yes, it's not really JS, but Angular best practice. I can live without it.

PR #1558, opened 2025-06-05
fix(form-core): fix DeepValue from record values being wrong

### LINKED ISSUE #1551: Records fields have wrong types
### Describe the bug

I wonder if it is possible to have such field value with TanStack Form. Pretty sure it should (and is) but the types seem wrong:
```ts
function App() {
  const form = createForm(() => ({
    defaultValues: {
      firstName: '',
      lastName: '',
      address: {
        home: {
          street: '',
          number: '',
        },
      } as Record<string, { street: string; number: string }>,
    },
    ...
  }));

  return (
    <div>
      <h1>Simple Form Example</h1>
      <form
        onSubmit={(e) => {
          e.preventDefault();
          e.stopPropagation();
          form.handleSubmit();
        }}
      >
        ...
        <div>
          <form.Field
            name={`address.${'home'}.street`}
            children={(field) => (
              <>
                <label for={field().name}>Street:</label>
                <input
                  id={field().name}
                  name={field().name}
                  value={field().state.value}
                  onBlur={field().handleBlur}
                  onInput={(e) => field().handleChange(e.target.value)}
                />
                <FieldInfo field={field()} />
              </>
            )}
          />
        </div>
```
In this example, I specify the field name ```name={`address.${'home'}.street`}``` where `home` is a variable in my real use case.

If you open this Stackblitz repro, you'll notice this does not work as expected.

There is no type issue on the `name` attribute of `form.Field` as it respect one of the expected value (`address.${string}.street`), although the type of the value itself is wrong:
```
(property) value: {
    street: string;
    number: string;
} & string
```
So is the `handleChange` definition.

I've discussed it on Discord and it seems this bug is confirmed.

https://stackblitz.com/edit/ts-form-object?file=src%2Findex.tsx&preset=node

### Steps to reproduce

1. Go to Stackblitz
2. Hover the typescript errors

### Expected behavior

Have the correct type ;)

Every time

### Screenshots or Videos

_No response_

### Platform

macOS, Chrome

### TanStack Form adapter

solid-form

### TanStack Form version

1.6.3

### TypeScript version

_No response_

### Additional context

_No response_

LeCarbonator:
Closes #1551 

This PR fixes a wrong intersection that occurs when iterating over template strings (such as records).
The issue arose because of TypeScript inferring the following:
```ts
type DeepKeysExample = 'records' | `records.${string}` | `records.${string}.name`

type GetValueFrom = 'records.foo.name'
// This matches both `records.${string}.name` as well as `records.${string}`.

// Generated DeepValue: 
// { name: string } & string

// With PR fix:
// string
```

> [!IMPORTANT]
> **TypeScript has a limitation with template string types. Nested records are not yet safe. Example below:**

```ts
type NestedRecord = {
  numberRecord: Record<string, number>
  objectRecord: Record<string, { age: number }>
}

type RecordWrapper = {
  records: Record<string, NestedRecord>
}

type Test1 = DeepValue<RecordWrapper, 'records.foo.numberRecord.bar'>
// Type: NestedRecord & number
// `records.${'foo'}.numberRecord.${'bar'}`
// `records.${'foo.numberRecord.bar'}`

type Test2 = DeepValue<RecordWrapper, 'records.foo.objectRecord.bar'>
// Type: NestedRecord & { age: number }
// `records.${'foo'}.objectRecord.${'bar'}`
// `records.${'foo.objectRecord.bar'}`

type Test3 = DeepValue<RecordWrapper, 'records.foo.objectRecord.bar.age'>
// Type: number
// `records.${'foo'}.objectRecord.${'bar'}.age`
```


binajmen:
thank you for fixing this! is there a timeline on when it will be included in a release? 👀

LeCarbonator:
@binajmen Well, the main issue is that this isn't really a fix. This PR has been a bit stale because of that ... but I'm not sure there will be a convenient solution that doesn't need reworking all of the DeepKeys types

LeCarbonator:
I made progress! It now shows possible values as union instead of an intersection. Will commit the changes tomorrow.

crutchcorn:
> it's green! what is blocking the merge?!

Please don't.

Issue #1559, opened 2025-06-06
form.moveFieldValues mutates the default values


tmarnet:
### Describe the bug

Hello 👋 

In the case of a field in `array` mode, it would appear like calling `field.moveValue` (and the underlying `form.moveFieldValues`) mutates the form's default values for that field array, thus not computing this field's `isDefaultValue` correctly

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-form-tgsosw1h

### Steps to reproduce

1. Click either item's move up/move down button
2. Notice how the form's "isDefaultValue" is still `true` in the debug view
3. Notice how the form's "default items" have changed and do not represent the actual default values

### Expected behavior

After moving either item up/down, I would expect the form's default values to not have changed and the form's `isDefaultValue` state to be `false`

### How often does this bug happen?

Every time

### Platform

- OS: macOS
- Browser: Chrome
- Version: 137.0.7151.56

### TanStack Form adapter

None

### TanStack Form version

1.11.1

LeCarbonator:
Fixed in `v1.12.2`

PR #1560, opened 2025-06-06
fix(form-core): prevent `FormApi#moveFieldValues` from mutating `defaultValues`

tmarnet:
This PR addesses the following issue: https://github.com/TanStack/form/issues/1559

💡 Open question: I don't know if other field array methods suffer from the same issue but, if some do, should we address those in this PR too?

### DIFF
--- a/packages/form-core/tests/FormApi.spec.ts
+++ b/packages/form-core/tests/FormApi.spec.ts
@@ -812,6 +812,11 @@ describe('form api', () => {
     form.mount()
     form.moveFieldValues('names', 1, 2)
 
+    expect(form.options.defaultValues?.names).toStrictEqual([
+      'one',
+      'two',
+      'three',
+    ])
     expect(form.getFieldValue('names')).toStrictEqual(['one', 'three', 'two'])
   })
 
--- a/packages/form-core/tests/FormApi.spec.ts
+++ b/packages/form-core/tests/FormApi.spec.ts
@@ -1016,6 +1016,10 @@ describe('form api', () => {
     form.reset()
     form.moveFieldValues('names', 1, 2)
     expect(form.options.defaultValues?.names).toStrictEqual(defaultValues.names)
+
+    form.reset()
+    form.clearFieldValues('names')
+    expect(form.options.defaultValues?.names).toStrictEqual(defaultValues.names)
   })
 
   it('should handle fields inside an array', async () => {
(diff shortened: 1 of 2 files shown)


packages/form-core/tests/FormApi.spec.ts
```diff
@@ -812,6 +812,11 @@ describe('form api', () => {
     form.mount()
     form.moveFieldValues('names', 1, 2)
 
+    expect(form.options.defaultValues?.names).toStrictEqual([
+      'one',
+      'two',
+      'three',
+    ])
```
tmarnet, line 819: This wouldn't pass before

LeCarbonator: This unit test is about the meta being correct. Since the added lines are about defaultValues remaining the same, could you create a copy of this unit test that always checks this condition after every array call?

LeCarbonator (commented):
Good catch! Just a nitpick with the unit test, but the fix itself looks  good to me!

packages/form-core/tests/FormApi.spec.ts
tmarnet: Sure thing!

I'm not sure what you mean exactly by "after every array call" tho, would you mind expanding on that please?

LeCarbonator: The unit test should call all array manipulation methods and then check if the defaultValues were modified. Basically a copy of this unit test, but instead of checking the created values, checking the defaultValues.

tmarnet: Gotcha! Do you prefer one test for all array methods or one test for each? 🙂

LeCarbonator: One test will do! If new array methods appear in the future, they can append it to that test.

tmarnet: Done! Let me know what you think.

LeCarbonator: Looks good! Could you give it a try in your stackblitz to see if it's fixed the issue?

```
npm i https://pkg.pr.new/@tanstack/react-form@1560
```

LeCarbonator: Sorry, the workflow hasn't run. It didn't upload the PR yet. It should be available now.

tmarnet: Just tried it, kooks like it fixed the issue, thanks!

LeCarbonator:
Thanks for the fix!

PR #1562, opened 2025-06-07
feat(form-core): Async field onChange with submition handling

tomrehnstrom:
Hello!

Async field level `onChange` handlers was one of the features that made me want to switch to `Tanstack/form` but I ran into a problem where I wanted to modify the value of the field using a debounced async function on the `onChange` but when submitting the form I want to ensure that the async `onChange` has been correctly ran before allowing submission/validation.

A concrete example is in a dynamic form I have users can add image upload fields, that I want to begin uploading to s3 `onChange`, but they should not be allowed to submit unless the upload function has been completed successfully and pressing submit should begin loading until all the actions/uploads are completed.

I have thrown together a working, but maybe not perfect implementation, and modified the simple example to allow playing around with it. I would be happy to finish the feature if others also agree this would be a good feature :)

### DIFF
--- a/examples/react/simple/src/index.tsx
+++ b/examples/react/simple/src/index.tsx
@@ -22,7 +22,7 @@ export default function App() {
     },
     onSubmit: async ({ value }) => {
       // Do something with form data
-      console.log(value)
+      console.log(value, 'Submitted')
     },
   })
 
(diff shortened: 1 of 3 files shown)


LeCarbonator:
Hmmm ... listeners currently don't cause any side effects, so this change seems conflicting with that.

If your use case is file uploads and waiting for them to finish, can't you disable the submit button while that upload is pending (instead of refactoring listeners to affect form state?)

@tomrehnstrom

tomrehnstrom:
@LeCarbonator Yeah I think I overengineered this solution a bit, I got it working in another way!

PR #1566, opened 2025-06-08
Docs: Add error handling section to submission handling docs

LeCarbonator:
- [x] React
- [x] Solid
- [x] Vue
- [ ] Angular

This PR adds some code snippets to get users started on the three common types of errors:
- Sync user errors
- Async user errors
- Server errors

The user errors are implemented using form / field validators (with links to the validation guide page). The Server error example is implemented using TanStack Query.

### DIFF
--- a/docs/framework/react/guides/submission-handling.md
+++ b/docs/framework/react/guides/submission-handling.md
@@ -73,10 +73,10 @@ const createAccountMutation = useMutation({
 
 const form = useForm({
   // ...
-  onSubmit: ({ value }) => {
+  onSubmit: async ({ value }) => {
     // If an error happens, they are accessible through
     // `createAccountMutation.error`
-    return createAccountMutation.mutateAsync(value)
+    await createAccountMutation.mutateAsync(value)
   },
 })
 ```
--- a/docs/framework/react/guides/submission-handling.md
+++ b/docs/framework/react/guides/submission-handling.md
@@ -22,7 +22,6 @@ These are validation issues that can be detected immediately on the client side,
 <form.Field
   name="age"
   validators={{
-    // This lets you catch simple issues early and prevent invalid data from being submitted.
     onChange: ({ value }) =>
       value < 13 ? 'You must be 13 to make an account' : undefined,
   }}
--- a/docs/framework/react/guides/submission-handling.md
+++ b/docs/framework/react/guides/submission-handling.md
@@ -3,6 +3,84 @@ id: submission-handling
 title: Submission handling
 ---
 
+## Error handling
+
+When a user submits a form, there are several ways errors can occur — each requiring its own handling approach. TanStack Form gives you the flexibility to handle errors your way. To help you get started, here’s how to approach the three most common types of errors you’ll run into:
+
+- **Synchronous user errors** – Things you can catch right away, like a field being left empty or text that's too short.
+- **Asynchronous user errors** – Issues that come up after submission, like finding out a username is already taken.
+- **Server errors** – Unexpected problems on the backend, such as a failed request or an unknown error.
+
+> [!TIP]
+> If you need help understanding how validation works in TanStack Form, be sure to check out the [form validation](./validation.md) guide!
+
+### Synchronous user errors
+
+These are validation issues that can be detected immediately on the client side, even before the form is submitted. They are typically managed using both form-level and field-level validators, as shown in the example below:
+
+```tsx
+<form.Field
+  name="age"
+  validators={{
+    // This lets you catch simple issues early and prevent invalid data from being submitted.
+    onChange: ({ value }) =>
+      value < 13 ? 'You must be 13 to make an account' : undefined,
+  }}
+  children={{() => <>{/* ... */}</>}}
+/>
+```
+
+### Asynchronous user errors
+
+Asynchronous errors usually occur after the form is submitted, often due to external checks — like verifying whether a username or email is already taken. These kinds of errors can be handled using the async variants of form and field validators.
+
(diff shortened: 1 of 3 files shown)


docs/framework/react/guides/submission-handling.md
```diff
@@ -3,6 +3,83 @@ id: submission-handling
 title: Submission handling
 ---
 
+## Error handling
+
+When a user submits a form, there are several ways errors can occur — each requiring its own handling approach. TanStack Form gives you the flexibility to handle errors your way. To help you get started, here’s how to approach the three most common types of errors you’ll run into:
+
+- **Synchronous user errors** – Things you can catch right away, like a field being left empty or text that's too short.
```
LeCarbonator:
```suggestion
- **Synchronous user errors** – Issues you can catch right away, like a field being left empty or text that's too short.
```

Needs to also be fixed for the other files.

docs/framework/react/guides/submission-handling.md
```diff
@@ -3,6 +3,83 @@ id: submission-handling
...
+const form = useForm({
+  // ...
+  validators: {
+    onSubmitAsync: async ({ value }) => {
+      // Validate the value on the server
+      const response = await createAccount(value)
+
+      if (response.isError) {
+        // Username is taken, return an error for the username field
+        return {
```
LeCarbonator: A `form` error is good enough for this, no need to spread to fields

docs/framework/react/guides/submission-handling.md
```diff
@@ -3,6 +3,79 @@ id: submission-handling
...
+  // ...
+  validators: {
+    onSubmitAsync: async ({ value }) => {
+      // Validate the value on the server
+      const response = await createAccount(value)
+
+      if (response.isError) {
+        // Username is taken, return an error for the username field
+        return 'Username is already taken!'
+      }
```
Pascalmh, line 49: This says 

> return an error for the username _**field**_

but the error seems to go to the **Form** not to the **Field**

docs/framework/react/guides/submission-handling.md
```diff
@@ -3,6 +3,79 @@ id: submission-handling
...
+const createAccountMutation = useMutation({
+  /* ... */
+})
+
+const form = useForm({
+  // ...
+  onSubmit: async ({ value }) => {
+    // If an error happens, they are accessible through
+    // `createAccountMutation.error`
+    await createAccountMutation.mutateAsync(value)
```
Pascalmh, line 74: Promoting "mutateAsync" here feels off:

> you should almost always use `mutate`
> – https://tkdodo.eu/blog/mastering-mutations-in-react-query#mutate-or-mutateasync

```suggestion
    createAccountMutation.mutate(value, {
      onSuccess: () => {
        // handle success by navigate to another route e.g.
      }
    })
```

LeCarbonator: Here's the thought process behind it:

* Is `mutate` synchronous? If not, `onSubmit` would finish immediately and `isSubmitting` would flash `true` for a moment
* Error handling informs TSF about `isSubmitSuccessful`, which would cause a disconnect if `mutate` errors

docs/framework/react/guides/submission-handling.md
LeCarbonator: Good catch! The comment is kind of redundant now. I'll remove it!

docs/framework/react/guides/submission-handling.md
Pascalmh: Then this seems like one of those exceptions to not prefere `mutate` over `mutateAsync` ;-)

It's perfect as it is, thanks for clarifying!

LeCarbonator: Do you reckon this should be mentioned somewhere? I assume there's lots of users combining Tanstack Form with TansStack Query

Pascalmh: There's some explenation here: https://tanstack.com/form/latest/docs/reference/type-aliases/baseformstate#issubmitting

I think it would be good to mention it on the "Submission Handling"-Page as well as it clears up a lot of questions.

docs/framework/react/guides/submission-handling.md
```diff
@@ -3,6 +3,79 @@ id: submission-handling
...
+
+```tsx
+<form.Field
+  name="age"
+  validators={{
+    onChange: ({ value }) =>
+      value < 13 ? 'You must be 13 to make an account' : undefined,
+  }}
+  children={{() => <>{/* ... */}</>}}
+/>
```
Pascalmh, line 29: IIRC @crutchcorn wanted to favor Form-Level-Validation over Field-Level-Validation, switching to Form-Level here would keep this guide on Form-Level throughout (reduced complexity).

Another Idea I want to pitch where this would help is: The "Error handling" should end with a "Putting it all together" chapter that shows a full `useForm` example with `onChange`, `onSubmitAsync` validators and the `onSubmit`.

[Ayush pointed out](https://discord.com/channels/719702312431386674/1383130262287941789/1383140872266186873) that it wasn't clear enough yet.
I think he's asking why `createAccount` is called once in `form.validators.onSubmitAsync` and then again in `form.onsubmit` (ony that it's now using TSQ and is called `createAccountMutation.mutateAsync`)

LeCarbonator: Good idea! Then it could represent three use cases that don't conflict name-wise!

LeCarbonator:
This has sparked some discussion internally. Since it's not clear how big the changes to the diffs will be, I'll mark it as draft for now.

Issue #1565, opened 2025-06-08
Clarify name prop usage for inputs and form controls when using Next.js Server Actions


jamestrenda:
I’ve been integrating @tanstack/react-form with a Next.js Server Action and ran into an issue where the formData was coming in as `null` on the server side.

Turns out, it was because I didn’t explicitly set the `name` attribute on the actual HTML `<input />` elements—only on the <AppField /> wrapper component. This tripped me up because the [form composition example](https://tanstack.com/form/latest/docs/framework/react/guides/form-composition#pre-bound-field-components) doesn’t include the name prop on the input field itself, while the [Next.js Server Actions example](https://tanstack.com/form/latest/docs/framework/react/examples/next-server-actions?path=examples%2Freact%2Fnext-server-actions%2Fsrc%2Fapp%2Fclient-component.tsx) does include it.

Here’s a minimal illustration of what was missing:

```ts
<form.AppField name="firstName">
  {(field) => <field.TextField label="First Name" />}
</form.AppField>
```

```ts
// text-field.tsx
import { useFieldContext } from './form-context.tsx'

export function TextField({ label }: { label: string }) {
  const field = useFieldContext<string>()
  return (
    <label>
      <div>{label}</div>
      <input
        name={field.name} // ✅ this is required for formData to work server-side
        // other props...
      />
    </label>
  )
}
```

In contrast, the form composition docs omit `name={field.name}` in the input, which led me to assume it was unnecessary or handled for us automatically:

```ts
// From docs (missing HTML name):
<input
  value={field.state.value}
  onChange={(e) => field.handleChange(e.target.value)}
  // name is not included here
/>
```

I totally understand this is just how HTML forms work — inputs need a name to be serialized in the POST request — but since form libraries often abstract away HTML standards, it wasn’t immediately obvious that this step was still required.

**Suggestion:**
It would be really helpful to add a small comment or note to the form composition guide or the Next.js Server Action guide that says something like:

💡 When using native form submission (e.g. with Next.js Server Actions), be sure to include the `name` attribute on your actual input fields, not just the` <form.AppField name="..." />` wrapper.

I'd be happy to submit a PR. Just let me know what section of the docs you think this belongs in. Maybe _both_?

Thanks for all the hard work on TanStack Form and for the prompt communication on the Discord server — I'm really excited about this library!

LeCarbonator:
I think there are two places where it should be mentioned:

[...]

PR #1574, opened 2025-06-10
docs(nextjs-ssr): clarify need for HTML name attribute in form submissions

jamestrenda:
add inline comments to client components in next.js server action examples highlighting the need to explicitly set the name attribute on input fields for native POST submissions (e.g. with Next.js Server Actions). This avoids confusion where inputs may otherwise be omitted from formData.

closes https://github.com/TanStack/form/issues/1565

### DIFF
--- a/docs/framework/react/guides/ssr.md
+++ b/docs/framework/react/guides/ssr.md
@@ -293,7 +293,7 @@ export const ClientComp = () => {
           return (
             <div>
               <input
-                name="age"
+                name="age" // must explicitly set the name attribute for the POST request
                 type="number"
                 value={field.state.value}
                 onChange={(e) => field.handleChange(e.target.valueAsNumber)}
--- a/examples/react/next-server-actions/src/app/client-component.tsx
+++ b/examples/react/next-server-actions/src/app/client-component.tsx
@@ -37,7 +37,7 @@ export const ClientComp = () => {
           return (
             <div>
               <input
-                name="age"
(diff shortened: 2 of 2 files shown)


docs/framework/react/guides/ssr.md
```diff
@@ -293,7 +293,7 @@ export const ClientComp = () => {
           return (
             <div>
               <input
-                name="age"
+                name="age" // must explicitly set the name attribute for the POST request
```
LeCarbonator: Would it make sense to use `name={field.name}` instead of a new string? Or is there something users should know about what names to choose for the JSX elements?

LeCarbonator (commented):
Looks good to me! There's just one question, but this is fine to merge IMO.

docs/framework/react/guides/ssr.md
jamestrenda: Yes, you're absolutely right (I feel like an AI assistant with that response 😄)! 

Since we have access to the `field` object on the render prop, we should definitely use `name={field.name}` rather than hard-coding the `name` attribute. That's what I'm doing in my own project in the component that uses `useFieldContext`, but I forgot we have access to the `field` object here as well. Good catch!

LeCarbonator: Sounds good, then! Could you check the whole file for the `name` prop? Once that's committed, I'll merge. The comment is fine for just the first instance of it imo.

PR #1584, opened 2025-06-18
chore: devinxi start example

Balastrong:
Update start to the devinxi version


SutuSebastian:
1. is it normal behaviour after form submit to display the raw response and not return it into state?

https://github.com/user-attachments/assets/2f429b5b-52a5-400d-8979-da255a37c098

2. If we update the packages we get the following error:

Screenshot 2025-06-19 at 11 46 54

Screenshot 2025-06-19 at 11 47 26

CC @tannerlinsley @crutchcorn

kusiewicz:
@SutuSebastian 

> 2. If we update the packages we get the following error:

https://github.com/TanStack/router/issues/4409

Balastrong:
@SutuSebastian 
> 1. is it normal behaviour after form submit to display the raw response and not return it into state?

In the current implementation, yes. I'm only updating Start to the latest version here but this can indeed be improved later :)
Currently Form doesn't have a field in the state to keep a success or response message but maybe this can be introduced for SSR adapters... or the response could flow to onSubmit... or a new onServerSubmit... just thinking out loud at this point :D

I'd say ideas are always welcome!

PR #1593, opened 2025-06-25
fix(form-core): prevent `onBlur` and `onChange` form listeners from canceling each other

dusan233:
When both `onBlur` and `onChange` form listeners are provided with their own debounce delays (`onBlurDebounceMs`, `onChangeDebounceMs`), triggering one event shortly after the other results in only the last listener running. 

Example: [https://codesandbox.io/p/sandbox/zxyv5k](https://codesandbox.io/p/sandbox/zxyv5k)

Ways to reproduce:

1. Type something into input, then click outside of input to trigger blur event (only blur listener callback gets called)
2. Click into input, then out of it, then type something into input (only change listener callback gets called)

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -1770,11 +1770,11 @@ export class FieldApi<
   private triggerOnChangeListener() {
     const formDebounceMs = this.form.options.listeners?.onChangeDebounceMs
     if (formDebounceMs && formDebounceMs > 0) {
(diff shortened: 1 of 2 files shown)


LeCarbonator:
LGTM! However, could you add a unit test to make sure we don't regress on this? thanks.

harry-whorlow:
LGTM 🚀, and nice catch.... I think I was responsible for that slip up, so thank you for fixing it :)

PR #1596, opened 2025-06-28
Create useExternalFormApi hook

matthewdavi:
This PR addresses an issue I raised here, https://github.com/TanStack/form/discussions/1595. 

The new `useExternalFormApi` hook would make working with forms more ergonomic, as they are already external stores and there are many usecases for reading / writing to a store from outside of the react lifecycle.

### DIFF
--- a/packages/react-form/src/index.ts
+++ b/packages/react-form/src/index.ts
@@ -2,7 +2,7 @@ export * from '@tanstack/form-core'
 
 export { useStore } from '@tanstack/react-store'
 
-export type { ReactFormApi, ReactFormExtendedApi } from './useForm'
+export type { ReactFormApi, ReactFormExtendedApi } from './useExternalFormApi'
 export { useForm } from './useForm'
 
 export type { UseField, FieldComponent } from './useField'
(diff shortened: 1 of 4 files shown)


crutchcorn:
This sounds like a terrible way to introduce memory leaks into your app. Can you provide programmatic usage for why we'd want to add this?

matthewdavi:
> This sounds like a terrible way to introduce memory leaks into your app. Can you provide programmatic usage for why we'd want to add this?

Yes actually, i've been chatting on Discord with @LeCarbonator about this and I've created a sample repo. Can I add you to it?

crutchcorn:
Sure! I'm hesitant, because it sounds a bit like going against React best conventions, but happy to take a look at least.

matthewdavi:
> Sure! I'm hesitant, because it sounds a bit like going against React best conventions, but happy to take a look at least.

I've added you. This seems fairly natural to me since it follows a similar headless pattern as Tanstack Store, and even Tanstack Query, which create their queryClients and stores at module level, which allows for much better ergonomics, being able to read / update values without being tied to the React lifecycle and passing props, or being within a specific context provider.

crutchcorn:
Per our convos privately, closing.

PR #1597, opened 2025-06-29
feat(solid-form): add `createFormHook` for solid-js.

MAST1999:
I was starting a new project and decided to use solid, and I found out that `createHookForm` was missing so here I am :smile:

I don't have too much experience with solid, but I think this looks good.


MAST1999:
Will start the work on docs after getting the green light on the PR.

LeCarbonator:
The feature's already in the works over at #1453 . Would you mind taking a look at it and giving feedback?

MAST1999:
> The feature's already in the works over at #1453 . Would you mind taking a look at it and giving feedback?

Looks like it has the same changes I've made, I've just addressed the comments there and used the types that were defined in the React implementation, I guess.

We can change the target branch from main to that PR if we want to track it there.

MAST1999:
I'm using this PR in my project and fixing problems for now.

crutchcorn:
@MAST1999 you are AWESOME. Thank you for taking this on! What can we do to support?

MAST1999:
> @MAST1999 you are AWESOME. Thank you for taking this on! What can we do to support?

Hey

I think it's ready, I'm using it in my hobby project and it's working.

But could always use a review or more testing.

crutchcorn:
Awesome! In that case do you wanna start writing some docs for it? Even if it's just copy+pasted from the React one with Solid.js syntax?

MAST1999:
Sounds good

Will get started!

crutchcorn:
@MAST1999 I saw that you added docs, are we ready for final review and/or merge or are we waiting on something else? :)

MAST1999:
> @MAST1999 I saw that you added docs, are we ready for final review and/or merge or are we waiting on something else? :) 

It's ready for final review.

I wanted to add examples as well but I noticed that we use released versions for examples, So I'll open another PR with examples after release.

crutchcorn:
Oh, actually, that's a great callout. Can we build out a `large-form` example before merging? It'll help me quickly validate that it's working as expected on my end as well.

We don't need to use the released version, `pnpm` will find the local version of the package and link to it. Then, when we release automatically it'll auto-bump this for us.

Completely safe.

crutchcorn:
Sorry for the continual last-minute requests 😅

MAST1999:
Don't worry about it.

This is what I wanted to do anyway 😁

MAST1999:
OK, found and interesting problem that I hadn't run into. It happens when I try using the store on the `field`.

I've added a comment on the `text-field` explaining the problem. You can reproduce it easily.

Looked around, but I couldn't figure out where the problem is exactly. If you access the fields itself for value and errors it works correctly now, if you try using the `field().store` it breaks.

crutchcorn:
Uhm... Wow.... Our Solid Store adapter is _real_ broken. I'm a bit surprised we didn't run into problems before. Let me try to fix that

crutchcorn:
Should be fixed now! Launching as soon as the CI passes! Thanks a million once again!

crutchcorn:
Oh dear... I've seem to have merged too early. Much of the composition's docs are still in React. @MAST1999 could you update those for us when you get a chance? :)

MAST1999:
> Oh dear... I've seem to have merged too early. Much of the composition's docs are still in React. @MAST1999 could you update those for us when you get a chance? :) 

Yep, will send a PR soon

Issue #1506, opened 2025-05-12
expecting type error when using a field component on a form field with the wrong type


realugi:
### Describe the bug

In the minimal, reproducible example, I am defining two properties for the form:
1. a string field `stringField`
2. a number field `numberField`

And I am also creating two field components:
1. TextField, which expects a string field/property via `useFieldContext<string>`
2. NumberField, which expects a number field/property via `useFieldContext<number>`

First, I am using the field components right by assigning the `stringField` to TextField and `numberField` to NumberField.
Then, I am using the field components wrong by providing the TextField with the `numberField` and the NumberField with `stringField`.

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-form-thxaofhg

### Steps to reproduce

1. Fill in the first two right fields
2. Click Submit
3. See the rightly typed form values
4. Then fill in the last two wrong fields
5. Click Submit
6. See the wrongly typed form values

### Expected behavior

As a user, I expect a type error, but I get no type error.

### How often does this bug happen?

Every time

### Platform

- OS: Linux (NixOS)
- Browser: Firefox
- Version: 138.0

### TanStack Form adapter

react-form

### TanStack Form version

v1.11.1

### TypeScript version

v5.8.3

### Additional context

I do not have a stackblitz account, so the link might be invalid by the time you click it.
So I downloaded the stackblitz example and put it in a github repository, just in case:
https://github.com/realugi/tanstack-form-minimal-reproducible-example

I created the stackblitz example by forking and modifying your large-form example.

Also this might be some of a nitpick, since everything else is so well typed and this is just a small detail. But still, I believe that you people strive for excellence and that this bug report might be useful to you.

Pascalmh:
@realugi your Stackblitz link is this: https://stackblitz.com/github/realugi/tanstack-form-minimal-reproducible-example (see https://developer.stackblitz.com/guides/integration/open-from-github for more info)

realugi:
thank you very much, I did not know about that

crutchcorn:
The suggested fix to this problem is:

```
  <form.AppField name="number">
    {(field) => (
      <field.String value={field.state.value} />
    )}
  </form.AppField>
```

Even if `field.state.value` goes unused in the `String` component's runtime.

That way the `String` component is defined as:

```
function String(props: {value: string}) {}
```

PR #1606, opened 2025-07-08
feat: Allow restricting field component to field value

LeCarbonator:
Related issue: #1506

This PR implements a way to create 'branded' field components. Those special field components may only be used if the `AppField`'s value matches it.

TODOS
- [ ] Can it be generic somehow? Can the component access that?
- [ ] Check need for implementation (and implement)
  - [x] React
  - [ ] Solid
  - [ ] Angular
- [ ] Amend documentation of Form Composition
  - [ ] React
  - [ ] Solid
  - [ ] Angular
- [ ] Unit tests

# The issue
```tsx
interface TextFieldProps {
  label: string
}

function TextField(props: TextFieldProps) {
  // we assume our field will be of type string
  const field = useFieldContext<string>();

  return <></>;
}

const { useAppForm } = createFormHook({
  fieldContext,
  formContext,
  fieldComponents: {
    TextField
  },
  formComponents: {},
})

function App() {
  const form = useAppForm({
    defaultValues: {
      a: '',
      b: 0,
    },
  })
 

  return (
    <>
      <form.AppField name="a">
        {(field) => {
          // okay, because 'a' is a string
          return <field.TextField label="" />;
        }}
      </form.AppField>
      <form.AppField name="b">
        {(field) => {
          // oops! 'b' is a number, not a string. No error shows up
          return <field.TextField label="" />;
        }}
      </form.AppField>
    </>
  )
}
```
# The solution
The PR addresses it as follows (feedback is appreciated!):
```tsx
interface TextFieldProps {
  label: string
}

function TextField(props: TextFieldProps) {
  // we assume our field will be of type string
  const field = useFieldContext<string>();

  return <></>;
}

// we're creating a field component restricted to strings, with the properties of TextFieldProps
const TextFieldComponent = createFieldComponent<string, TextFieldProps>(TextField);

const { useAppForm } = createFormHook({
  fieldContext,
  formContext,
  fieldComponents: {
    TextField: TextFieldComponent,
    OldTextField: TextField
  },
  formComponents: {},
})

function App() {
  const form = useAppForm({
    defaultValues: {
      a: '',
      b: 0,
    },
  })
 

  return (
    <>
      <form.AppField name="a">
        {(field) => (
          <>
             {/* Okay, because 'a' is a string */}
             <field.TextField label="" />
             <field.OldTextField label="" />
          </>
        )}
      </form.AppField>
      <form.AppField name="b">
        {(field) => (
          <>
            {/* Error! TextField cannot be used on number fields */}
            <field.TextField label="" />
             {/* Directly passed compoents are considered 'global' */}
             <field.OldTextField />
          </>
        }}
      </form.AppField>
    </>
  )
}
```


LeCarbonator:
Wow. You'd think if one misclick is able to mark a PR as ready for review, you'd have the option to mark it as draft on mobile.

Doesn't seem to be possible. I'll mark it as draft again when I'm on desktop ...

adamkangas:
The current implementation works great for simple types, and I could certainly make use of it for my most common form fields to cut down on some boilerplate and current need to pass `field` or `value` into the component for "full" type safety.

What were you wondering about specifically in terms of generics?

I've had a play with trying to register a generic form components similar to `<RecordSelect />` which has a value of `TValue  | null` (where `TValue extends { id: string; }`)

After registering the component, it's fairly straightforward to define various simple data shapes that extend `{id: string;}` and store them in form fields. These easily qualify for the generic `field.RecordSelect` because they satisfy `AppFieldComponentsOfType`.  

However, the component ends up locked to a strict option type of `{ id: string; }`

I'm no library-level TS wizard so I'm a bit grasping at straws for we'd ever be able to inject the correct type values into these generic components at the `AppField` level, where we know the concrete `TData` we're dealing with. Perhaps there's an approach where we're registering functions conforming to a certain signature which in turn return components, instead of registering the components themselves? (This is more "can we do it?" than "should we do it?" thinking)

LeCarbonator:
You mean a case where the component has properties that should adapt to the field value using it?

Yeah, that is very difficult in the current iteration of this PR, which is why it has been stale while I consider some other options.

A simple example of this would also be Radio buttons. The `value` of the radio button should not be `string`, but restricted to the field value using it (Example: `'a' | 'b'`). A generic component solves this very easily, but this implementation cannot be compatible with it.

adamkangas:
Yes, Radio and a basic Select (constrained to a string union) are simpler examples of the challenges here.

This crude API seems to work in my application code, the challenge would obviously be to feed `createGenericFoo` functions into `createFormHook()` and get properly-scoped `GenericFoo` components outputted on `AppField`

```
function makeCreateGenericFieldComponent<
  TBaseFieldValue, // For use in AppFieldComponentsOfType or an equivalent helper
  TGetScopedComponent extends <T extends TBaseFieldValue>(value: T) => React.FC<any>, // Allows us to get the scoped component once we know the field value
>(args: { baseFieldValue: TBaseFieldValue; getScopedComponent: TGetScopedComponent }) {
  return args
}

// We would want to somehow pass this into createFormHook
const createGenericSelect = makeCreateGenericFieldComponent({
  baseFieldValue: '' as string, 
  getScopedComponent: <TValue extends string>(value: TValue) => {
    return Select<TValue>
  },
})

// We would want something like this to happen inside AppField once we have access to the field value
const myAnimal = "dog" as "dog" | "cat" | "bird";
const AnimalSelect = createGenericSelect.getScopedComponent(myAnimal);
//     ^ yay correctly inferred as Select<"dog" | "cat" | "bird">
```

For this UI element:

```
type SelectFieldProps<TValue extends string> = {
  label: string
  options: readonly TValue[]
  getOptionLabel?: (option: TValue) => string
}

function Select<TValue extends string>({
  label,
  options,
  getOptionLabel,
}: SelectFieldProps<TValue>) {
  const field = useFieldContext<TValue>()
  return (
    <label>
      <div>{label}</div>
      <select
        value={field.state.value}
        onChange={(e) => {
          const option = options.find((o) => o === e.target.value)
          if (!option) throw new Error('Invariant')
          field.handleChange(option)
        }}
      >
        {options.map((option, index) => (
          <option key={index} value={option}>
            {getOptionLabel ? getOptionLabel(option) : option}
          </option>
        ))}
      </select>
    </label>
  )
}
```

LeCarbonator:
Sounds interesting! After some other PR work, I'll freshen this one up a bit and see if I can find a nice solution for it.

kepta:
Have you looked at this approach - https://github.com/TanStack/form/discussions/1240 , I feel using this approach is a much simpler option than to create this type mapping

LeCarbonator:
It is much simpler, but as you saw, the `field` assignment would only serve for type inference and is not a prop that is passed at all. That's why this PR tried to check out alternative approaches.

PR #1607, opened 2025-07-08
docs(solid-form): update examples from react to solid.

MAST1999:
I missed a few references in solid's form composition guide.

This fixes those.


MAST1999:
Tell me if there's anywhere else that needs updating.

crutchcorn (approved):

PR #1622, opened 2025-07-22
Add alternative validation methods

crutchcorn:
This PR adds in a new (optional) property called `validationLogic` that allows us to emulate alternative validation methods, like the one from RHF where validation does not occur until the first submission, then it validates on change:

```tsx
const form = useForm({
    defaultValues: {
        name: '',
    },
    // These are the defaults, but demonstrate how you can change them
    validationLogic: revalidateLogic({mode: "submit", modeAfterSubmission: "change"}),
    validators: {
        onDynamic: z.object({
            name: z.string().min(3, 'Name must be at least 3 characters long'),
        }),
    }
})
```

> The above is React, but will be supported day one in all supported framework adapters.

This is a draft pull request and will not be landing immediately, as it is a substantial internal refactor of our code. Once code is done, we'll need docs for all frameworks

# TODO

- [x] Fix types on core
- [ ] Make types on validator functions more strict
- [x] Add RHF props for `reValidateMode` and `mode`
- [x] Fix types for React
- [x] Fix types for Angular
- [x] Fix types for Vue
- [x] Fix types for Lit
- [x] Fix types for Solid
- [x] Fix types for Svelte
- [x] Write docs


packages/form-core/src/ValidationLogic.ts
```diff
@@ -0,0 +1,199 @@
...
+    any,
+    any,
+    any,
+    any
+  >[keyof FormValidators<any, any, any, any, any, any, any, any, any, any>]
+  cause: 'change' | 'blur' | 'submit' | 'mount' | 'server' | 'dynamic'
+}
+
+export interface ValidationLogicProps {
+  // TODO: Type this properly
```
LeCarbonator, line 21: Is this TODO still up to date?

packages/form-core/src/ValidationLogic.ts
```diff
@@ -0,0 +1,199 @@
...
+  form: AnyFormApi
+  // TODO: Type this properly
```
LeCarbonator, line 23: Pending TODO

packages/form-core/src/ValidationLogic.ts
```diff
@@ -0,0 +1,199 @@
...
+        form: props.form,
+      })
+    }
+    case 'change': {
+      // Run change, server validation
+      return props.runValidation({
+        validators: [onChangeValidator, onServerValidator],
+        form: props.form,
+      })
+    }
```
LeCarbonator, line 195: Have we always run `onServer` for onChange / onBlur?

packages/form-core/src/ValidationLogic.ts
crutchcorn: This is pending, but not a big blocker. I wanna leave it as a TODO but not fix it ASAP

packages/form-core/src/ValidationLogic.ts
crutchcorn: This is fixed now

PR #1632, opened 2025-07-25
fix(form-core): canSubmitWhenInvalid not allowing form submission

814k31:


LeCarbonator:
This might be a confusion about what `canSubmitWhenInvalid` is supposed to do. The purpose is that `canSubmit` does not trigger an early return and will retry validation before submission. It should **not** call `onSubmit` if errors showed up during validation.

packages/form-core/tests/FormApi.spec.ts
```diff
@@ -3255,6 +3255,31 @@ describe('form api', () => {
...
+        },
+      },
+      onSubmit: onSubmitFake
+    })
+    form.mount()
+    expect(form.state.isValid).toBe(false)
+    await form.handleSubmit()
+    expect(onSubmitFake).toHaveBeenCalled();
+  })
+
```
LeCarbonator: This is not intended behaviour. However, it may hint at the documentation for `canSubmitWhenInvalid` needing more information. If you want to change the PR to be that instead, I'm happy to review that instead.

LeCarbonator (changes requested):

packages/form-core/tests/FormApi.spec.ts
814k31: I’m happy for you to do that

Ive rejigged our form so I don’t necessarily need this feature now

But I did notice there’s an onSubmitInvalid option

Could also be worth adjusting the new test to make sure that’s called irregardless of the canSubmitWhenInvalid (I could be wrong but I don’t think I saw any tests for that)

im happy to adjust that test on Monday if needed

LeCarbonator: Sure, that sounds good!

814k31: I’ve reverted the code now and added the test

it would only trigger the onSubmitInvalid callback when the canSubmitWhenInvalid is true so it makes sure that’s works

ill leave it in your court now to add more documentation if you want or any changes to the or you would like

PR #1635, opened 2025-07-26
chore: sync repo config

lachlancollins:
## 🎯 Changes

- Update issues template
- Migrate scripts from JS to TS
- Use Node v24
- Fix some broken links

## ✅ Checklist

- [x] I have followed the steps listed in the [Contributing guide](https://github.com/TanStack/config/blob/main/CONTRIBUTING.md).
- [x] I have tested and linted this code locally.


LeCarbonator:
Thanks for the fixes! Is there anything to keep in mind so it stays in sync for the future?

lachlancollins:
@LeCarbonator nothing we aren't already doing, the TanStack Config project tries to sync as much as possible, but many things need to be local to a repo unfortunately.

LeCarbonator:
@lachlancollins alright, noted. Thanks for taking care of that then!

PR #1641, opened 2025-07-29
fix(react-form): retain field and form context during Hot Module Replacement

LeCarbonator:
This PR moves the field and form context to module scope and exports them through the context creation. It avoids Vite's HMR creating new context and breaking the connection by accident.

This will need testing to confirm that it fixed the problems. Until that is confirmed, it's marked as draft.


LeCarbonator:
The minimal reproduction is confirmed to work with the PR version. Will merge in a bit unless there's an issue somewhere.

PR #1659, opened 2025-08-02
feat(core): form identifiers

harry-whorlow:
Add's form identifiers to form in preparation for dev-tools


harry-whorlow:
Im thinking about adding fieldId too, that you an pass to form props (which will add generated path to the form's fieldId) and field props (overrides form fieldId). But I'm not certain on this yet.

harry-whorlow:
I'm not sure it's useful to store the ID in the store, I figured someone might want to access it through the useStore... not entirely convinced though

LeCarbonator:
@harry-whorlow The only way to change the `formId` is to pass a different value to the `useForm` hook, at which point the id might as well be passed by whatever's changing anyways.

I don't mind it being in store or it being basic state, it's up to you.

harry-whorlow:
@LeCarbonator yeah, I was thinking about this last night... seems like extra code for no real benefit, and at least for now is probably not going to be used very often at least till dev tools are added.

I'm unconvinced we even need to include it in the docs, or at least the current implementation of the docs. I'm struggling to see where this would be appropriate to put since it's pretty intuitive.

PR #1660, opened 2025-08-02
docs(react-form): fix invalid HTML markup in Form Composition

Jaime02:
The HTML `label` tag should not contain a `div`, so this change replaces the `div` with a `span`

Source:
https://caninclude.onrender.com/caninclude?child=div&parent=label


crutchcorn:
Thanks!

Jaime02:
By the way, @LeCarbonator mentioned that there are more snippets that have the same issue, he used the regex `<label>\n.*<div` to find possible matches

PR #1661, opened 2025-08-02
fix(react-form): export returned types from `createFormHook`

### LINKED ISSUE #1646: Typescript monorepo: The inferred type of 'useAppForm' cannot be named without a reference to
### Describe the bug

On line 17 of `packages/form/src/form.ts` you see the Typescript error.

```
The inferred type of 'useAppForm' cannot be named without a reference to '../../../node_modules/@tanstack/react-form/dist/esm/createFormHook'. This is likely not portable. A type annotation is necessary. ts(2742)
```

https://github.com/kristian-puccio/ts-form-mono

### Steps to reproduce

1. pnpm i
2. Open the file `packages/form/src/form.ts`

### Expected behavior

Typescript not to throw an error, earlier versions on Tanstack form didn't have an error here.

None

### Screenshots or Videos

Image

### Platform

macOS

### TanStack Form adapter

react-form

### TanStack Form version

1.15.1

### TypeScript version

5.8.3

### Additional context

If you remove the `extends` from `packages/form/tsconfig.json` the error goes away.
This is the standard NX way of setting up a TS mono repo.
This error doesn't happen with Tanstack Form 1.14.2 or earlier.

LeCarbonator:
Close #1646 

This type export is required so that libraries with declaration file generation can create explicit types instead of non-portable references.

It's finnicky to trigger the warning or error, so feedback will be needed from users in the linked issue before this is safe to merge.

- [ ] Get feedback from affected users in #1646

### DIFF
--- a/packages/react-form/src/createFormHook.tsx
+++ b/packages/react-form/src/createFormHook.tsx
@@ -133,9 +133,6 @@ interface CreateFormHookProps<
   formContext: Context<AnyFormApi>
 }
 
-/**
- * @private
- */
 export type AppFieldExtendedReactFormApi<
   TFormData,
   TOnMount extends undefined | FormValidateOrFn<TFormData>,
--- a/packages/react-form/src/createFormHook.tsx
+++ b/packages/react-form/src/createFormHook.tsx
@@ -133,6 +133,9 @@ interface CreateFormHookProps<
   formContext: Context<AnyFormApi>
 }
 
+/**
+ * @private
+ */
 export type AppFieldExtendedReactFormApi<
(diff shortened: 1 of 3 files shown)


crutchcorn (changes requested):
`@private` only marks if documentation should be made, not if the code should be exported.

mhuggins:
Is anyone else still running into this issue?  I'm seeing this when creating fields for `@monaco-editor/react` with the latest version of `@tanstack/react-form` (1.23.7).

Screenshot 2025-10-21 at 2 26 24 PM

PR #1675, opened 2025-08-04
docs(react-form): Fix field group name

bennettdams:
These changes harmonizes the occurrences of the password field group for all examples to have the same name.


LeCarbonator:
Good catch, thanks!

PR #1679, opened 2025-08-06
fix(form-core): infer `formOptions` parameter's type correctly

a-is-4-adam:
- Add TFormData generic parameter to formOptions function
- Add test to verify listener type preservation with formOptions
- Current implementation has intentional type issues for investigation

### DIFF
--- a/packages/form-core/src/formOptions.ts
+++ b/packages/form-core/src/formOptions.ts
@@ -4,35 +4,59 @@ import type {
   FormValidateOrFn,
 } from './FormApi'
 
+/*
+
+These types need to do two things:
+
+1. Validator generics need to depend on the TFormData generic
+2. The resulting needs to allow overriding values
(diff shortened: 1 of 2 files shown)


LeCarbonator:
There was a suggestion with using reverse mapped types, but I'll test it locally first because it may end up being too separated from the original types.

Do you mind if I commit the changes here if they work?

a-is-4-adam:
> There was a suggestion with using reverse mapped types, but I'll test it locally first because it may end up being too separated from the original types.
> 
> Do you mind if I commit the changes here if they work?

Yeah go for it :)

LeCarbonator:
well, it was worth a shot.

LeCarbonator:
Alright, now we're talking. Would you mind giving this implementation a user test first? @a-is-4-adam

Issue #1526, opened 2025-05-23
form.setFieldValue doesn't trigger validation


amirhhashemi:
I noticed that calling `form.setFieldValue("field-name", myvalue)` doesn't trigger validation. [Looking at the code](https://github.com/TanStack/form/blob/8b2b65cd5bff82598996d4a1beeae345b4412f93/packages/form-core/src/FormApi.ts#L1910), similar methods, like `pushFieldValue`, call `form.validateField()` after updating the value. That doesn't happen in `setFieldValue()`.

LeCarbonator:
`setFieldValue` is meant for programmatically setting a field value, while `pushFieldValue` is an array specifically meant to handle change (like `field.handleChange`).

Could you provide a reproducible example of where the validation from `setFieldValue` is required / desired?

I can see reasoning for both not forcing validation and forcing validation. It's possible to add it as option parameter like the `dontUpdateMeta` prop, but I'd like to know what the use case is first.

amirhhashemi:
In my situation, I have a file input and I want to implement the logic for handling file changes outside of JSX for better readability. I cannot access `field.handleChange` outside the `Form.Field` component unless I pass it to my `handlePictureChange` function. I found that `form.setFieldValue` is the closest alternative I could use, and I expected it to behave similarly to `field.handleChange`. It took me some time to realize why changing the file doesn't trigger validation.

In my opinion, `form.setFieldValue` should trigger validation by default, with an option like `dontValidate` available to disable it if needed.

```tsx
import { useState, type ChangeEvent } from "react";
import { useForm } from "@tanstack/react-form";
import { z } from "zod/v4-mini";

const schema = z.object({
  profilePicture: z.optional(z.file().check(z.maxSize(2_000_000))),
});

function ProfileForm() {
  const [avatarPreview, setAvatarPreview] = useState<string>();

  const form = useForm({
    defaultValues: { profilePicture: undefined } as {
      profilePicture: File | undefined;
    },
    validators: {
      onChange: schema,
    },
  });

  async function handlePictureChange(e: ChangeEvent<HTMLInputElement>) {
    const file = e.target.files?.[0];
    if (file) {
      form.setFieldValue("profilePicture", file);
      const reader = new FileReader();
      reader.onload = () => {
        setAvatarPreview(reader.result as string);
      };
      reader.readAsDataURL(file);
    }
  }

  return (
    <form
      onSubmit={async (e) => {
        e.preventDefault();
        await form.handleSubmit();
      }}
[...]

PR #1680, opened 2025-08-07
fix(form-core): Trigger listeners and validation on `setFieldValue`

LeCarbonator:
Also adds two meta properties, `dontValidate` and `dontRunListeners`. Should help in case someone wants extreme control, but it's more for our sake.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -9,6 +9,7 @@ import {
   getAsyncValidatorArray,
   getBy,
   getSyncValidatorArray,
+  mergeOpts,
 } from './utils'
 import { defaultValidationLogic } from './ValidationLogic'
 import type { DeepKeys, DeepValue, UnwrapOneLevelOfArray } from './util-types'
@@ -1350,11 +1351,19 @@ export class FieldApi<
    * Sets the field value and run the `change` validator.
    */
   setValue = (updater: Updater<TData>, options?: UpdateMetaOptions) => {
-    this.form.setFieldValue(this.name, updater as never, options)
+    this.form.setFieldValue(
+      this.name,
+      updater as never,
+      mergeOpts(options, { dontRunListeners: true, dontValidate: true }),
+    )
 
-    this.triggerOnChangeListener()
+    if (!options?.dontRunListeners) {
+      this.triggerOnChangeListener()
+    }
 
-    this.validate('change')
+    if (!options?.dontValidate) {
+      this.validate('change')
+    }
   }
 
   getMeta = () => this.store.state.meta
@@ -1400,11 +1409,17 @@ export class FieldApi<
    */
   pushValue = (
     value: TData extends any[] ? TData[number] : never,
-    opts?: UpdateMetaOptions,
+    options?: UpdateMetaOptions,
   ) => {
-    this.form.pushFieldValue(this.name, value as any, opts)
+    this.form.pushFieldValue(
+      this.name,
+      value as any,
+      mergeOpts(options, { dontRunListeners: true }),
+    )
 
-    this.triggerOnChangeListener()
+    if (!options?.dontRunListeners) {
+      this.triggerOnChangeListener()
+    }
   }
 
   /**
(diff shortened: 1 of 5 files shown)


LeCarbonator:
I initially thought I should mark as draft to refactor how form listeners are currently set up. However, it appears that it would be a breaking change since `fieldApi` is always passed to it.

So I suppose this is already done.

packages/form-core/src/FieldApi.ts
```diff
@@ -1350,11 +1351,19 @@ export class FieldApi<
    * Sets the field value and run the `change` validator.
    */
   setValue = (updater: Updater<TData>, options?: UpdateMetaOptions) => {
-    this.form.setFieldValue(this.name, updater as never, options)
+    this.form.setFieldValue(
+      this.name,
+      updater as never,
+      mergeOpts(options, { dontRunListeners: true, dontValidate: true }),
```
marc-niclas, line 1357: so if `options === undefined`, the way `mergeOpts` works, is that it will return `dontValidate === true`. therefore, by default, calling `setFieldValue` does not validate?

LeCarbonator: See lines 1360-1366. We are telling `setFieldValue` that we're taking care of it ourselves and it doesn't need to invest its work for that.

`mergeOpts` simply ensures that the overrides are present for this internal call

MelDommer:
Hello just wanted to say I ran into this issue just a few hours ago.

I have a form a very complex one 100+ fields.
I had a situation where changing field A ran an onChange listener, which programmatically set Field B.  Well field B happened to also have a listener that programatically set field C.

The form is pretty linear in nature so this helps populate fields with the recommended values as things are being setup.

The issue I ran into was documented here:
https://github.com/TanStack/form/issues/1526#issuecomment-2949049565

As I scrolled through I saw @LeCarbonator ask for testing:
https://github.com/TanStack/form/issues/1526#issuecomment-3199449377

I wanted to let you know I installed this version from the PR and I am correctly seeing listeners being triggered from setFieldValue, where I wasn't before (but expected to.)

I hope this helps and I look forward to seeing this make it into a new release soon!

ws-rush:
I'm working with a large, 260+ field form with complex dependencies that has become difficult to maintain using React Hook Form. After research, I've narrowed my choices to TanStack Form and Alibaba Formily because of their excellent side-effect management. I intend to test this PR to compare them for my project. Even if I don't end up using TanStack Form, I'm eager to see this PR merged quickly.

MelDommer:
I appear to have found a workaround for this.  I have tested it and it seems to be working but maybe there is some hidden reason not to do it this way.

Example say you have a fields that have a "recommended" value, but can be changed from the recommended to something else (although this is rare.)

Field A - Independent field (Has a listener that sets field B to the recommended value.)

Field B - used in truth table (Has a listener that sets E to the recommended value.)
Field C - used in truth table (Has a listener that sets E to the recommended value.)
Field D - used in truth table (Has a listener that sets E to the recommended value.)

Field E - has a recommended value from B,C,D.

When the user selects a value from field "A" the listener runs and sets field "B" 
```ts
// Field A's listener
fieldApi.form.setFieldValue("B", "new value"); // does not trigger B's listener.
```
This has a cascade effect and now B's listener doesn't evaluate the truth table to set "E" to the recommended value.

This PR does fix my issue and works well for my use case and it much less verbose than the workaround I just found.

Workaround is to get the instance and call handleChange from the listeners.  It seems to work the same as this PR's changes and I haven't noticed any issues, but maybe this workaround is not good.
```ts
// fieldApi.form.setFieldValue("B", "new value");
fieldApi.form.fieldInfo.B.instance?.handleChange("new value"); // this does trigger "B"'s listeners.
```

Field's B,C,D use the same syntax for setting "E" after evaluating to come up with what it should be eg:
```ts
// B,C,D's listeners.
fieldApi.form.fieldInfo.E.instance?.handleChange("new value");
```

I will use this workaround for now until setFieldValue works the same way.

LeCarbonator:
The reason it wasn't done was mostly because it was used internally. With the overrides in place in the PR, I expect this to get merged by the end of the week.

PR #1690, opened 2025-08-14
fix(react-core): nextjs `onServerValidate` decode options types

tarantoj:
The `decode` function from `decode-formdata` [is overloaded](https://github.com/fabian-hiller/decode-formdata/blob/e8b6689734ad14df7f4248ce066512ef84bc596f/src/decode.ts#L35), leading to some type quirks when trying to pass in `FormDataInfo` to `serverValidate`


tarantoj:
Unfortunately, I don't think actually resolves the issue I'm having. I'm going to close this pull request and create an issue.

PR #1691, opened 2025-08-14
fix: sync validation errors for delayed-mounted fields

jiji-hoon96:
Fixes validation errors not appearing on fields that mount after form initialization 

## Problem

When fields are mounted after the form's initial validation runs, validation errors from form-level validators (like [onMount]) are not displayed on these delayed fields, leading to inconsistent validation UX.

## Solution

- **FormApi**: Store all field errors (including for unmounted fields) in `_allFieldErrors` state
- **FieldApi**: Sync existing errors from form state when fields mount, marking them as originating from form validation

## Changes

- Added `_allFieldErrors` property to BaseFormState to persist validation errors for all fields
- Modified validateSync to store field errors for both mounted and unmounted fields
- Enhanced `FieldApi.mount()` to sync existing errors from form state on field mount
- Added comprehensive tests covering delayed mount, multiple fields, and remount scenarios

Closes #17


brandonleichty:
👀

jiji-hoon96:
@LeCarbonator I thought this issue needed to be resolved promptly, and I'm curious to hear your thoughts on the matter.

brandonleichty:
I'm also quite curious about this one. I've had to use `hidden` in all of my forms, rather than conditional rendering. Which doesn't feel great—especially for future readers of the code.

LeCarbonator:
I'm just not sure if this is the way to go about it. I've been tinkering behind the scenes to see what truly causes this, but I suppose I should focus more on it so it can get fixed

jiji-hoon96:
I'm not entirely sure why the delayed-mounted field isn't receiving form validation either.

While my proposed method could resolve it, errors exist in both `fieldMeta` and `_allFieldErrors`. This raises concerns about duplicate states and memory overhead issues, as it would persistently store errors for all unmounted fields.

If you have any ideas, please share them so we can explore them together.

I have three potential approaches in mind

1.  Event-driven approach: Re-trigger form validation when the field mounts, re-executing the validation results only for this field.
2.  Store validation results as a function and evaluate them when needed.
3.  Subscribe to the form's validation state when the field mounts.

LeCarbonator:
The issue is that while mapping form validators to respective fields, it only iterates through the existing record of (previously registered) fields.
Ideally, it should have the following behaviour:

* If a field error does not have an entry in the fieldInfo record, create one
* If `deleteField` or `removeValue` is called, it should remove that entry

jiji-hoon96:
@LeCarbonator  FormApi has been improved with a comprehensive error-handling approach that includes all field errors, and features have been added to automatically generate fieldInfo and prevent memory leaks in the deleteField method.

brandonleichty:
@jiji-hoon96 does this mean the issue should be resolved?

jiji-hoon96:
@brandonleichty Yes!

packages/form-core/src/FormApi.ts
```diff
@@ -1532,9 +1537,35 @@ export class FormApi<
...
+        const allFieldsToProcess = new Set([
+          ...Object.keys(this.state.fieldMeta),
+          ...Object.keys(fieldErrors || {}),
+        ] as DeepKeys<TFormData>[])
+
+        for (const field of allFieldsToProcess) {
+          if (fieldErrors?.[field] && !this.fieldInfo[field]) {
+            this.getFieldInfo(field)
+          }
+
```
LeCarbonator: This is going in the right direction, but `_allFieldErrors` is not the way. Iterating over previously mounted fields in `fieldMeta`  is not enough for validation errors (like you correctly asserted), so ones that are missing should be added to it.

Fields derive their data from `fieldMeta`, so if they were added here, they would have the data not just on mount, but on construction.

One issue that we might run into with this approach is that wrong field names can result in a locked state. There is no field to read the stored error, so it is never visually shown. I **don't** think that's an issue because:

* Our standard schemas are type safe when it comes to keys.
* Standard schemas send a copy of the error list to the form error state, so it can be figured out if there's an issue.

Overall, I think the right way to go is to add missing `fieldMeta` from field errors if they're not present. Use the default field meta state for the other props being added.

packages/form-core/tests/FieldInfo-management.test.ts
LeCarbonator, line 1: This should be part of `FieldApi.spec.ts`, since they're all related to deleting / delay mounting fields.

packages/form-core/tests/FieldInfo-management.test.ts
```diff
@@ -0,0 +1,381 @@
...
+
+      form.mount()
+
+      const existingField = new FieldApi({
+        form,
+        name: 'existingField',
+      })
+      existingField.mount()
+
+      await new Promise((resolve) => setTimeout(resolve, 100))
```
LeCarbonator: There is a `sleep` helper from `utils.ts` you can use, along with `vi.useFakeTimers()`. See other async unit tests for details.

packages/form-core/tests/FieldInfo-management.test.ts
```diff
@@ -0,0 +1,381 @@
...
+      form.mount()
+
+      const field = new FieldApi({
+        form,
+        name: 'fieldToDelete',
+      })
+      field.mount()
+
+      expect(form.fieldInfo.fieldToDelete).toBeDefined()
+      expect(form.state._allFieldErrors?.fieldToDelete).toBeDefined()
```
LeCarbonator: The above comment suggests that no `_allFieldErrors` should be added. Remember to address these unit tests related to it.

packages/form-core/tests/hidden-field-mount.test-d.ts
LeCarbonator, line 1: Same as above, this deals with `FieldApi`, so it should be part of `FieldApi.spec.ts`

LeCarbonator (changes requested):
Thanks for the PR! This is pretty close, but there are some things to address. If you'd like, I can help out with addressing them.

packages/form-core/src/FormApi.ts
jiji-hoon96: Thank you for the detailed feedback! 
You're absolutely right. I've refactored the implementation to use `fieldMeta` instead of `_allFieldErrors`.

packages/form-core/tests/FieldInfo-management.test.ts
jiji-hoon96: Great catch! I've replaced all setTimeout promises with the `sleep` helper from `utils.ts` and added `vi.useFakeTimers() / vi.useRealTimers()` to the async tests, following the pattern used in other test files.

PR #1692, opened 2025-08-14
feat(core): Form Devtools

harry-whorlow:


harry-whorlow:
Some spoilers you those of you following, if any of you have any suggestions drop them here. 🤘

https://github.com/user-attachments/assets/2c600a12-4e98-4181-baee-1f7d214535f2

beaussan:
> Some spoilers you those of you following, if any of you have any suggestions drop them here. 🤘

IMO, it's a great start! 

I would say the fields value / states are more important than the forms options or such. I'm biased, since we use react-hook-form devtools quite a lot, but having a view grouped by fields is what we get the most value of the react hook form devtools. Usually, what we are trying to answer is questions like:
- what's the form's internal value for this element ?
- is it dirty or not ? 

With the current demo, I have to scroll quite a lot, and the value is not colocated with the state.

The form submit history is amazing ! :D 

So I would say, definitely useful in the curent state, but ux could be improved a bit to get useful info quicker

image

harry-whorlow:
@beaussan thats good to hear!

I was planning on getting neoDrag involved, since were going to be using it shortly in the devtools for split panels, that way you can drag to order of the info sections to whatever you like, then I'll store it in something like local so it stays. 

But thank you for the feedback, I definitely see your point about the fields being the important part. I might re-order it for now, then when neoDrag is added swap it back for sake of order.

AlemTuzlak:
@beaussan react-hook-form has devtools???? oh my how did I not know about this

docs/framework/react/guides/devtools.md
```diff
@@ -0,0 +1,58 @@
...
+Install the [TanStack Devtools](https://tanstack.com/devtools/) library, this will install the devtools core as well as provide you framework specific adapters.
+
+```bash
+npm i @tanstack/devtools
+```
+
+Next in the root of your application import the `TanstackDevtools` from the required framework adapter (in this case @tanstack/react-devtools).
+
+```tsx
+import { TanstackDevtools } from '@tanstack/react-devtools'
```
AlemTuzlak, line 20: I would change this to TanStackDevtoolsPanel (also notice the capital S) to be consistent with all the other devtools we have. and another thing, I'd add a plugin like in this PR so you can just add the plugin instead of manually defining the name, it will also allow you to add breaking changes without breaking it for the users:
https://github.com/TanStack/pacer/pull/62

packages/form-core/package.json
```diff
@@ -51,6 +51,7 @@
     "src"
   ],
   "dependencies": {
+    "@tanstack/devtools-event-client": "^0.2.2",
```
AlemTuzlak: bump to 0.2.3, there was a minor bug in 0.2.2 that could affect form as well

packages/form-core/src/EventClient.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export type RequestFormForceReset = {
+  id: string
+}
+
+type EventMap = {
+  'form-devtools:broadcast-form-state': BroadcastFormState
+  'form-devtools:broadcast-form-submission-state': BroadcastFormSubmissionState
+  'form-devtools:broadcast-form-unmounted': BroadcastFormUnmounted
```
AlemTuzlak: with event systems you usually are already implicitly implying you're broadcasting something so you could omit the broadcast and just call it somethign like `form-state-changed`, then the listeners feel more natural:
```ts
listener.on("form-state-changed")
```

packages/form-devtools/src/TanstackFormDevtoolsPanel.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export interface TanstackFormDevtoolsPanelConfig {}
+
+class TanstackFormDevtoolsPanel {
+  #isMounted = false
+  #dispose?: () => void
+
+  constructor() {}
+
+  mount<T extends HTMLElement>(el: T) {
```
AlemTuzlak: you can also now add light/dark theme support with this, check out the reference implementation in Pacer:
https://github.com/TanStack/pacer/pull/62/files#diff-fdb04873a8517e255f24ae322819948fc123464f5fa04180f0e6a3291335a56bR14

and then the plugin that consumes it:
https://github.com/TanStack/pacer/pull/62/files#diff-5a4d1d5da0058864889491e7d630577c76284ec7a611b69e06c7015657dad38cR4

packages/form-devtools/src/components/Shell.tsx
```diff
@@ -0,0 +1,79 @@
...
+  })
+
+  onCleanup(() => {
+    document.removeEventListener('mousemove', handleMouseMove)
+    document.removeEventListener('mouseup', handleMouseUp)
+  })
+
+  return (
+    <div class={styles().devtoolsPanel}>
+      <div class={styles().stickyHeader}>TanStack Form</div>
```
AlemTuzlak: update to the latest devtools-ui and you get:
- MainPanel
- Header
- HeaderLogo 

to use here

packages/form-devtools/src/components/StateHeader.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  const updatedAt = createMemo(() => state()?.date.unix() ?? dayjs().unix())
+
+  const relativeTimeText = createMemo(() => {
+    const diffSeconds = now() - updatedAt()
+    if (diffSeconds < 60) {
+      return `${diffSeconds} second${diffSeconds !== 1 ? 's' : ''} ago`
+    }
+    return dayjs.unix(updatedAt()).fromNow()
+  })
```
AlemTuzlak: in one of the demos you sent I know there was a bug that said "last updated -1s ago" not sure if that's fixed?

packages/form-devtools/src/contexts/eventClientContext.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+function useProviderValue() {
+  const [store, setStore] = createStore<DevtoolsState>(devtoolsStateInit)
+
+  createEffect(() => {
+    const cleanup = formEventClient.on('broadcast-form-state', (_e) => {
+      const e = _e as unknown as {
+        type: EventClientEventNames
+        payload: BroadcastFormState
+      }
```
AlemTuzlak: why this?

AlemTuzlak (approved):
I left thoughts/comments on everything, feel free to do with the feedback as you see fit!

docs/framework/react/guides/devtools.md
harry-whorlow: Screenshot 2025-09-06 at 11 29 37

Is it not exported as TanStackDevtools though?

AlemTuzlak: I am stupid, I thought this was form devtools

harry-whorlow: it's cool, you had me confused for a second.

docs/framework/react/guides/devtools.md
```diff
@@ -0,0 +1,58 @@
...
+---
+
+TanStack Form comes with a ready to go suit of devtools.
+
+## Setup
+
+Install the [TanStack Devtools](https://tanstack.com/devtools/) library, this will install the devtools core as well as provide you framework specific adapters.
+
+```bash
+npm i @tanstack/devtools
```
AlemTuzlak: Shouldn't this be react-devtools, or ideally point to geting started of the devtools so you don't have to worry about it?

AlemTuzlak: The sentence above is not correct as the core does not provide the adapters 😅

docs/framework/react/guides/devtools.md
```diff
@@ -0,0 +1,58 @@
...
+
+import App from './App'
+
+createRoot(document.getElementById('root')!).render(
+  <StrictMode>
+    <App />
+
+    <TanstackDevtools
+      plugins={[
+        {
```
AlemTuzlak: When I suggested the plugin this is where it would go, probably forgot to change? I saw you added it.

packages/form-devtools/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@tanstack/form-devtools",
+  "version": "1.17.0",
```
AlemTuzlak: do you want to release the beta as version 1.17?

harry-whorlow: good catch

.changeset/config.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@3.1.1/schema.json",
+  "changelog": [
+    "@svitejs/changesets-changelog-github-compact",
+    { "repo": "TanStack/devtools" }
```
AlemTuzlak: I think you might've missed something :D

harry-whorlow: The key to originality is hiding your sources

AlemTuzlak (approved):

PR #1697, opened 2025-08-15
fix(form-core): call `onSubmitInvalid` even when `canSubmit` is false

### LINKED ISSUE #1696: `form.onSubmitInvalid` not called when `canSubmit` is false
### Describe the bug

When `handleSubmit` is called while `canSubmit` is `false` an early return is hit and `onSubmitInvalid` is not called.

https://github.com/TanStack/form/blob/1e58e57987e4f79fb23d0034d1cdceebb9a7299b/packages/form-core/src/FormApi.ts#L1918

https://stackblitz.com/edit/tanstack-form-dqy8vhes?file=src%2Findex.tsx

### Steps to reproduce

1. Go to the Stackblitz
2. 👁️ Observe: `"canSubmit": false,` (see JSON-dump in `<pre>`-Tag)
3. Click the Submit Button
4. 👁️ Observe: No `console.log` for `"onSubmitInvalid"`

### Expected behavior

As a User I expect `form.onSubmitInvalid` to be called whenever the form is submitted in an invalid state

Every time

### Screenshots or Videos

_No response_

### Platform

irrelevant

### TanStack Form adapter

react-form

### TanStack Form version

1.19.2

### TypeScript version

_No response_

### Additional context

_No response_

Pascalmh:
`onSubmitInvalid` callback in `FormApi` is called when a submission attempt is made while `canSubmit` is false

Fixes #1696

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -2017,13 +2017,13 @@ export class FormApi<
     const submitMetaArg =
       submitMeta ?? (this.options.onSubmitMeta as TSubmitMeta)
 
-    if (!this.state.canSubmit) {
+    if (!this.state.canSubmit && !this._devtoolsSubmissionOverride) {
       this.options.onSubmitInvalid?.({
         value: this.state.values,
         formApi: this,
         meta: submitMetaArg,
       })
-      if (!this._devtoolsSubmissionOverride) return
+      return
     }
 
     this.baseStore.setState((d) => ({ ...d, isSubmitting: true }))


packages/form-core/src/FormApi.ts
```diff
@@ -1915,11 +1915,18 @@ export class FormApi<
...
     const submitMetaArg =
       submitMeta ?? (this.options.onSubmitMeta as TSubmitMeta)
 
+    if (!this.state.canSubmit) {
+      this.options.onSubmitInvalid?.({
+        value: this.state.values,
+        formApi: this,
+        meta: submitMetaArg,
+      })
+      return
```
kusiewicz: While this ensures `onSubmitInvalid` gets called, it happens before the submit validation. This means the `formApi` will have a stale error state when the callback is triggered.

LeCarbonator:
@kusiewicz It happens before the submit validation, but only in cases where `canSubmit` is false. That means there are currently errors present that have not been resolved, which is not stale.

If a user wanted to rerun validators even if `canSubmit` is false, they can use the `canSubmitWhenInvalid` property

kusiewicz:
> @kusiewicz It happens before the submit validation, but only in cases where `canSubmit` is false. That means there are currently errors present that have not been resolved, which is not stale.
> 
> If a user wanted to rerun validators even if `canSubmit` is false, they can use the `canSubmitWhenInvalid` property

Ok i got it. I just thought that there:

```
  onSubmitInvalid: ({ formApi }) => {
    console.log(formApi.state.fieldMeta.xxx.errors);
  }
```

we should expect most recent errors, didn't know about `canSubmitWhenInvalid`

LeCarbonator:
well, they are the most recent errors. The most recent errors are preventing a submit altogether, so this PR asserts that attempting one should still trigger `onSubmitInvalid`

jiji-hoon96:
Hello @Pascalmh!

I encountered the same issue mentioned by @kusiewicz. Calling `onSubmitInvalid` before performing validation may result in outdated error states being passed to the callback.  

Would you consider restructuring the logic to first perform the validation, check `canSubmit`, and then call `onSubmitInvalid` with the updated state? In this case, the user would receive the most recent validation error in the callback.

If it's okay with you, would it be possible for me to work on this and submit a PR?

Pascalmh:
@jiji-hoon96 sure, if you want you can PR against my fork and I'll merge your changes in so they appear in this PR here - or you can create a new PR, whatever you prefere 👍

LeCarbonator:
@jiji-hoon96 Are you sure it's stale? Even with `canSubmitWhenInvalid: true`?

jiji-hoon96:
Yes. I have confirmed its validity through testing. I may have made a mistake, so I would appreciate it if you could take a look at #1711 PR.

jiji-hoon96:
https://github.com/Pascalmh/tanstack-form/pull/5

I have submitted a pull request with some modifications to @Pascalmh`s work.

LeCarbonator:
This has been long overdue. Thanks for the fix!

PR #1706, opened 2025-08-19
fix(form-core): prevent runtime errors when using `deleteField`

kusiewicz:
During implementing [this](https://github.com/TanStack/form/discussions/1684) and testing `deleteField` I found out that this method does not work.

When a form field has `onChange` validation and the `deleteField` method is called on it, an error appears.

```
FormApi.ts:1586 Uncaught TypeError: Cannot read properties of undefined (reading 'errorMap')
    at FormApi.ts:1586:27
    at functionalUpdate (utils.ts:26:8)
    at FormApi.ts:2058:20
```
error: https://bolt.new/~/vitejs-vite-ceu2fl73

I think flow look like this:

1. Remove button click => `form.deleteField('firstName')`.
2. Delete function changes `values` so `onChange` validator runs.
3. This validator runs on the form level and still tries to set `errors.fields.email = 'error message'` - so `validateSync` tries to update metadata of deleted field.
4. `setFieldMeta` calls `functionalUpdate` with `undefined` (since field does not exists in `fieldMetaBase`).
5. Updater tries to do `...prev` where `prev` now is undefined -> error.

So `validateSync` iterates through field list that already semes to be outdated. There is a check `if (!fieldMeta)` in this function but I guess we have to add as well `if (!fieldMetaBase)`.

### DIFF
--- a/packages/react-form/tests/useForm.test.tsx
+++ b/packages/react-form/tests/useForm.test.tsx
@@ -893,7 +893,7 @@ describe('useForm', () => {
     expect(result).toHaveTextContent('1')
   })
 
-   it('should not error when using deleteField in edge cases', async () => {
+  it('should not error when using deleteField in edge cases', async () => {
     function Comp() {
       const form = useForm({
         defaultValues: {


kusiewicz:
Please, verify the idea. If I'm thinking in a proper direction I will proceed and add unit tests.

kusiewicz:
@LeCarbonator? :)

LeCarbonator:
Will try to review this in the coming days

LeCarbonator (approved):
I'm not convinced this would be the ideal solution, but this minor fix helps catch the irritating runtime errors. Thanks for the step-by-step explanation!

PR #1711, opened 2025-08-21
fix(form-core): onSubmitInvalid callback not being called when canSubmit is false

jiji-hoon96:
Problem The onSubmitInvalid callback was not being called when canSubmit is false because the  handleSubmit method returned early before running validation.

## Solution
- Modified handleSubmit logic to always run validation regardless of canSubmit state
- Call onSubmitInvalid with updated validation state when canSubmit is false
- Respect canSubmitWhenInvalid option behavior

## Changes
- Updated FormApi.handleSubmit() method in FormApi.ts
- Added comprehensive test cases for onSubmitInvalid callback scenarios
- Added tests for canSubmitWhenInvalid option interaction

#1697 #1696

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -1915,8 +1915,6 @@ export class FormApi<
       )
     })
 
-    if (!this.state.canSubmit) return
-
     const submitMetaArg =
       submitMeta ?? (this.options.onSubmitMeta as TSubmitMeta)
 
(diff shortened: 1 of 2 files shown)


LeCarbonator:
The problem is not that canSubmit causes an early return. The errors were caused by validation that has not yet been resolved, which means you have no reason to rerun validation on data that's already invalid.

The user has the option to bypass this using `canSubmitWhenInvalid`. However, the default behaviour is that `canSubmit: false` should not run `validate('submit')`.

The unit tests could be useful for ensuring proper `onSubmitInvalid` behaviour, though. Would you mind adding them to #1697 ?

jiji-hoon96:
I understand.
Then I will close the PR and upload only the modified unit test to #1697.

PR #1717, opened 2025-08-28
fix(form-core): prevent stale reactivity when using Date objects

### LINKED ISSUE #1712: `isDefaultValue` not updating when changing date field
### Describe the bug

On changing a date field, it seems tanstack is not changing `isDefaultValue` to `false`, as I would expect.

https://stackblitz.com/edit/vitejs-vite-sobdcz8j?file=src%2FApp.tsx

### Steps to reproduce

1. Change the date
2. Observe equality check is false and `isPristine` is `false`
3. `isDefaultValue` is still `true`
4. Changing the text field however does change `isDefaultValue` to `false`

### Expected behavior

Changing a date field should change `isDefaultValue` to `false`

Every time

### Screenshots or Videos

Image

### Platform

- macOS 15.5
- chrome/brave

### TanStack Form adapter

None

### TanStack Form version

1.19.2

### TypeScript version

_No response_

### Additional context

_No response_

m2na7:
## Description
Fixes issue where `isDefaultValue` was not updating correctly when changing date fields.

## Problem
The `evaluate` function was treating Date objects as regular objects, causing incorrect equality comparisons for date values.

## Solution
- Added specific Date object comparison using `getTime()` method
- Added comprehensive tests for Date object comparison scenarios

## Fixes
Closes #1712

### DIFF
--- a/packages/form-core/src/utils.ts
+++ b/packages/form-core/src/utils.ts
@@ -431,6 +431,10 @@ export function evaluate<T>(objA: T, objB: T) {
     return false
   }
 
+  if (objA instanceof Date && objB instanceof Date) {
+    return objA.getTime() === objB.getTime()
+  }
+
   if (objA instanceof Map && objB instanceof Map) {
     if (objA.size !== objB.size) return false
     for (const [k, v] of objA) {
@@ -539,3 +543,4 @@ export function createFieldMap<T>(values: Readonly<T>): { [K in keyof T]: K } {
 
   return output
 }
+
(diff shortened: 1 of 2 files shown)


marc-niclas:
Can we merge this? This is an issue in our prod and it was just reported as an issue by a user so please fix this!

m2na7:
@LeCarbonator I’d like to get your thoughts on this PR! Please check it at your convenience.

LeCarbonator:
Sounds good! We've had discussions about migrating this to tanstack store. Until then, this will serve as a good fix. Thanks!

PR #1718, opened 2025-08-28
fix(form-core): fix formApi.reset() in form onSubmit

### LINKED ISSUE #1681: form.reset during onSubmit ignores new default values
### Describe the bug

### Steps to reproduce

1. Uncheck the checkbox.
2. Click the Submit button

### Expected behavior

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: Any
- Browser: Any
- Node Version: 22

### TanStack Form adapter

react-form

### TanStack Form version

1.18.0

### TypeScript version

v5.8.3

### Additional context

djordje-st:
Fixes #1681

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -600,6 +600,10 @@ export type BaseFormState<
    */
   fieldMetaBase: Record<DeepKeys<TFormData>, AnyFieldMetaBase>
   /**
+   * The default values that correspond to the current form state, used for determining isDirty correctly
+   */
+  _stateDefaultValues?: TFormData
+  /**
    * A boolean indicating if the form is currently in the process of being submitted after `handleSubmit` is called.
    *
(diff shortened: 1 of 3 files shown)


LeCarbonator:
This addresses the resetting values, but what about meta properties like `isSubmitting`, `isSubmitted` etc.?

They might still be affected

djordje-st:
@LeCarbonator good shout, should be handled now

gregorbg:
Not to be pushy or anything, but I would really appreciate to have this bugfix released. Is there any kind of timeline on the horizon for shipping this PR?

matthewbbrandt:
+1, it would be great to get this fixed. I have used the setTimeout workaround (discussed in [#1681](https://github.com/TanStack/form/issues/1681) ), but it would be nice to be able to remove it.

chinanderm:
Can this one get merged in? Quite the unintuitive behavior this bug is causing.

Vito-Ethan:
bump

PR #1719, opened 2025-08-29
docs(react-form): fix incorrect error display in example

taylous:
## Summary

- removed a stray semicolon (;) that appeared after a JSX comment
- replaced comment with a regular line comment (// ...) for better readability

## Description

i think that semicolon was unnecessary and could confuse readers when copying or reading the example. 🤔


LeCarbonator:
Thanks!

PR #1729, opened 2025-09-13
Draft: Investigate `FieldApi.update`

LeCarbonator:
May be tied to field values setting to undefined resetting to default Values. May also be tied to array problems.

This is mostly to get investigation going by having a 'broken' starting point and analyzing from there.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -1315,21 +1315,26 @@ export class FieldApi<
 
     // Default Value
     if ((this.state.value as unknown) === undefined) {
-      const formDefault = getBy(opts.form.options.defaultValues, opts.name)
+      const formDefault = getBy(
+        opts.form.options.defaultValues,
+        opts.name,
+      ).value
 
       const defaultValue = (opts.defaultValue as unknown) ?? formDefault
 
       // The name is dynamic in array fields. It changes when the user performs operations like removing or reordering.
       // In this case, we don't want to force a default value if the store managed to find an existing value.
-      if (nameHasChanged) {
-        this.setValue((val) => (val as unknown) || defaultValue, {
-          dontUpdateMeta: true,
-        })
-      } else if (defaultValue !== undefined) {
-        this.setValue(defaultValue as never, {
-          dontUpdateMeta: true,
-        })
-      }
+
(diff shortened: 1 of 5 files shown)


LeCarbonator:
Notes to self:

Swapping, deletion and basic reset seems to work. However, there's conflicts with the reset that need further investigating.

LeCarbonator:
`form.reset()` still has the issue if the array field's size changes.

`swapValues`, `moveValues` and `removeValue` seem to be fixed.

LeCarbonator:
`form.reset()` and array methods are all fixed.

Todos:

- [x] Test at workplace (known location where previous implementation crashed due to dnd)
- [x] Test performance (what does the rerendering look like now?)
   - [x] children callback
   - [x] Form Composition
- [ ] Check other frameworks
   - [x] Solid
   - [ ] Angular
   - [ ] Vue
   - [ ] Svelte
   - [ ] Lit

LeCarbonator:
Solid appears to be struggling with this implementation still. I'll look into what can be done to fix it.

LeCarbonator:
~~Looks like svelte has a problem when using a keyed list~~

LeCarbonator:
Arrays look to be fixed, at least confirmed for React.

However, it looks like `defaultValue` and `defaultMeta` have been broken in the process. This might also highlight stale unit tests, so I'll have to go through that thoroughly.

packages/react-form/src/useField.tsx
```diff
@@ -217,7 +217,13 @@ export function useField<
     extendedApi.Field = Field as never
 
     return extendedApi
-  })
+    // We only want to
+    // update on name changes since those are at risk of becoming stale. The field
+    // state must be up to date for the internal JSX render.
+    // The other options can freely be in `fieldApi.update`
+    // eslint-disable-next-line react-compiler/react-compiler
```
harry-whorlow: If we're leaving out dependencies then isn't the compiler skipping the optimisation of all fieldApi's

harry-whorlow:
pr works in local code base, and fixes array bug.

LeCarbonator:
I'll add a solidjs test later. After that, it's good to merge.

willbowdler-cc:
Please merge this 🙏

PR #1734, opened 2025-09-16
Update README

LadyBluenotes:
- Updating README so it's more consistent across TanStack Libraries
- Added CTA/ sponsors
- Added ecosystem links


README.md
LeCarbonator, line 1: The logos of partners are hard to read with dark mode. This also seems to happen with other packages like TanStack Router. Do you think it's fine?

image

LadyBluenotes: Yeah, I'm not quite sure how to solve this, I noticed it during the previews. I might have to see what people suggest.

LeCarbonator: Sounds good.

The router preview for reference:

image

LadyBluenotes: I'd assume the best option is to add a white background to them??? At the moment, I'm pulling most of the logos from the tanstack site / whatever was pre-existing.

Mostly didn't want to add a bunch of images to the media folders of all repos, but that may be the only option

LeCarbonator: Probably. Either way, it's not blocking this PR, so I'll merge it

PR #1746, opened 2025-09-18
docs(react-form): Fix capitalization of TanStackDevtools references in devtools docs

jameskirkby:
Replaces `TanstackDevtools` with `TanStackDevtools` in `devtools.md`, as the former is not exported from the `@tanstack/react-devtools` package.


harry-whorlow:
thanks, and good catch!  🚀

PR #1752, opened 2025-09-21
fix(core): pollyfill get random values for native environments

harry-whorlow:
adds polyfill for crypto in native environments


crutchcorn (changes requested):
This should not be the solution, as it will lead to a number of problems

Issue #1744, opened 2025-09-17
`revalidateLogic` does not work with React Native


RobertOstermann:
### Describe the bug

Describe the bug

Not sure if this should be listed as a bug or feature request. The `revalidateLogic` function used with the `onDynamic` validator doesn't seem to work with React Native. The docs indicate that TanStack Form should work out of the box with React Native, so not sure if this is a bug or just this feature does not work for React Native.

### Your minimal, reproducible example

N/A

### Steps to reproduce

- Create a React Native project
- Add the `validationLogic: revalidateLogic()` to the `useAppForm` hook
- Attempt to run the project using React Native on an IOS simulator
- Note the error: "Uncaugh Error: (0, _classPrivateFieldLooseBase2(...)_eventTarget]().addEventListener is not a function (it is undefined)

### Expected behavior

All features in TanStack Form should work in React Native. If this is expected to not work, the [docs](https://tanstack.com/form/v1/docs/framework/react/guides/dynamic-validation) on dynamic validation should mention it.

### How often does this bug happen?

Every time

### Screenshots or Videos

<img width="345" height="741" alt="Image" src="https://github.com/user-attachments/assets/99f02ce5-007d-4a88-8cd6-7134a2102476" />

### Platform

Platform
- Mac with IOS simulator

### TanStack Form adapter

react-form

### TanStack Form version

v1.21.1

### TypeScript version

v5.8.3

DvzZDev:
Same error happens to me after upgrading to SDK 54. Any fix?
(Edit: Upgrading from 1.12.1 to 1.20.0 fixed the issue for me.)

RobertOstermann:
Interesting, looks like `v1.20.0` is working, but `v1.21.1` is not for me. I can use `v1.20.0` in the meantime. Does `v1.21.1` also not work for you?

patrikduksin:
+1 on 1.21.1 + expo 54

jordankkk:
Same here (on 1.21.1) with bare react native 0.81.4, but i get ```TypeError: crypto.randomUUID is not a function (it is undefined)```, It was working on 1.20.0.

<img width="564" height="318" alt="Image" src="https://github.com/user-attachments/assets/e43fa9f9-33c6-4649-9851-4c7070314f54" />

crutchcorn:
The bug is here:

[...]

PR #1753, opened 2025-09-21
fix(core): move to @lukeed/uuid for native environments

harry-whorlow:
Draft for now so I can track it, yet to be tested.

Implements @lukeed/uuid so were not relying on crypto.getRandomValue in native environments.


harry-whorlow:
looks to fix #1744

harry-whorlow:
Somehow I missed some of my experimental code and it was included into the PR, this returns formId to a getter again

PR #1758, opened 2025-09-23
fix(core, devtools): optimise event emissions

harry-whorlow:


packages/form-devtools/src/components/DetailsPanel.tsx
```diff
@@ -8,69 +8,36 @@ import { StateHeader } from './StateHeader'
...
+    isFormValidating: state()?.isFormValidating,
+    isSubmitted: state()?.isSubmitted,
+    isSubmitting: state()?.isSubmitting,
+    isSubmitSuccessful: state()?.isSubmitSuccessful,
+    submissionAttempts: state()?.submissionAttempts,
+    errors: state()?.errors,
+    errorMap: state()?.errorMap,
+  }))
+
+  createEffect(() => console.log(selectedInstance()?.state))
```
AlemTuzlak: Hmm

Issue #1648, opened 2025-07-31
`withFieldGroup` array handling with nested


TheYoxy:
### Describe the bug

When combining the `withFieldGroup` for a complex array handling, the nested array element seems out of sync with the actual value of the group.

Also, this is unrelated but when looking at the extra props of the render function, it seems to pass the fields and the form. Idk if this is expected or if the props must be sanitized when rendering to remove theses values before passing the props to the render method. *(They are undocumented and not typed)*

### Your minimal, reproducible example

https://codesandbox.io/p/devbox/misty-monad-8t29lx?workspaceId=ws_PkVpfRWHUp5vMscgD59gVj

### Steps to reproduce

1. Try to move the value with the `move down`|`move up` buttons.
2. See that the values from the base array mapping are correct, but the values rendered in the element aren't.

### Expected behavior

I think the nested `group` value should remain sync with the array mapping from the parent `group`

### How often does this bug happen?

Every time

### Platform

- Browser: Tested on chrome 138.0.7204.183 (Official Build) (64-bit)

### TanStack Form adapter

react-form

### TanStack Form version

v1.15.1

vkhalikov:
I encountered the same issue. Looks like this happens even without nesting `withFieldGroup` components.
https://codesandbox.io/p/devbox/suspicious-elgamal-79j4tr?workspaceId=ws_MqxYqe2Tbvb7TuAH8jpm52

Seems like the reason for this is that `useFieldGroup` doesn't react to changes to `opts.fields`.
I suppose `FieldGroupApi` should have something similar to `FieldApi`'s `update` method, which will then be used in `useFieldGroup`:

```
  useIsomorphicLayoutEffect(() => {
    formLensApi.update(opts)
  })
```

@LeCarbonator can you confirm?

LeCarbonator:
I suppose that reactive `fields` should be possible and is not considered by the field group at the moment.

I was initially hesitant to check this since there's array-related issues that should be solved first before determining that this is field group related instead of a core issue of arrays.

The examples help though! I'll have a look sometime.

vkhalikov:
In case someone needs a quick solution:
Adding a key prop to a `withFieldsGroup` component when rendering mitigates the issue by forcing it to remount.

Here's an updated sandbox with side by side comparison.
https://codesandbox.io/p/devbox/condescending-wind-2p6qk4?workspaceId=ws_MqxYqe2Tbvb7TuAH8jpm52

[...]

PR #1783, opened 2025-09-30
feat(solid-form): add withFieldGroup API

Vijayabaskar56:
## 🎯 Changes

Add the `withFieldGroup` API for Form Composition in Solid Form

## ✅ Checklist

- [X] I have followed the steps in the [Contributing guide](https://github.com/TanStack/form/blob/main/CONTRIBUTING.md).
- [X] I have tested this code locally with `pnpm test:pr`.

## 🚀 Release Impact

### DIFF
--- a/packages/solid-form/src/createField.tsx
+++ b/packages/solid-form/src/createField.tsx
@@ -11,12 +11,13 @@ import type {
   DeepValue,
   FieldAsyncValidateOrFn,
   FieldValidateOrFn,
+  FieldValidators,
   FormAsyncValidateOrFn,
   FormValidateOrFn,
   Narrow,
 } from '@tanstack/form-core'
 
-import type { Accessor, Component, JSX, JSXElement } from 'solid-js'
+import type { Accessor, JSX, JSXElement } from 'solid-js'
 import type { CreateFieldOptions, CreateFieldOptionsBound } from './types'
 
 interface SolidFieldApi<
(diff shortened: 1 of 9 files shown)


LeCarbonator:
Looking good! There's one thing that I'm curious about:

#1648 lists a problem with React's implementation of `withFieldGroup`. Does this implementation also suffer this problem?

Vijayabaskar56:
Most likely, but will try to reproduce and let you know

Vijayabaskar56:
@LeCarbonator  , I have tested it , first I would thought that, the implementation is same so there is a change , but to my suprise, the bug is not presented in the solid version, then later only realized that solids signal based reactive pattern does not dependent on component state to rerender the component, and state are independent. so case in point , this #1648 issues from react version of `withFieldGroup` will not be a problem in solid version

Riddhiman007:
Wow nice, can't wait for this feature anymore

LeCarbonator:
@Vijayabaskar56 Sounds good, then! I guess this PR is not blocked by that then. We'll try to review it soon.

Vijayabaskar56:
looking forward to it.. 😁.

Riddhiman007:
When this pr will be accepted??? I'm waiting for this feature

LeCarbonator:
@Riddhiman007 It looks like the solid-form maintainers are busy. I suppose I'll look over it instead.

Keep in mind that you can [install PR versions](https://github.com/TanStack/form/pull/1783#issuecomment-3354803949) if you desperately need access to it before the merge.

packages/solid-form/src/createField.tsx
```diff
@@ -625,6 +626,94 @@ interface FieldComponentProps<
...
+    /**
+     * An optional list of field names that should trigger this field's `onChange` and `onChangeAsync` events when its value changes
+     */
+    onChangeListenTo?: DeepKeys<TLensData>[]
+    /**
+     * An optional list of field names that should trigger this field's `onBlur` and `onBlurAsync` events when its value changes
+     */
+    onBlurListenTo?: DeepKeys<TLensData>[]
+  }
+}) => JSX.Element
```
LeCarbonator, line 715: We had a React PR related to this being widened to `ReactNode`.

It needs some fixing on the React side of things because React 17 does not allow `ReactNode` in function components, but I'm not sure if there's a similar problem for Solidjs.

If so, widen the return type to allow returned values like `null` or `undefined`.

LeCarbonator (approved):
Great implementation! As far as I'm concerned, this is ready to go. Let me know your thoughts on the one comment, but it's not a blocking request.

We'll merge later this week.

packages/solid-form/src/createField.tsx
Vijayabaskar56: hey , I look into the Solidjs Element Type

`type Element = Node | ArrayElement | (string & {}) | number | boolean | null | undefined`

it already contains null or undefined type included so It wouldn't be necessary

LeCarbonator: Awesome! That's settled then. As mentioned, this is good to go! Merge will happen in the next couple days.

Vijayabaskar56: Alright thanks ✌️

LeCarbonator:
The CI wrongly interpreted the package versions. This is merged, but the release will come once #1857 is fixed and merged.

Thanks again for the SolidJS port!

Vijayabaskar56:
Thanks you for the Review @LeCarbonator

PR #1789, opened 2025-10-04
feat(devtools): update to devtools utils, add solid devtools

harry-whorlow:
updates form devtools to devtool utils and constructors, adds support for solid-form


packages/form-core/src/EventClient.ts
```diff
@@ -62,6 +62,8 @@ class FormEventClient extends EventClient<EventMap> {
   constructor() {
     super({
       pluginId: 'form-devtools',
+      debug: true,
```
samtsai, line 65: Should this be determined by environment/variable?

harry-whorlow: My bad, releasing a patch for it now. I apologise, it should of been caught.

harry-whorlow: @samtsai Fixed in `#1.25.0`, and added regression test to make sure this doesn’t happen again.

samtsai: Awesome! Thanks for the quick turn around!

Issue #1808, opened 2025-10-14
Deleted fields with defaultValue are restored on re-render after deleteField()


kusiewicz:
### Describe the bug

When calling `form.deleteField('fieldName')` on a field that has a `defaultValue` defined, the field is successfully removed from `form.state.values`. However, after React re-renders, the field **automatically reappears** with its `defaultValue`, making it impossible to truly delete fields with default values.

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-form-cnn3izz6?file=src%2Fapp.tsx

### Steps to reproduce

1. Go to the `StackBlitz` example
2. Click `remove fullname` button
3. Look on live state

### Expected behavior

After calling `form.deleteField('fieldName')`, the field should:
- Be permanently removed from `form.state.values`
- Stay removed even after React re-renders
- Not be restored by any internal lifecycle methods

### How often does this bug happen?

Every time

### Platform

everywhere

### TanStack Form adapter

None

### TanStack Form version

1.23.4

### TypeScript version

^5

LeCarbonator:
The reproduction was fixed as of `v1.25.x`

PR #1809, opened 2025-10-14
fix(form-core): fix deleteField method

kusiewicz:
## 🎯 Changes

Fixes: https://github.com/TanStack/form/issues/1808

_If this idea make sense, I will add tests_

After calling `form.deleteField('fieldName')`, the field is removed from `form.state.values`, but after a moment it returns with defaultValue and appears at the end of the values object.

The problem occurred due to React lifecycle and the logic in `FieldApi.mount()`

**Sequence of events**:
1. User clicks `delete` -> `form.deleteField('field')` executes -> value is remo ved from `form.state.values` -> field metadata is removed - cool, but
2. React detects state change -> component re-renders
3. During re-render the field is still in the DOM (not unmounted)
4. React calls `FieldApi.update()` or lifecycle hooks -> this calls `setFieldValue` 
5. In `FieldApi.mount()` there was logic:

```
   if ((this.options.defaultValue as unknown) !== undefined) {
     this.form.setFieldValue(this.name, this.options.defaultValue as never, {
       dontUpdateMeta: true,
     })
   }
```
I think this uncoditionally restored `defaultValue` on every mount.update

PS: Even if we add a check in `mount()` - `setFieldValue` can be called from other places (e.g., from `FieldApi.update()` during React lifecycle).

I added a mechanism for tracking intentionally deleted fields to prevent their their accidental restoration.

1.  Added `_deletedFields: Set<string>` in `FormApi` - it stores names of fields that were intentionally deleted by the user
2. `deleteField()` adds the field to `_deletedFields`
3. `setFieldValue()` blocks operations on deleted fields

## ✅ Checklist

- [X] I have followed the steps in the [Contributing guide](https://github.com/TanStack/form/blob/main/CONTRIBUTING.md).

## 🚀 Release Impact

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -749,7 +749,9 @@ export interface FormState<
   in out TOnDynamic extends undefined | FormValidateOrFn<TFormData>,
   in out TOnDynamicAsync extends undefined | FormAsyncValidateOrFn<TFormData>,
   in out TOnServer extends undefined | FormAsyncValidateOrFn<TFormData>,
-> extends BaseFormState<
+>
+  extends
+    BaseFormState<
       TFormData,
       TOnMount,
       TOnChange,
@@ -896,8 +898,7 @@ export class FormApi<
   in out TOnDynamicAsync extends undefined | FormAsyncValidateOrFn<TFormData>,
   in out TOnServer extends undefined | FormAsyncValidateOrFn<TFormData>,
   in out TSubmitMeta = never,
-> implements FieldManipulator<TFormData, TSubmitMeta>
-{
+> implements FieldManipulator<TFormData, TSubmitMeta> {
   /**
    * The options for the form.
    */
@@ -1191,8 +1192,8 @@ export class FormApi<
         const hasOnMountError = Boolean(
           // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
           currBaseStore.errorMap?.onMount ||
-            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-            fieldMetaValues.some((f) => f?.errorMap?.onMount),
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+          fieldMetaValues.some((f) => f?.errorMap?.onMount),
         )
 
         const isValidating = !!isFieldsValidating


LeCarbonator:
Blocked by #1729 . This seems critical to test with the linked PR.

crutchcorn:
@kusiewicz can you fix merge conflicts and then @LeCarbonator if you could, re review for @kusiewicz since the blocker is merged?

Thanks!

kusiewicz:
> @kusiewicz can you fix merge conflicts and then @LeCarbonator if you could, re review for @kusiewicz since the blocker is merged?
> 
> Thanks!

Will do today!

packages/form-core/src/FormApi.ts
```diff
@@ -2250,6 +2258,14 @@ export class FormApi<
     const dontRunListeners = opts?.dontRunListeners ?? false
     const dontValidate = opts?.dontValidate ?? false
 
+    if (this._deletedFields.has(field as string)) {
+      return
+    }
+
+    if (!opts?.dontUpdateMeta) {
+      this._deletedFields.delete(field as string)
+    }
```
LeCarbonator, line 2267: This feels like a no-op.
* If the field is in the set, you return early.
* If the field is not in the set, delete it from the set.

packages/form-core/src/FormApi.ts
```diff
@@ -2250,6 +2258,14 @@ export class FormApi<
     const dontRunListeners = opts?.dontRunListeners ?? false
     const dontValidate = opts?.dontValidate ?? false
 
+    if (this._deletedFields.has(field as string)) {
```
LeCarbonator, line 2261: Use your newly added `isFieldDeleted` instead.

kusiewicz:
> Sort of looks promising, but looking at the cause of the initial problem, it was because of the way defaultValues were handled in `fieldApi.update`. They have been changed in the recent updates due to arrays being stale.
> 
> I can't reproduce the linked issue on the latest TSF version. Can you confirm?
> 
> If the issue still exists, then I think this implementation would make sense. Adding tests to avoid regression would be needed before a merge, but let's confirm that it's still a needed fix first.

Guys, problem does not exist in the newest `main` version - I think it can be closed. 

@LeCarbonator @crutchcorn

LeCarbonator:
Well, glad to hear! Closing the PR, then.

PR #1826, opened 2025-10-26
feat(react-form): add `useTypedAppFormContext`

LeCarbonator:
## 🎯 Changes

This PR adds the `useTypedAppFormContext` hook, which accepts a `formOptions` for typing, while using context to get the form instance instead.

This is a less type safe variant of `withForm`, which is why we will only recommend it in cases where passing as prop is not possible.
## 🚀 Release Impact

## Todos
- [ ] Finalize naming of the hook
- [ ] Add docs section to address its use cases

### DIFF
--- a/.changeset/ninety-crews-smash.md
+++ b/.changeset/ninety-crews-smash.md
@@ -1,5 +1,5 @@
 ---
-'@tanstack/react-form': patch
+'@tanstack/react-form': minor
 ---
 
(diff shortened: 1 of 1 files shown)


jmn-lime:
Just wondering when we can expect this to be released?

LeCarbonator:
In the next few days! I let this one stay for too long. I'll write the documentation and add it in.

PR #1850, opened 2025-11-12
fix(react-form): remove `useId` and adjust types for React 17 compatibility

harry-whorlow:


JonnyPickard:
I was close to submitting a similar PR as this also fixes no custom form ids in devtools (which would be super nice for debugging complex multipart forms):

image

Thanks!

packages/react-form/src/useForm.tsx
```diff
@@ -182,7 +187,7 @@ export function useForm<
     TSubmitMeta
   >,
 ) {
-  const formId = useId()
+  const formId = useRef(opts?.formId ?? uuid())
```
LeCarbonator: This runs `uuid` on every render, which seems excessive. 
A one time initialization should work here.

```suggestion
  const formId = useRef<string>(opts?.formId as never);
  
  if (!formId.current) {
    formId.current =  uuid();
  }
```

LeCarbonator: Also, does this address the `querySelector` problem you encountered in previous iterations of uuid?

harry-whorlow: That was stable reference issue, but given this should be stable it should work... 

That being said, I have had some occurrences where, for one reason or another the form ID changes and I'll get a warning saying I the ID can't be found. But that was happening with useId implementation, so I'm a little confused as to if it was user error on my side or an implementation issue.

LeCarbonator: Perhaps a hydration issue?

Issue #1781, opened 2025-09-30
Devtools show null instead of NaN


Sam-Kruglov:
### Describe the bug

I set the field to `Number.NaN` but in devtools I see `null`. 

I set all number fields to `NaN` in tanstack form but I convert it to empty string when actually setting `<input value={...}` inside of my `field.MyInput` because that's what the "not set" value for all inputs is supposed to be in HTML for whatever reason, even for `type="number"`.

### Your minimal, reproducible example

too busy

### Steps to reproduce

set some number to nan in default values and observe devtools

### Expected behavior

`value: NaN`

actual: `null` (not even the `value:` part. Also for groups it'll show `value: {null,null, "someStringField": "", null}`

### How often does this bug happen?

Every time

### Screenshots or Videos

<img width="178" height="126" alt="Image" src="https://github.com/user-attachments/assets/87f4c5b0-8d52-464e-9c85-a13593a95476" />

### Platform

mac, chrome

### TanStack Form adapter

react-form

### TanStack Form version

1.23.0

### TypeScript version

5.9.2

PR #1862, opened 2025-11-18
solved Devtools show null instead of NaN #1781

KirtiGautam620:
### Summary
This PR fixes an issue where TanStack Form Devtools incorrectly showed `null`
for fields whose value was explicitly set to `Number.NaN`. This caused numeric
fields that were intentionally initialized or updated to `NaN` to appear as
`null` or even disappear in the Devtools UI, making debugging confusing.

The browser correctly treats `<input type="number">` with value `""` as "not
set", so many users (including the reporter) map `NaN -> ""` when binding input
values. However, inside the form state, TanStack Form should preserve the actual
value `NaN`. Devtools was stringifying the value and normalizing it to `null`,
leading to incorrect visualization.

---
## Problem
When a field value is `Number.NaN`, Devtools displays:

### DIFF
--- a/packages/form-devtools/src/components/DetailsPanel.tsx
+++ b/packages/form-devtools/src/components/DetailsPanel.tsx
@@ -7,6 +7,25 @@ import { StateHeader } from './StateHeader'
 
 import type { Accessor } from 'solid-js'
 
+const preprocessValue = (value: any): any => {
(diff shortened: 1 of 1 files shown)


KirtiGautam620:
@lachlancollins 
 hi! i have worked and fixed this issue. kindly have a look over it and let me know what i can change. 
Thanks.

PR #1873, opened 2025-11-25
fix(form-core): validate fields without components when setFieldValue…

techcodie:
… is called

- Fix validateField to run form-level validation for fields without field instances
- Add test cases to verify validation works for fields without components
- Resolves issue where setFieldValue didn't trigger validation for fields without <form.Field> components

Fixes: Form validation not working for fields without associated components

## ✅ Checklist

- [ ] I have followed the steps in the [Contributing guide](https://github.com/TanStack/form/blob/main/CONTRIBUTING.md).

## 🚀 Release Impact


seanwu1105:
Ah, I think this is actually fixed by https://github.com/TanStack/form/pull/2001.

PR #1876, opened 2025-11-27
fix(form-core): update pacer to pacer-lite

harry-whorlow:


packages/form-core/src/utils.ts
```diff
@@ -600,3 +601,14 @@ export function uuid(): string {
...
+export const throttleFormState = liteThrottle(
+  (form: AnyFormApi) =>
+    formEventClient.emit('form-state', {
+      id: form.formId,
+      state: form.store.state,
+    }),
+  {
+    wait: 300,
+  },
+)
```
harry-whorlow, line 617: @KevinVandy you mentioned that the throttler should be defined outside the class correct?

KevinVandy: yes, stable reference is required if you actually want it to throttle. I think this would have been the fix for everything to do with the memory leak too?

KevinVandy:
This looks perfect to me. Were there any noticable improvements with how the throttling worked that you could tell?

packages/form-core/src/utils.ts
harry-whorlow: Theres another pr open atm by @LeCarbonator, but this is one of the pieces

mhornbacher:
Hey team! are we able to get this merged into the next release?

PR #1880, opened 2025-11-28
PREACT ADAPTER: update all react tooling and use preact instead

theVedanta:
## 🎯 Changes

Preact adapter, almost as same as the react one, but with preact dependencies and eslint rules, etc.

## ✅ Checklist

## 🚀 Release Impact

Right now, it uses @tanstack/react-store, due to no preact-store package right now. But we should see that soon and get rid of all the react dependency here. You can also try to clone the preact-store package and symlink it right now for testing purposes.


theVedanta:
This now uses preact-store and also works on the `examples/simple` use-case. I will write some tests and more examples to try and break it!

Issue #1611, opened 2025-07-11
Next Server Actions don't work with next.js version 15.4


sbaechler:
### Describe the bug

After updating to 15.4-canary I keep getting a server error when I post a form. The error seems to happen when parsing the return value from the server action. It is thrown by "server-action-reducer" which is a next.js module. Apparently the mimetype form/multipart is not allowed. (My form is not multipart)

When I downgrade to 15.3 the error disappears.

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-form-c425rd6a?file=package.json

### Steps to reproduce

1. Enter a value between 8 and 12.
2. Hit Submit

### Expected behavior

Submit should work with server action validation.

### How often does this bug happen?

Every time

### Screenshots or Videos

<img width="1139" height="347" alt="Image" src="https://github.com/user-attachments/assets/e69478f7-5008-4079-947c-92788c80f397" />

### Platform

- OS: Mac OS
- Browser: Brave
- Node: 24

### TanStack Form adapter

react-form

### TanStack Form version

1.14.0

luchillo17:
Did you change the stackblitz repro? I'm not seeing the same error than in the screenshot, the error I see is due to the expected shape of the errors from the server being different from the top level form errors loop (no `issues` in `issueObject`).

<img width="365" height="146" alt="Image" src="https://github.com/user-attachments/assets/a34f586c-b51f-43a3-b49f-8a889863e5f1" />

crutchcorn:
Can you retest with:

```
npm i https://pkg.pr.new/@tanstack/react-form-nextjs@1890
```

I believe we have a fix for many (if not all) Form SSR behaviors

crutchcorn:
Should be fixed in 1.28.1

Issue #1609, opened 2025-07-08
onServerValidate doesn't work with Zod in NextJs


heinsenberg82:
### Describe the bug

`onServerValidate` doesn't seem to work properly with Zod schemas.
Instead of populating the form with error messages, the form is receiving entire objects. Here's a link that reproduces the issue: https://stackblitz.com/edit/tanstack-form-3wvkfapn?file=src%2Fapp%2Faction.ts

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-form-3wvkfapn?file=src%2Fapp%2Faction.ts

### Steps to reproduce

Go to link to reproduce the issue. Try to enter a number like '9'.

### Expected behavior

Form should be populated with error messages, not zod error objects.

### How often does this bug happen?

None

### Platform

Windows 11

### TanStack Form adapter

None

### TanStack Form version

1.14.1

heinsenberg82:
Based on this (https://github.com/TanStack/form/releases/tag/v0.43.0), I made another fork following (I think) the instructions there, but still not working: https://stackblitz.com/edit/tanstack-form-q5ym4nkq?file=src%2Fapp%2Fclient-component.tsx

tomdev10:
I have the same issue, also not 100% sure how to use `onServerValidate` with a `zod` schema, but for me the validation passes every time, even when supposed to error. Tested this with both the methods of including the schema in `onServerValidate` shown above.

crutchcorn:
Can you retest with:

```
npm i https://pkg.pr.new/@tanstack/react-form-nextjs@1890
```

I believe we have a fix for many (if not all) Form SSR behaviors

crutchcorn:
Should be fixed in 1.28.1

Issue #1571, opened 2025-06-10
First input change after failed submit ignored with Next server actions


evaogbe:
### Describe the bug

In a form using Next server actions and validation after submitting an invalid input, the next character typed into the input is not shown. The input only shows from the second character on

### Your minimal, reproducible example

https://stackblitz.com/edit/nextjs-dbp2u3qq?file=app%2Fexample-form.tsx

### Steps to reproduce

1. Press submit button
2. Type a character into the input

### Expected behavior

The first character typed should be seen in the input

### How often does this bug happen?

Every time

### Platform

- OS: MacOS
- Browser: Chrome and Firefox

### TanStack Form adapter

react-form

### TanStack Form version

1.12.2

### TypeScript version

5.2.2 and 5.8.3

Sushant-Borsania:
This is happening with the remix form as well.
https://stackblitz.com/github/tanstack/form/tree/main/examples/react/remix?embed=1&theme=light&preset=node&file=src/index.tsx

Sushant-Borsania:
Looking at the other [issue](https://github.com/TanStack/form/issues/1307) mentioned regarding the similar issue, I believe this may help short term:
```
const form = useForm({
    ...formOpts,
    transform: useTransform(
      (baseForm) => {
        if (!actionData) return baseForm;
        const { values, ...serverState } = actionData;
        return mergeForm(baseForm, serverState);
      },
      [actionData]
    ),
  });
```

crutchcorn:
Can you retest with:

```
npm i https://pkg.pr.new/@tanstack/react-form-nextjs@1890
```

I believe we have a fix for many (if not all) Form SSR behaviors

crutchcorn:
Should be fixed in 1.28.1

evaogbe:
Sorry for the late reply. Things have been crazy lately. It works with 1.28.1. Thank you!

Issue #1260, opened 2025-03-09
FieldAPI: state.meta.errors are not updating properly when using server-validations for errors


devBaunz:
### Describe the bug

Basically, the `field.state.meta.errors` (field as `AnyFieldApi`) is not properly set (i.e. differs from `form.store.state.errors`) when using transformation to properly handle server-side validations, I have a very detailed example below showing expected code (`form.tsx` accessible via `/form`) vs. the code I had to hack for this to work (`home.tsx` accessible as homepage)

The difference is basically manually parsing errors via:
```
  useEffect(() => {
    if (actionData) {
      const formErrors = form.store.state.errors;
      const realErrors = formErrors.map((error) =>
        error as unknown as {
          [K in keyof z.infer<typeof ZodSchemaServer>]: z.ZodIssue[]
        }
      );
      console.log("real-errors", JSON.stringify(realErrors));
      setAllErrors(() => {
        return realErrors.reduce((result, errorObj) => {
          Object.keys(errorObj).map(key => {
            const K = key as keyof z.infer<typeof ZodSchemaServer>;
            if (!result[K]) result[K] = [];
            if (errorObj[K]) {
              result[K].push(...errorObj[K]);
            }
          });
          return result;
        }, {} as {[K in keyof z.infer<typeof ZodSchemaServer>]: z.ZodIssue[]});
      });
    }
  }, [actionData]);
```

Reproduction: [here](https://codesandbox.io/p/devbox/fast-forest-fjwkyw)

### Your minimal, reproducible example

https://codesandbox.io/p/devbox/fast-forest-fjwkyw

### Steps to reproduce

1. Go to link
2. Run it
3. See code difference between `home.tsx` and `form.tsx`

### Expected behavior

As a user, I expect server-side-validation to not require me to manually set the errors, as a side-note, I first tried using `setFieldMeta` manually too, which also did nothing

### How often does this bug happen?

None

### Platform

Any

### TanStack Form adapter

None

### TanStack Form version

latest

### TypeScript version

latest

theVedanta:
Currently, there is only support for accessing the`formState` to get the `formErrors` and render them separately, one at a time.

```typescript
{formErrors.map(
        (error) =>
          error?.age &&
          error?.age.map((fieldError) => (
            <p key={fieldError.message} className="text-red-700">
              {fieldError?.message}
            </p>
          ))
      )}
```

There has been some discussion on discord about adding support for server-related errors in the fields as well, so we should see it soon!

[...]

PR #1890, opened 2025-12-01
Fix issues with SSR

crutchcorn:
Without introducing breaking changes this should fix a ton of issues with SSR

# TODO

- [x] Write initial `transform` test for core
- [x] Test Start no JS SSR
- [x] Test Remix no JS SSR

### DIFF
--- a/.changeset/swift-ads-warn.md
+++ b/.changeset/swift-ads-warn.md
@@ -1,11 +1,11 @@
 ---
-"@tanstack/form-core": patch
-"@tanstack/react-form-nextjs": patch
-"@tanstack/react-form-remix": patch
-"@tanstack/react-form-start": patch
-"@tanstack/react-form": patch
-"@tanstack/form-example-react-next-server-actions-zod": patch
-"@tanstack/form-example-react-next-server-actions": patch
+'@tanstack/form-core': patch
+'@tanstack/react-form-nextjs': patch
+'@tanstack/react-form-remix': patch
+'@tanstack/react-form-start': patch
+'@tanstack/react-form': patch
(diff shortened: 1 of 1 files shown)


crutchcorn:
List of related issues:

https://github.com/TanStack/form/issues/1611

https://github.com/TanStack/form/issues/1609

https://github.com/TanStack/form/issues/1571

https://github.com/TanStack/form/issues/1260

https://github.com/TanStack/form/issues/1944

packages/form-core/src/utils.ts
```diff
@@ -615,3 +615,47 @@ export const throttleFormState = liteThrottle(
...
+export function deepCopy<T>(obj: T): T {
+  if (obj === null || typeof obj !== 'object') {
+    return obj
+  }
+
+  if (obj instanceof Date) {
+    return new Date(obj.getTime()) as any
+  }
+
+  if (obj instanceof Array) {
```
LeCarbonator:
```suggestion
  if (Array.isArray(obj)) {
```

packages/react-form-nextjs/tests/useTransform.spec.tsx
```diff
@@ -0,0 +1,7 @@
+import { describe, expect, it } from 'vitest'
+
+describe('test suite', () => {
+  it('should work', () => {
+    expect(true).toBe(true)
+  })
+})
```
LeCarbonator, line 7: What's the purpose of this? It seems to be present in all 3 subpackages.

crutchcorn: Mostly just to avoid having to rip out the testing behavior in the `useTransform` otherwise

Issue #1832, opened 2025-10-30
useFieldContext reactivity broken by React Compiler and ternary operator


JamesHyot:
### Describe the bug

When using useFieldContext with React Compiler and a ternary expression to render or not an input, the reactivity of the field breaks and its values is not updated when typing.

This is caused by the Compiler changing the dependency for a rerender from "field.state.value" to "field", which is always the same object so no UI update is made.
I have no idea if this is really an issue with @tanstack/form or the React Compiler itself, though I can get why the compiler would want to memo based on "field" and not only its subproperties.

This example doesn't work (and neither does a {canBeExtended && (<input value={field.state.value} onChange={(e) => field.handleChange(e.target.value)} />)} :

```
import { useFieldContext } from './form-context.tsx'

export function TextField({ label, canBeExtended }: { label: string, canBeExtended: boolean }) {
  // The `Field` infers that it should have a `value` type of `string`
  const field = useFieldContext<string>()
  return (
    <label>
      <span>{label}</span>
			{canBeExtended ?
				(<input
					value={field.state.value}
					onChange={(e) => field.handleChange(e.target.value)}
					/>)
				: (<div className='extended'>
					<input
						value={field.state.value}
						onChange={(e) => field.handleChange(e.target.value)}
					/>
				</div>)}
    </label>
  )
}
```
[...]

PR #1893, opened 2025-12-01
Fix react compiler

crutchcorn:
This PR fixes issues with the React compiler where there were edgecases we'd hit previously.

Closes: https://github.com/TanStack/form/issues/1832

### DIFF
--- a/.changeset/social-candies-count.md
+++ b/.changeset/social-candies-count.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/react-form': patch
+'@tanstack/form-core': patch
+---
+
+Fixed issues with React Compiler


packages/react-form/src/useField.tsx
TkDodo: note: for the compiler to work, this write to a ref should be in an effect that executes on each render:


```suggestion
const optsRef = useRef(opts)
useEffect(() => {
  optsRef.current = opts
}
```

TkDodo: creating an instance in memo is risky because memo is a perf optimization, not a semantic guarantee. The best thing would probably be to write to create `fieldApi` in react state, then either update it in an effect or re-create it during render. This would also help us get rid of the `optsRef` altogether I think:

```
const [prevOptions, setPrevOptions] = useState(() => ({ form: opts.form, name: opts.name })
const [fieldApi, setFieldApi] = useState(() => new FieldApi(...))

if (reactiveOptions.form !== opts.form || reactiveOptions.name !== opts.name) {
  setFieldApi(new FieldApi(...))
  setPrevOptions({ form: opts.form, name: opts.name })
}
``` 

This is the new get-derived-state-from-props pattern described here: 

- https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
- https://react.dev/reference/react/useState#storing-information-from-previous-renders

packages/react-form/src/useField.tsx
```diff
@@ -219,16 +288,12 @@ export function useField<
...
 
     extendedApi.Field = Field as never
 
     return extendedApi
-    // We only want to
-    // update on name changes since those are at risk of becoming stale. The field
-    // state must be up to date for the internal JSX render.
-    // The other options can freely be in `fieldApi.update`
-  }, [opts.form, opts.name])
+  }, [reactiveFieldApi])
```
TkDodo: can we not combine this with the memo that returns `reactiveFieldApi` ? I think we only need `extendedFieldApi`, and the `reactiveFieldApi` is just an in between step? pseudo code:

```
const extendedFieldApi = useMemo(() => {
  const reactiveFieldApi = { ... }

  reactiveFieldApi.Field = Field as never

  return reactiveFieldApi
}
```

crutchcorn: I think that makes sense! Thanks!

packages/react-form/src/useField.tsx
crutchcorn: Would `useLayoutEffect` work okay here?

TkDodo: shouldn’t matter if we only use the ref in this component, but I think with my below suggestion, we can completely get rid of the ref unless it’s used somewhere else that I am not seeing

crutchcorn:
@TkDodo ready for re-review! Thanks for your help here

PR #1896, opened 2025-12-02
feat(form-core): add onDynamicListenTo validator option

### LINKED ISSUE #1698: onDynamicListenTo does not exist
### Describe the bug

Not sure if this should be listed as a bug or feature request. I noticed there was the recent addition of the `onDynamic` validation options, which is awesome! However, the `validators` section of a `form.Field` does not provide an `onDynamicListenTo` option to link the fields and the current props are missing JSDoc comments which are present for the other options.

N/A

### Steps to reproduce

1. Add to the validators section of a `form.Field` component
2. Note there is no `onDynamicListenTo` prop, but there is an `onDynamic` option
3. The other validator props have a `ListenTo` option to link fields
4. The new `onDynamic` options also do not have JSDoc comments describing behavior

### Expected behavior

`OnDynamic` should have an `onDynamicListenTo` prop to allow for linking fields with the `onDynamic` validation. I also would expect the `onDynamic` props to have JSDoc comments like the other fields have.

None

### Screenshots or Videos

_No response_

### Platform

- Version: 1.19.2

### TanStack Form adapter

react-form

### TanStack Form version

1.19.2

### TypeScript version

v5.8.3

### Additional context

_No response_

jiji-hoon96:
Adds support for onDynamicListenTo to enable cross-field dynamic validation.

## Changes
  - Added `onDynamicListenTo` to `FieldValidators` interface
  - Implemented dynamic cause handling in `getLinkedFields` method
  - Added unit tests for both sync and async scenarios

  ```tsx
  // Password confirmation with cross-field validation
  <form.Field
    name="confirmPassword"
    validators={{
      onDynamicListenTo: ['password'],  // ← Listen to password changes
      onDynamic: ({ value, fieldApi }) => {
        const password = fieldApi.form.getFieldValue('password')
        return value !== password ? 'Passwords must match' : undefined
      },
    }}
  />
```

  Fixes #1698

## Summary by CodeRabbit

* **New Features**
  * Added `onDynamicListenTo` for rerunning dynamic validators when specified linked fields change or blur.
  * Supports synchronous and asynchronous dynamic validation, including configurable async debounce timing.
  * Dynamic validation errors now clear after subsequent successful validation.

* **Documentation**
  * Added guidance and examples for linking fields with dynamic validators.
  * Clarified dynamic validator configuration and validation behavior.

### CODE REVIEW

File: examples/react/simple/src/index.tsx
Comment (LeCarbonator) [inline examples/react/simple/src/index.tsx:1]: I'm not sure if adding this to the simple example makes sense, since
* It's not a simple implementation
* It's unrelated to the other fields

Having linked fields as examples is a good idea though. Maybe it could be its own example? Perhaps it can be added to `large-form` instead. I'll leave it up to you.

File: packages/form-core/src/FieldApi.ts [hunk 1]
```diff
@@ -1362,6 +1366,23 @@ export class FieldApi<
...
+          errorMap: {
+            ...prev.errorMap,
+            [dynamicErrKey]: undefined,
+          },
+          errorSourceMap: {
+            ...prev.errorSourceMap,
+            [dynamicErrKey]: undefined,
+          },
+        }))
+      }
```
LeCarbonator: This seems out of place for `setValue`. Why does `onDynamic` need special behaviour in this case?

packages/form-core/src/FieldApi.ts
```diff
@@ -1638,9 +1668,25 @@ export class FieldApi<
...
+                ...prev.errorMap,
+                [dynamicErrKey]: undefined,
+              },
+              errorSourceMap: {
+                ...prev.errorSourceMap,
+                [dynamicErrKey]: undefined,
+              },
+            }))
+          }
+        }
```
LeCarbonator: Looks to be the same lines of code as the one in `setValue`, but this time it's in `validateSync` which is called in `setValue` as well.

packages/form-core/src/FieldApi.ts
```diff
@@ -1825,11 +1871,27 @@ export class FieldApi<
...
+              ...prev.errorMap,
+              [dynamicErrKey]: undefined,
+            },
+            errorSourceMap: {
+              ...prev.errorSourceMap,
+              [dynamicErrKey]: undefined,
+            },
+          }))
+        }
+      }
```
LeCarbonator: Third copy of (seemingly) the same lines of code.

packages/form-core/src/ValidationLogic.ts
```diff
@@ -193,6 +198,13 @@ export const defaultValidationLogic: ValidationLogicFn = (props) => {
         form: props.form,
       })
     }
+    case 'dynamic': {
+      // Run dynamic, server validation
+      return props.runValidation({
+        validators: [onDynamicValidator, onServerValidator],
+        form: props.form,
+      })
+    }
```
LeCarbonator, line 221: I assumed the point of `onDynamic` was that it would be dictated by `validationLogic`. However, it now needs to be addressed in `revalidateLogic`. I feel like there's some underlying issue that should be fixed, but I'll need to look at it myself to know for sure.

packages/form-core/tests/FieldApi.spec.ts
```diff
@@ -1908,6 +1908,136 @@ describe('field api', () => {
...
+      'Passwords do not match',
+    ])
+  })
+
+  it('should run onDynamicAsync on a linked field', async () => {
+    vi.useFakeTimers()
+    let resolve!: () => void
+    let promise = new Promise((r) => {
+      resolve = r as never
+    })
```
LeCarbonator: This on-demand resolving is very difficult to follow in this unit test due to how hacky it is.

I think an implementation with `vi.useFakeTimers` and a `setTimeout` + `vi.awaitAllTimersAsync` is way clearer in terms of code flow.

LeCarbonator (changes requested):
I like the idea, but the implementation needs some adjustment. There's a chance the issues come from a core problem with `onDynamic`, but it's not clear from this diff alone.

## Missing changes

~~There's also `FieldGroupApi.getFormFieldOptions` that needs to include `onDynamicListenTo`. Follow the `onChangeListenTo` / `onBlurListenTo` lines in there as reference.~~ Addressed in [89adf3a](https://github.com/TanStack/form/pull/1896/commits/89adf3ae235a44e0388a2aa4f42df2dcc8bae973)

examples/react/simple/src/index.tsx
jiji-hoon96: I believe it makes more sense to include it in `large-form` rather than as a separate example.
Linked field validation is a pattern commonly used in real-world scenarios, so it fits naturally within that example.
I'll move it to `large-form`. Is that okay with you?

packages/form-core/src/FieldApi.ts
jiji-hoon96: You're right 
I was trying to clear dynamic errors before change validation runs, but if `validate('change')` passes, it should naturally clear all errors anyway.

packages/form-core/src/FieldApi.ts
jiji-hoon96: This is duplicated. OMG

packages/form-core/src/FieldApi.ts
jiji-hoon96: OMG...

packages/form-core/tests/FieldApi.spec.ts
jiji-hoon96: I'll refactor to use `vi.useFakeTimers()` with `setTimeout` + `vi.awaitAllTimersAsync()` as you suggested. Much clearer.

packages/form-core/src/ValidationLogic.ts
jiji-hoon96: I think there's an architectural issue here too.

`onDynamic` is triggered by other fields' changes (via `onDynamicListenTo`), not by the standard validation event flow. This forces  both `defaultValidationLogic` and `revalidateLogic` to handle it specially.

@LeCarbonator  Should `onDynamic` be a first-class event type, or just metadata that says "run this validator when linked fields change"?

I'm  wait for your review on the best approach.

jiji-hoon96:
@LeCarbonator Merged `main` and worked through the review: example reverted to `main`, JSDoc added to the `onDynamic` options on both `FieldValidators` and `FormValidators` (the other half of #1698), guide entry, changeset, and the async test moved to fake timers.

**On "a core problem with `onDynamic`"**

There is one, but it is the bridge rather than `onDynamic`. This line

```ts
modifiedValidators.onChange = modifiedValidators.onDynamic
```

tells `getSyncValidatorArray` / `getAsyncValidatorArray` that a dynamic validator is a change validator, so everything keyed on `cause` downstream resolves against the wrong option. Three symptoms:

- **Timing gate bypassed.** Under `revalidateLogic()` before the first submit, the field's own change gives `{}` (gated correctly) while a linked field's change gives `{ onDynamic: 'MISMATCH' }`. Same under `defaultValidationLogic`, where `onDynamic` should never run at all.
- **`onDynamicAsyncDebounceMs` dropped**, since the debounce is also picked by `cause`. Set to 500, the validator fired on the first keystroke. Fixed in 522a5a27.
- **`onServer` clobbering `onDynamic` errors**, since the `cause = 'dynamic'` stamp hit every validator the event returned. Fixed while merging.

Two patches to one three-line remap argues for removing it rather than patching again. The obvious direction is to dispatch a real `dynamic` cause and let `validationLogic` gate it, but `revalidateLogic` gates on `[modeToWatch, 'submit'].includes(event.type)` and `modeToWatch` only ever holds `change | blur | submit`, so `dynamic` matches nothing. That is still the December question and it is your call.

**Two things before I go further**

1. Is `main` still taking features? The last `feat:` there was 2026-05-26 and everything since has gone to `alpha`. If v1 is frozen this cannot land as-is.
2. `alpha` has no `onDynamic`, `validationLogic` or `getLinkedFields`, and `triggers: [{ trigger, when }]` plus `triggerDebounceMs` and `runOnSubmit` already cover what `onDynamic` and `revalidateLogic` did together, so `onDynamicListenTo` has no v2 counterpart. What is left is `listenToFields` from your TODO in `FormApi.lib.ts`. Happy to take it on, but the open questions there (path prefixes, rebinding across trie moves, clearing `onSubmit` / `onMount`, DX) need your direction first.

PR #1913, opened 2025-12-06
fix(react-form): fix cache components support

mhornbacher:
## 🎯 Changes

Attempts to resolve #1907

Removes `Math.random()` introduced in #1893 from the hot path allowing the page to be pre-built with a a user provided form id.

source: https://nextjs.org/docs/app/getting-started/cache-components#non-deterministic-operations

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/react-form/src/useFormId.ts
+++ b/packages/react-form/src/useFormId.ts
@@ -2,4 +2,5 @@ import * as React from 'react'
 import { useUUID } from './useUUID'
 
 /** React 17 does not have the useId hook, so we use a random uuid as a fallback. */
-export const useFormId = React.version.split('.')[0] === '17' ? useUUID : React.useId
+export const useFormId =
+  React.version.split('.')[0] === '17' ? useUUID : React.useId
--- a/packages/react-form/src/useUUID.ts
+++ b/packages/react-form/src/useUUID.ts
@@ -1,5 +1,5 @@
-import { useState } from "react"
-import { uuid } from "@tanstack/form-core"
+import { useState } from 'react'
+import { uuid } from '@tanstack/form-core'
 
 /** Generates a random UUID. and returns a stable reference to it. */
 export function useUUID() {


mhornbacher:
@crutchcorn ah I must have misunderstood the PR instructions. I will update the changeset file :)

crutchcorn:
S'all good :) Changeset is a tricky tool to get used to

mhornbacher:
Thank you so much for the pointer. Please let me know if changing it to patch is enough or if the file should be fully removed. It's my first contribution here and I am learning the ropes ;)

Also you mentioned that you are unclear if this is expected behavior either. I am not sure what that means...

I think the proper fix for this is to use `useId` which generates a consistent id based on the react component hierarchy and works with streaming. But that would break react 17 compatibility which is a major unjustified breaking change to support the cutting edge of only one, however popular, meta-framework (Next.JS 16).

crutchcorn:
> I think the proper fix for this is to use `useId`

I think you're right...

> But that would break react 17 compatibility

And also right here. What if we detect `React.version` in a custom `useFormId` hook and conditionally return `useId` or our `uuid()` implementation?

That shouldn't break the rules of React hook, especially if we do that replacement like so:

```tsx
import {version as reactVersion, useId} from "react";

function useUUID() {
    /* UUID code here */
}

const useFormId = Number(reactVersion.split(".")[0]) > 17 ? useUUID : useId;
```

WDYT?

<details>
<summary>Aside about contributing</summary>

> It's my first contribution here and I am learning the ropes

For the record, even if this PR isn't merged exactly as-is, please don't hesitate to find other tickets and/or reach out to be mentored (by me!) on how to continue contributing. We always enjoy and welcome new contributors.

</details>

packages/react-form/src/useId.ts
```diff
@@ -0,0 +1,13 @@
+import { version as reactVersion, useId as useReactId } from 'react'
+import { uuid } from '@tanstack/form-core'
+
+/** React 17 does not have the useId hook, so we use a random uuid as a fallback. */
+export function useId() {
+  if (reactVersion.split('.')[0] === '17') {
```
crutchcorn: I'd rather not do this check at Hook-time, as it feels brittle. Can we switch to using a top-level check like I'd posted? :)

packages/react-form/src/useId.ts
```diff
@@ -0,0 +1,13 @@
+import { version as reactVersion, useId as useReactId } from 'react'
+import { uuid } from '@tanstack/form-core'
+
+/** React 17 does not have the useId hook, so we use a random uuid as a fallback. */
+export function useId() {
```
crutchcorn: Also, can we rename this to `useFormId` just to avoid confusion when reading code?

crutchcorn (changes requested):

packages/react-form/src/useId.ts
mhornbacher: 💯 of course. that difference fully slipped my mind

crutchcorn:
@mhornbacher wanna test this in your project before we go live? :) 

Check the install commands here: https://github.com/TanStack/form/pull/1913#issuecomment-3621098936

Or by using:

```shell
npm i https://pkg.pr.new/@tanstack/react-form@1913
```

And let me know if this fixed the bug. If it did, we'll merge and release ASAP

mhornbacher:
@crutchcorn yes. Sorry it took a minute, I wanted to isolate other possible causes but I can confirm that this does _**not**_ fix the issue

Testing component (sorry for the ugly code)

I will have to dig deeper to find the cause. Are you ok if I use the CI pipeline here to push some experiments to try and isolate it? (do you pay for CI runs/package hosting or is it covered in a GitHub/StackBlitz OSS tier)

```jsx
"use client";

import { useForm } from "@tanstack/react-form";

export function ForgotForm() {
  const form = useForm({}); // commenting out this line enables pre-rendering.
  return <>{'hi'}</>
}
```

Before and after removing `const form = useForm({})`

Screenshot 2025-12-06 at 4 10 57 PM
Screenshot 2025-12-06 at 4 11 39 PM
Screenshot 2025-12-06 at 4 13 11 PM

crutchcorn:
Glad we were able to confirm that this doesn't fix - that's better than pushing and being off-base.

> Are you ok if I use the CI pipeline here to push some experiments to try and isolate it? (do you pay for CI runs/package hosting or is it covered in a GitHub/StackBlitz OSS tier)

Go for it! We don't pay for hosting, we're covered by OSS.

If you need any help debugging, let us know in the Discord and we'll help however we can :)

mhornbacher:
@crutchcorn I have not found the cause yet but I was actually able to confirm that `uuid` does not cause this at all (copied the code into my project and ran it)

However it does cause an annoying hydration issue and `useId` solves that :)

I went through each execution line in a debugger and am fully lost as to what could possibly be causing this. maybe a set of fresh eyes tomorrow will make it clear :(

crutchcorn:
Makes sense. I'm game to merge to fix the hydration issues.

Need anything from us before we merge for that purpose?

EDIT: I realized a problem. We can't import useId from React. We need to do React.useId and import React via `import * from React` otherwise we'll get runtime problems in 17

crutchcorn (changes requested):
Sorry about so many reviews 😭

Noticed we need to make a new useUUID hook that returns a stable reference on 17, otherwise we'll have bugs.

mhornbacher:
Its all good. I do want to try and help fix the root problem if I can. I am more of an application and infrastructure engineer than a library maintainer so I am still getting used to the concerns of library development. Its a different world and its fun :)

For the project I am building it is really a non-issue because we can wrap the pages that need a form in a `use-cache` directive or a `<Suspense />` tag is fine as the app is basically full of dynamic or per-user authenticated content anyway.

mhornbacher:
> Sorry about so many reviews 😭
> 
> Noticed we need to make a new useUUID hook that returns a stable reference, otherwise we'll have bugs.

No problem. The original `useState` implementation should work right?

crutchcorn:
Yup! `useState` impl should be fine.

And agreed haha. Library code is very different but rewarding

mhornbacher:
I pushed a `useMemo` implementation (just seemed more correct than using a state that's setter is discarded).

I'll update it to the `useState` implementation now :)

crutchcorn:
Just for the future, useMemo is not guaranteed to run only once. `useState` with a function, on the other hand, is. :) That's why we went with it instead

mhornbacher:
Oh, I did not know that. Thank you!

mhornbacher:
I got somewhere! I setup a minimal reproduction within the `monorepo` and got it to spit out the exact function causing the issue. `crypto.randomUUID()` is called somewhere in `throttle`...

Sorry I am just overly excited

crutchcorn:
Oh! Interesting! We can't have crypto calls in our codebase because we need to support React Native as well, just FYI. 

That is to say that when you find it, hunt it down and kill it with prejudice haha

mhornbacher:
Found it here in https://github.com/TanStack/pacer: https://github.com/TanStack/pacer/blob/6238bcf592edf01bb188ea0b412026692fc24513/packages/pacer/src/async-retryer.ts#L365-L383

It is related to the DevTools so likely not running in React Native at all :)

What is strange is that this is in the `asyncRetry` which _should_ not be being called from the `throttle` function... 🤔

mhornbacher:
Verified that removing the import from `@tanstack/pacer` resolves the issue fully.

Does it make sense to remove it and replace it with a simple `debounce` to avoid bringing in `crypto`?

crutchcorn:
Let's tag in @KevinVandy who maintains Pacer, so we can patch upstream :)

crutchcorn:
Let's tag in @KevinVandy who maintains Pacer, so we can patch upstream :)

KevinVandy:
This was fixed in Pacer a couple weeks ago. I thought @harry-whorlow might be updating soon? 

Also, TanStack Form should install pacer-lite instead of pacer now for smaller bundle size.

harry-whorlow:
I have an open branch amt, just need to push it... Probably in the morning

mhornbacher:
@KevinVandy nice! I opened a PR to remove it entirely as its only used in one spot for dev-tools but it looks like you guys are on it!

https://github.com/TanStack/form/pull/1916

harry-whorlow:
Pacer is updated in [1876](https://github.com/TanStack/form/pull/1876) to pacer lite. 

What was the issue with changesets?

harry-whorlow:
>I think the proper fix for this is to use `useId`

Yeah, this breaks react 17, hence why we implemented our own uuid

> And also right here. What if we detect `React.version` in a custom `useFormId` hook and conditionally return `useId` or our `uuid()` implementation?

Crazy, I didn’t realise you could import the react versions. I like this solution.

As for the changeset it looks like the perhaps the cli steps were misunderstood, since I can't see anything that would make Changesets bump a major. But yeah a patch version would suffice 😄

mhornbacher:
@harry-whorlow neither did I. @crutchcorn is a legend!

Do you need me to close this and open a new PR to flush out the major version bump?

crutchcorn:
@harry-whorlow the issue with patchsets have been updated since my comment :) A bit of back-n-forth yesterday, especially cuz I was on mobile out on the town reviewing.

@mhornbacher nope! This looks great to me; merging now after workflows pass.

crutchcorn:
This is merged into `main`! Now we'll wait for @harry-whorlow to merge their PR to `pacer-lite` and we'll cut a release

mhornbacher:
Thank you guys so very much. Your dedication to this project and its performance for the wider web is above and beyond what anyone could imagine to ask for ❤️

PR #1927, opened 2025-12-11
chore(devtools): bump devtools versions

harry-whorlow:
bumps event client disabling events on the server


packages/form-devtools/package.json
```diff
@@ -53,7 +53,7 @@
   ],
   "dependencies": {
     "@tanstack/devtools-ui": "^0.4.4",
```
LeCarbonator, line 55: No bump for devtools-ui?

harry-whorlow: 0.4.4 is latest

Screenshot 2026-01-05 at 12 05 28

PR #1930, opened 2025-12-12
fix(react-form): prevent array field re-render when child property changes

### LINKED ISSUE #1925: Entire `array` Field rerenders when a property on any array element is mutated (regression in 1.27.0)
### Describe the bug

Starting from tag 1.27.0, when working with a `form.Field` that uses `mode=array`, when any child Field under the array Field is changed (resulting in a modification to a property on one of the array's elements, not to the ordering/size of the array), the entire array Field rerenders.

This is not what happened prior to 1.27.0; in older releases, only the child Field that was touched would re-render (along with anything subscribing to that Field's backing data), as one would expect.

I noticed this in my own application but it even happens in the simple demo on the docs site: https://tanstack.com/form/latest/docs/framework/react/examples/array?panel=sandbox

The result is that in large complex forms that utilize array fields to show many editable/reorderable field groups, performance is quite poor when users type into any field.

https://codesandbox.io/p/devbox/github/tanstack/form/tree/main/examples/react/array?embed=1&theme=dark

### Steps to reproduce

1. Open devtools and enable rerender highlighting
2. Click "Add Person" three times
3. Focus Name field for middle person
4. Enter a few characters
5. Observe the entire array field rerenders, not just the single name field being modified
  - The profiler will report "Hook 5 changed" which appears to be the useSyncExternalStore containing the entire array. Array identity is changing?

### Expected behavior

When modifying a child `form.Field` mounted under a `form.Field` of `mode="array"`, only the child field should re-render when it changes. The entire `array` `form.Field` container should not re-render.

Every time

### Screenshots or Videos

Image

### Platform

- OS: Linux
- Browser: Firefox 145.0.1
- React: 19.2.0, 19.2.1

### TanStack Form adapter

react-form

### TanStack Form version

Observed in 1.27.0 / 1.27.2

Does not occur in 1.26.0

### TypeScript version

_No response_

### Additional context

- Does not seem to depend on...
  - whether the form uses a zod/standard schema validator
  - whether `onBlur` vs `onChange` validators are used
  - whether the array Field is using `withFieldGroup` or `withForm` composition
  - whether React Compiler is being used

I am using React Compiler in my application. However the demo application on the Tanstack Form docs site does not seem to be using Compiler and exhibits the same behavior.

tt-a1i:
## Summary

- Fix array fields with `mode="array"` incorrectly re-rendering when a property on any array element is mutated
- This was a regression introduced in v1.27.0 by the React Compiler compatibility changes (#1893)

## Problem

As reported in #1925, starting from v1.27.0, when working with a `form.Field` that uses `mode="array"`, when any child Field under the array Field is changed, the entire array Field re-renders.

### Root Cause

In the React Compiler fix (#1893), a new `reactiveStateValue` subscription was added:

```typescript
const reactiveStateValue = useStore(fieldApi.store, (state) => state.value)
```

This subscribes to the entire `state.value`. For array fields, when any child property changes, the array reference changes, causing `reactiveStateValue` to change, which triggers `extendedFieldApi` recreation and component re-render.

### Solution

Make the `reactiveStateValue` subscription aware of `mode`:

```typescript
const reactiveStateValue = useStore(
  fieldApi.store,
  (opts.mode === 'array'
    ? (state) => Object.keys((state.value as unknown) ?? []).length
    : (state) => state.value) as (state: typeof fieldApi.state) => TData | number,
)
```

For `mode="array"`:
- `reactiveStateValue` now tracks only the array length
- The `state.value` getter returns the actual value from `fieldApi.state.value`
- This ensures re-renders only occur when items are added/removed, not when properties change

## Test Plan

- [x] Added test case: `should not rerender array field when child field value changes`
- [x] All existing tests pass (79 tests in react-form, 403 tests in form-core)

Fixes #1925

### DIFF
--- a/.changeset/fix-array-field-rerender-regression.md
+++ b/.changeset/fix-array-field-rerender-regression.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/react-form": patch
+'@tanstack/react-form': patch
 ---
 
 fix(react-form): prevent array field re-render when child property changes


codingedgar:
heavely depended on this behaviour, any way to bypass?

Issue #1932, opened 2025-12-13
🐛 File input field: Picking a different file does not trigger rerender (deep equality issue with File/Blob objects)


UdaraWanasinghe:
### Describe the bug

When using a file input with TanStack Form (react-form), if a user picks a different file (even with different content and lastModified/name), the internal form state updates, but the field component does not rerender. The new file appears in the internal store, but the field.state.value and field render output remain unchanged. This makes it impossible to show the newly picked file in the UI. This seems due to TanStack Form's deep equality check (evaluate function) which for File/Blob objects treats different files as equal, since their own enumerable properties are empty. Object.keys(file) is []. As a result, the derived state is not replaced and UI updates are skipped.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/923ck3

### Steps to reproduce

1. Create a Field configured for a file (avatar/image upload) in react-form
2. Select any file via the browser file picker
3. Select a different file (with a different name/content/lastModified)
4. Observe: field.handleChange is called, internal state updates, but the form.Field component does not rerender.
5. If you wrap the File in an envelope (e.g., {file, _uid: Date.now()}), rerenders work as expected

### Expected behavior

As a user, I expect picking a different File object to update the field's value, trigger a rerender, and display the new file in the UI. 

### Platform

- OS: Windows, macOS (any)
- Browser: Chrome, latest

### TanStack Form version

v4.0.0 (latest at time of writing)

### TypeScript version

v5.4.4

### Additional context

Deep equality function (evaluate) relies on Object.keys for deep objects, but File/Blob exposes properties only via prototype getters, so two File instances are treated as equal if no envelope is used. Other primitive or array types work as expected. Suggested fix: detect File/Blob specially and fall back to reference/identity check, or treat any non-enumerable object without primitive valueOf as unequal from another instance.

UdaraWanasinghe:
Temporary fix

```typescript
<input
  type="file"
  multiple={false}
  onChange={(e) => {
    const file = e.target.files?.[0];
    if (file) {
      Object.defineProperties(file, {
        name: {
          value: file.name,
          enumerable: true,
        },
        size: {
          value: file.size,
          enumerable: true,
        },
        type: {
          value: file.type,
          enumerable: true,
        },
      });

      field.handleChange(file as any);
    }
  }}
/>
```

joaoGabriel55:
I am working on that! Thanks @UdaraWanasinghe for the temporary fix.

crutchcorn:
[...]

PR #1939, opened 2025-12-15
fix(form-core): 🐛 File input field - Picking a different file does not trigger rerender (deep equality issue with File/Blob objects)

joaoGabriel55:
## 🎯 Changes

Issue: #1932

**Fix:** File input field — picking a different file does not trigger rerender due to deep equality issue with `File`/`Blob` objects.

**Root cause:** `Object.keys(new File(...))` returns `[]` (empty array). The `evaluate()` function compares objects by iterating over `Object.keys`, so two different `File` instances with zero own enumerable keys are considered "equal" — the for-loop vacuously succeeds. This means `setFieldValue` with a new `File` doesn't trigger a state change.

**Fix:** Add a `Blob` guard in `evaluate()` (placed after the `Object.is` referential check and null checks, but before the generic key comparison). Since `File extends Blob`, this catches both types. If two `Blob`/`File` objects are the same reference, `Object.is` already returns `true` upstream. If they're different references, we return `false` — correctly triggering a rerender.

**React Native compatibility:** The guard uses `typeof Blob !== 'undefined'` to prevent `ReferenceError` in environments where `Blob` is not defined. If `Blob` doesn't exist, the check is skipped entirely — which is correct since no `Blob` objects can exist in that environment either.

**Changes:**
- `evaluate()` in `utils.ts`: added `Blob` instance guard (+12 lines)
- `utils.spec.ts`: added `evaluate` tests for `File` and `Blob` objects (same ref, different refs, nested in objects)
- `FormApi.spec.ts`: added integration test for `setFieldValue` with `File` values

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * File and Blob comparisons in form change detection now use referential identity instead of key-based comparison, preventing false change events and preserving File/Blob instance semantics in form fields.

* **Tests**
  * Added tests for File-type field updates and detailed File/Blob equality scenarios, covering same-reference and distinct-instance behaviors within fields and nested structures.

* **Chores**
  * Recorded a patch-level changeset noting the fix.


packages/form-core/tests/FormApi.spec.ts
```diff
@@ -4096,6 +4096,37 @@ it('should generate a formId if not provided', () => {
...
+  const firstFile = new File(['first'], 'first.png', { type: 'image/png' })
+  form.setFieldValue('avatar', firstFile)
+
+  const firstValue = form.state.values.avatar as { file: File; uuid: string }
+
+  expect(firstValue).toBeDefined()
+  expect(firstValue.file instanceof File).toBe(true)
+  expect(firstValue.file.name).toBe('first.png')
+  expect(typeof firstValue.uuid).toBe('string')
+  expect(firstValue.uuid.length > 0).toBe(true)
```
fernandollisboa:
```suggestion
  expect(firstValue.uuid.length).toBeGreaterThan(0)
```

packages/form-core/tests/FormApi.spec.ts
```diff
@@ -4096,6 +4096,37 @@ it('should generate a formId if not provided', () => {
...
+
+  const secondFile = new File(['second'], 'second.png', { type: 'image/png' })
+  form.setFieldValue('avatar', secondFile)
+
+  const secondValue = form.state.values.avatar as { file: File; uuid: string }
+
+  expect(secondValue.file.name).toBe('second.png')
+  expect(typeof secondValue.uuid).toBe('string')
+  expect(secondValue.uuid.length > 0).toBe(true)
```
fernandollisboa:
```suggestion
  expect(secondValue.uuid.length).toBeGreaterThan(0)
```

packages/form-core/tests/FormApi.spec.ts
```diff
@@ -4096,6 +4096,37 @@ it('should generate a formId if not provided', () => {
   expect(form.formId.length).toBeGreaterThan(1)
 })
 
+it('it should set a file value with uuid when setting a field value with File', () => {
```
fernandollisboa:
```suggestion
it('should set a file value with uuid when setting a field value with File', () => {
```

packages/form-core/src/utils.ts
```diff
@@ -48,11 +48,16 @@ export function getBy(obj: unknown, path: string | (string | number)[]): any {
  * @private
  */
 export function setBy(obj: any, _path: any, updater: Updater<any>) {
+  const isFile = updater instanceof File
```
LeCarbonator: React Native does not have `File` defined and will throw a `ReferenceError`. You cannot implement the fix this way.

LeCarbonator (changes requested):

packages/form-core/src/utils.ts
joaoGabriel55: Oh. Truly. Good catch, thanks!

joaoGabriel55: Oh. Truly. Good catch, thanks!

packages/form-core/src/utils.ts
```diff
@@ -52,7 +66,10 @@ export function setBy(obj: any, _path: any, updater: Updater<any>) {
 
   function doSet(parent?: any): any {
     if (!path.length) {
-      return functionalUpdate(updater, parent)
+      return functionalUpdate(
+        isFile(updater) ? { file: updater, uuid: uuid() } : updater,
```
LeCarbonator: This isn't type safe. This wrongly asserts File field values as `{ file: File | (oldFile: File) => File, uuid: string }` instead of
1. actually running the updater
2. Preserving the expected field value's structure

packages/form-core/src/utils.ts
```diff
@@ -422,6 +439,15 @@ export const isGlobalFormValidationError = (
 }
 
 export function evaluate<T>(objA: T, objB: T) {
+  if (objA instanceof File && objB instanceof File) {
```
LeCarbonator: It looks like the previous comment wasn't addressed, despite the request for another review. This will throw `ReferenceError`s in environments where `File` is not present (such as React Native).

packages/form-core/tests/FormApi.spec.ts
```diff
@@ -4096,6 +4096,37 @@ it('should generate a formId if not provided', () => {
...
+      avatar: undefined,
+    } as { avatar: unknown },
+  })
+
+  form.mount()
+
+  const firstFile = new File(['first'], 'first.png', { type: 'image/png' })
+  form.setFieldValue('avatar', firstFile)
+
+  const firstValue = form.state.values.avatar as { file: File; uuid: string }
```
LeCarbonator: These type assertions are a huge problem. It cannot stay.

Luckily, as long as `evaluate` properly handles `File` types, this whole uuid trickery won't be needed. Just make sure to add unit tests for a File comparison with `evaluate`. I believe the existing tests can be found in `utils.spec.ts`.

LeCarbonator (changes requested):
It looks like the previous comment hasn't been addressed between the previous review and this review request. 

The actual required changes don't look that bad, though. Perhaps I'll have some time to help out later this week with that.

joaoGabriel55:
Sorry for the delay with that @LeCarbonator

crutchcorn:
Closing, as https://github.com/TanStack/form/pull/2140 seems to fix the same issues but is handled more generically. Thank you for the genuinely thoughtful PR, however!

PR #1954, opened 2025-12-25
feat(form-core): delay onMount listeners until defaultValues are present

ws-rush:
This PR modifies the behavior of onMount listeners and validators in FormApi and FieldApi to better support async initial values.

Previously, onMount listeners and validators ran immediately upon mounting, regardless of whether defaultValues were available. This often led to premature validation errors or side effects when fetching data asynchronously.

With this change:

onMount listeners and validators (both at the form and field level) are delayed if defaultValues is undefined or null.
They will automatically execute once defaultValues becomes defined (e.g., when data finishes loading).
This allows developers to initialize useForm immediately but defer validation until data is ready, simplifying the handling of loading states.
Internal changes:

Added _hasMounted and _isMounted flags to FormApi and FieldApi to track initialization state and prevent duplicate execution.
Updated mount and update methods in both APIs to respect the presence of defaultValues.
Updated documentation with an example of handling async initial values.
✅ Checklist
I have followed the steps in the [Contributing guide](https://github.com/TanStack/form/blob/main/CONTRIBUTING.md).
I have tested this code locally with pnpm test:pr.
🚀 Release Impact
This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
This change is docs/CI/dev-only (no release).

### DIFF
--- a/packages/react-form/tests/useFormDelayed.test.tsx
+++ b/packages/react-form/tests/useFormDelayed.test.tsx
@@ -1,4 +1,3 @@
-
 import { render, waitFor, screen } from '@testing-library/react'
 import React, { useEffect, useState } from 'react'
 import { useForm } from '../src/useForm'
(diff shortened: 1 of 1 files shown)


ws-rush:
WIP, I submitted PR to check it in real projects using pkg.new

ws-rush:
I think we have a bigger issue related to `onMount` in versions above 126.0, I will report them after make stackblitz to ensure

LeCarbonator:
Let us know what you'll find. However, since there seems to be a lot of deleted tests, I'll mark the PR as draft.

PR #1963, opened 2025-12-31
fix(vue-form): prevent full array re-renders in array mode

### LINKED ISSUE #1961: Entrire array elements are re-rendered when updating one array element
### Describe the bug

It seems like when you use array fields in TanStack Form with vue, it re-renders each element in the list when any of the array elements is changed. It doesn't slow down the performance in case you have one input field, but once your form has more items, it becomes extremely slow.

https://stackblitz.com/edit/tanstack-form-cuwgrt8f?file=src%2Fmain.ts

### Steps to reproduce

1. Open the repro repo
2. Click the "Open preview in a new tab" button (it is needed to see Vue-scan changes, they won't be visible when the view is split.
3. Add a few fields to the form
4. Type any value in any input
5. See that all items are updated in the list, when changes happens only in one item

### Expected behavior

As a user, I would expect that only the changed fields are re-rendered, while in actual behaviour in re-renders all fields

Every time

### Screenshots or Videos

_No response_

### Platform

- **OS**: MacOS
- **Browser**: Chrome
- **Version**: 143.0.7499.170 (Official Build) (arm64)

### TanStack Form adapter

vue-form

### TanStack Form version

1.27.7

### TypeScript version

5.8.2

### Additional context

_No response_

imramkrishna:
Closes #1961

## Problem
In the Vue adapter, `useField` currently subscribes to the entire field state via:

```ts
useStore(fieldApi.store, (state) => state)

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Improvements**
  * Reduced unnecessary re-renders by refining how field state is tracked; array-mode fields now track length for efficient updates.
  * Field meta (touched, blurred, dirty, errors, validating) updates independently for more accurate UI feedback.

* **Tests**
  * Added a test verifying array-mode fields update and render correctly when pushed.

* **Documentation**
  * Updated Vue guides to explicitly show mode="array" for array fields.

* **Examples**
  * Example app updated to use array-mode field configuration.

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/TanStack/form/pull/1963)
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


imramkrishna:
CI is green and coverage is fixed.
This PR is ready for review.

crutchcorn:
Wait, I'm confused - shouldn't this be avoided as-of https://github.com/TanStack/form/releases/tag/%40tanstack%2Fvue-form%401.27.7 ?

imramkrishna:
I checked the 1.27.7 changes and this does not reintroduce the behavior avoided there.

The change in 1.27.7 avoids subscribing to the full field state to prevent unnecessary updates.
This PR follows the same intent by narrowing subscriptions further in array mode, limiting value tracking to length-only while still respecting the current store architecture.

If there’s a specific concern with multiple useStore subscriptions in Vue, I’m happy to adjust the approach.

msanchezdev:
I see this mentions Vue but the behavior also happens in react

mkeyy0:
@crutchcorn Yeah, it should be avoided, but while it is fixed in React, in Vue, it still re-renders each field after each change in any array field that significantly affects the performance. You could see the repro from the linked issue

crutchcorn:
Good catch, all. I've updated the code to reflect the React adapter better and updated the Vue array docs to include `mode`. Merging now. Might wait a moment to release so I can look at the other adapters and try to fix there as well.

Thanks for your patience with us!

PR #1971, opened 2026-01-03
fix(form-core): set isValidating synchronously before debounce delay

### LINKED ISSUE #1833: when debouceAsync is present, isValidating should still be true in time between sync finishing, and async starting
### Describe the bug

Im trying to make a form, where part of the validation is done on server, so im using both, onChange and onChangeAsync, the later with onChangeAsyncDemounceMs set to 500. 

I noticed that in the time frame between onChange finishing and OnChangeAsync starting I do not have a way to disable submit button, as isValidating is set to true after onChange

https://stackblitz.com/edit/tanstack-form-db1bcvv7?file=src%2Findex.tsx

### Steps to reproduce

1. fill firstname with 'er" - button is dissabled
2. fill firstname with 'error'- button is enabled (onChange passed) 
and after 500 s (the onChangeAsync fires) the button is disabled again

### Expected behavior

When any debouceMs is set, the isValidating should still stay as true, or we should get additional property which would represent that period of time.

Every time

### Screenshots or Videos

_No response_

### Platform

MacOs, chrome

### TanStack Form adapter

react-form

### TanStack Form version

1.23.7, 1.23.8

### TypeScript version

_No response_

### Additional context

_No response_

tt-a1i:
## Summary

When async validation is scheduled with debounce, the `isValidating` flag was being set AFTER an `await`, causing a gap where callers would see `isValidating: false` during the debounce period. This caused `form.canSubmit` to be `true` when it should be `false`.

**Before (bug):**
1. User types value that passes sync validation
2. Async validation is scheduled with 500ms debounce
3. `isValidating` remains `false` during the debounce period ← BUG
4. Submit button is enabled when it shouldn't be

**After (fix):**
1. User types value that passes sync validation  
2. Async validation is scheduled with 500ms debounce
3. `isValidating` is set to `true` immediately (synchronously)
4. Submit button is correctly disabled

## Changes

- Move `isValidating = true` assignment to happen synchronously, before any `await`, when async validators are detected
- Add test case to verify `isValidating` is true during debounce period

## Test plan

- [x] Added unit test for the specific scenario (sync passes, async debounced)
- [x] All 363 existing form-core tests pass
- [x] TypeScript compiles without errors

Fixes #1833

## AI Assistance Disclosure

I used Codex to review the changes, sanity-check the implementation against existing patterns, and help spot potential edge cases.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -1782,8 +1782,13 @@ export class FieldApi<
     )
     const hasAsyncValidators = hasOwnAsyncValidators || hasLinkedAsyncValidators
 
-    // Set isValidating for linked fields (own field already set above before await)
+    // Set isValidating for linked fields and current field when linked fields are validating
+    // This preserves original behavior where current field shows validating
+    // when any of its linked fields are validating
     if (hasLinkedAsyncValidators) {
+      if (!this.state.meta.isValidating) {
+        this.setMeta((prev) => ({ ...prev, isValidating: true }))
+      }
       for (const linkedField of linkedFields) {
         linkedField.setMeta((prev) => ({ ...prev, isValidating: true }))
       }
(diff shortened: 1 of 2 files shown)


LeCarbonator:
Hmmm ... I think the proposal makes sense. I'll do some digging to see if this is unintentional or not.

Thanks for the PR!

PR #1987, opened 2026-01-09
fix(form-core): Prevent synchronous validators from returning Promises

mhuggins:
## 🎯 Changes

Passing an asynchronous function to a synchronous validator (i.e.: `onBlur`, `onChange`, `onDynamic`, `onSubmit`) is a bit of a foot-gun given that it does not produce any typescript errors, but it also results in the form/field validation function running after the core validation logic.

This updates the types for these validator functions on both FormApi and FieldApi to prevent passing a function that returns a Promise.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Synchronous validators now enforce stricter type-checking to prevent returning Promises, catching incorrect usage patterns at compile-time and improving developer experience.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/form/pull/1987)


mhuggins:
Just pushed an update that should address the lint issue, sorry about that!

packages/form-core/src/util-types.ts
```diff
@@ -3,6 +3,23 @@
...
+ */
+export type RejectPromiseValidator<T> = T extends (...args: any[]) => infer R
+  ? // Check if R is `any` - if so, allow it (common with mocks)
+    0 extends 1 & R
+    ? T
+    : // Otherwise, reject if it's a Promise
+      [R] extends [Promise<any> | PromiseLike<any>]
+      ? never
+      : T
+  : T
```
LeCarbonator, line 21: Is this truly needed? Can the return type not be `Omit<T, Promise<any> | PromiseLike<any>>`?

mhuggins: I'm assuming you mean `Exclude` rather than `Omit`, but no that does not work here.  I believe it is a limitation of TypeScript, but the type is still treated as `unknown` it seems even if attempting to approach it from this angle.

mhuggins: FWIW, this is how the `type-fest` package accomplish an "is any" check as well: https://github.com/sindresorhus/type-fest/blob/main/source/is-any.d.ts

mhuggins: I could leverage `NoInfer` here instead of wrapping `R` in an array if preferred, however:

```suggestion
export type RejectPromiseValidator<T> = T extends (...args: any[]) => infer R
  ? // Check if R is `any` - if so, allow it (common with mocks)
    0 extends 1 & R
    ? T
    : // Otherwise, reject if it's a Promise
      NoInfer<R> extends Promise<any> | PromiseLike<any> ? never : T
  : T
```

mhuggins:
Just wanted to follow up to see if there's anything that's needed for this to be considered for acceptance.  Thanks! 😄

LeCarbonator:
Core concept looks good to me. I'm mostly waiting to check up on some things like instantiation count and testing some surface-level alternatives. Should be done by the end of the week.

mhuggins:
@LeCarbonator Just wanted to follow up to see if there's anything else needed here.  I can rebase my branch, just let me know!

crutchcorn (approved):
This is awesome. Sorry for missing it on my end. Got @LeCarbonator's approval privately as well. Thank you for this!

mhuggins:
Awesome, thanks for getting this merged!

PR #1997, opened 2026-01-16
fix(form-core): Resolve onSubmitInvalid not being called with invalid default values

### LINKED ISSUE #1990: `onSubmitInvalid` not called when `canSubmit` is false (v1.27.7)
### Describe the bug

`onSubmitInvalid` callback is not being called when the form is invalid and `handleSubmit()` is invoked, even though `canSubmit` is `false` and `isValid` is `false`.

This appears to be the same issue as #1696, which was supposedly fixed in PR #1697 (merged September 30, 2025). However, the issue persists in v1.27.7.

```tsx
import { useForm } from '@tanstack/react-form'
import { zodValidator } from '@tanstack/zod-form-adapter'
import { z } from 'zod'

const schema = z.object({
  value: z.number().min(1, 'Value must be greater than 0'),
})

function MyForm() {
  const form = useForm({
    defaultValues: { value: 0 },
    validators: {
      onChange: schema,
      onSubmit: schema,
    },
    onSubmitInvalid: () => {
      console.log('onSubmitInvalid called') // This is NEVER called
    },
    onSubmit: ({ value }) => {
      console.log('onSubmit called', value)
    },
  })

  return (
    <button
      onClick={() => {
        console.log('Before handleSubmit:', {
          canSubmit: form.state.canSubmit,
          isValid: form.state.isValid,
          submissionAttempts: form.state.submissionAttempts,
        })
        form.handleSubmit()
        console.log('After handleSubmit:', {
          canSubmit: form.state.canSubmit,
          isValid: form.state.isValid,
          submissionAttempts: form.state.submissionAttempts,
        })
      }}
    >
      Submit
    </button>
  )
}
```

### Steps to reproduce

1. Create a form with `onChange` and `onSubmit` validators
2. Set default values that are invalid (e.g., `value: 0` when validator requires `min(1)`)
3. Define `onSubmitInvalid` callback
4. Click the submit button
5. Observe that `onSubmitInvalid` is never called

### Expected behavior

`onSubmitInvalid` should be called when:
- `canSubmit` is `false`
- `isValid` is `false`
- `handleSubmit()` is invoked

Every time

### Screenshots or Videos

Image

### Platform

- OS: macOS
- Browser: Chrome
- TanStack Form version: 1.27.7
- TypeScript version: 5.x

### TanStack Form adapter

None

### TanStack Form version

1.27.7

### TypeScript version

5.2.2

### Additional context

Looking at the source code in `FormApi.js`, the check at line ~462 should call `onSubmitInvalid`:

```javascript
if (!this.state.canSubmit && !this._devtoolsSubmissionOverride) {
  this.options.onSubmitInvalid?.({
    value: this.state.values,
    formApi: this,
    meta: submitMetaArg
  });
  return;
}
```

However, it seems like `this.state.canSubmit` might be reading a stale or different value than what's visible from `form.state.canSubmit` in React, possibly due to the derived store not being synchronized at the time of the check.

Dhavanesh24cs412:
- Remove stale canSubmit check that prevented onSubmitInvalid from being called
- On first render, canSubmit was true even with invalid default values
## Problem

When a form is initialized with invalid default values, the `onSubmitInvalid` callback was not being called on the first submission attempt. This occurred because an early `canSubmit` check in the `_handleSubmit` method would short-circuit the validation flow before reaching the proper validation logic.

## Root Cause

The issue was located in `packages/form-core/src/FormApi.ts` in the `_handleSubmit` method. An early check:

```typescript
if (!this.state.canSubmit && !this._devtoolsSubmissionOverride) {
  // Early return without proper validation
  return
}
```

This check was preventing the form from:
1. Calling `validateAllFields('submit')` to validate all fields
2. Executing the `onSubmitInvalid` callback when validation fails
3. Properly handling forms with invalid default values

## Solution

Removed the stale early `canSubmit` check from the `_handleSubmit` method. The form now:
- Proceeds directly to `validateAllFields('submit')`
- Properly calls `onSubmitInvalid` when validation fails
- Correctly handles submission attempts on forms with invalid default values
- Still respects validation logic through proper checks after validation occurs

## Changes Made

- ✅ Removed early `canSubmit` guard in `_handleSubmit` method
- ✅ Validation now properly executes on all submissions
- ✅ `onSubmitInvalid` callback is invoked when appropriate
- ✅ Fixes edge case with invalid default form values

## Testing

This fix addresses the issue described in #1990 where forms with invalid default values were not triggering the `onSubmitInvalid` callback.

Closes #1990- Fixes: onSubmitInvalid not called when form is invalid with default values

## ✅ Checklist

- [ ] I have followed the steps in the [Contributing guide](https://github.com/TanStack/form/blob/main/CONTRIBUTING.md).

## 🚀 Release Impact

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -2066,15 +2066,6 @@ export class FormApi<
     const submitMetaArg =
       submitMeta ?? (this.options.onSubmitMeta as TSubmitMeta)
 
-    if (!this.state.canSubmit && !this._devtoolsSubmissionOverride) {
-      this.options.onSubmitInvalid?.({
-        value: this.state.values,
-        formApi: this,
-        meta: submitMetaArg,
-      })
-      return
-    }
-
     this.baseStore.setState((d) => ({ ...d, isSubmitting: true }))
 
     const done = () => {


Dhavanesh24cs412:
This solution effectively addresses the issue where `onSubmitInvalid` wasn't being triggered for forms with invalid default values. By removing the early `canSubmit` check and allowing validation to execute properly, forms can now correctly invoke the callback even on initial submission. The approach maintains backward compatibility while fixing the edge case.

LeCarbonator:
In what way is `canSubmit` stale? If you have errors present that are not resolved, why would a submission change that?

Dhavanesh24cs412:
I would solve this issue by tonight! The 9 lines of code i deleted in my branch executed well once a run, i would double check and cinfigure the changes

crutchcorn (changes requested):
We need some kind of tests to validate that this fix works as-intended

PR #2006, opened 2026-01-25
fix(form-core): establish Field-over-Form prioritization for isDefaultValue and resets

### LINKED ISSUE #1973: `isDefaultValue` is `true` when field value is `undefined`, and also for two distinct values
### Describe the bug

Currently `isDefaultValue` reports true when the field value matches the form-level default, _or_ the field-level default, with the following consequences:

When the two defaults disagree, `isDefaultValue` reports `true` for both of them.

When only one default is specified, the other (missing) default is still tested against. As a result, `undefined` is always treated as a default value, which is unfortunate for forms that want to use `undefined` as one of the possible field values.

### Proposal

The goal is for `isDefaultValue` to be true if and only if `form.reset()` would make no changes. `form.reset()` picks a value to use. `isDefaultValue` could do the same: select a default first, then `evaluate(curFieldValue, selectedDefault)`.

In my testing, when both defaults are specified the initial render uses `field-default`, and `field-default` also "wins" in `form.reset()`. So I'm thinking to change the current implementation from this:

    isDefaultValue = evalute(curValue, formDefaultValue) || evaluate(curValue, fieldDefaultValue)
    
to this:

    isDefaultValue = evalute(curValue, fieldDefaultValue ?? formDefaultValue)

This addresses both scenarios: the (somewhat nonsensical) disagreeing defaults, and the unwanted `isDefaultValue=true` when the field value is `undefined`.

I could have a go at a PR if this change is acceptable.

https://codesandbox.io/p/sandbox/tanstack-form-isdefaultvalue-1-n9f6zn

### Steps to reproduce

In Example 1 on CodeSandbox, entering `<undefined>` sets the form value to `undefined`, and isDefaultValue becomes true. However clicking `form.reset()` changes it back to `foo`, which is the true default - showing that isDefaultValue=true was a lie. In my real code I ran into this while attempting to map an empty string to a luxon `DateTime | undefined` field.

In Example 2, entering both `foo-field` and `foo-form` results in isDefaultValue=true. But the initial value is foo-field, and that is also what we get if we form.reset(). So it should be the only value for which isDefaultValue returns true, even if this example is a bit of a misuse of default values.

### Expected behavior

`isDefaultValue` should be true if and only if `form.reset()` has no changes to make, because every value is already the default value.

Every time

### Screenshots or Videos

_No response_

### Platform

Windows, Firefox 146

### TanStack Form adapter

None

### TanStack Form version

1.27.7

### TypeScript version

5.9.3

### Additional context

_No response_

Kyujenius:
## 🎯 Changes

Fixes : #1973

This PR is built on the fundamental premise that **"The most specific scope (Field) must always take precedence over the generic scope (Form)."** Similar to variable shadowing in programming or specificity in CSS, an explicit declaration at the field level represents a stronger developer intent than a global default.

Currently, the lack of a clear hierarchy leads to architectural inconsistencies:
1. **The Paradox**: `isDefaultValue` might claim a value is "default," yet `reset()` restores a completely different value from the form-level config.
2. **The False Positive**: Clearing a field to `undefined` incorrectly triggers `isDefaultValue: true` even when explicit defaults are defined at the form level.

By unifying `isDefaultValue`, `form.reset()`, and `form.resetField()` under a single **Prioritization Strategy (Field over Form)**, this PR ensures that TanStack Form behaves as a predictable, high-integrity state machine.

### Introduction of Prioritized Default System
This PR introduces a unified prioritization strategy where **Field-level defaults always override Form-level defaults**.

#### Key Changes:
1. **`isDefaultValue` Determination**: Now uses a single prioritized default value (`Field ?? Form`) for comparison instead of a logical OR.
2. **`form.reset()` Consistency**: Now merges form-level defaults with currently mounted field-level defaults, preventing the "reset betrayal" where values would unexpectedly change to form-defaults.
3. **`form.resetField()` Consistency**: Updated to respect the same priority when resetting individual fields.

### Fixed Scenarios:
- **Scenario 1**: When a default value is specified, setting the field to `undefined` now correctly sets `isDefaultValue: false`.
- **Scenario 2**: When form-level and field-level defaults differ, both `isDefaultValue` and `reset()` now consistently follow the field-level default.

### ⚠️  I Updated Existing Tests 

I’d like to verify if this matches what the maintainers had in mind.

Modified current test cases in `FieldApi.spec.ts` that were previously based on the incorrect logical OR assumption. These tests were updated to reflect the new prioritized logic, ensuring that `isDefaultValue` only returns `true` for the actual value that `form.reset()` would restore.

### isDefaultValue Check Logic AS-IS / TO-BE
|AS-IS|TO-BE|
|---|---|
|image|image|

### resset() Logic AS-IS / TO-BE
|AS-IS|TO-BE|
|---|---|
|image|image|

###

## IMO
I realize this shifts the precedence closer to a 'Variable Shadowing' model. If this is too big of a breaking change, I'm happy to discuss putting this behind a flag or finding a middle ground. My main goal is to ensure reset() and isDefaultValue rely on the same logic. I believe this prioritization makes the library significantly more predictable. By aligning the internal logic with how developers naturally think about scope and specificity,

## ✅ Checklist

- [x] I have tested this code locally with `pnpm test:pr`. (Verified `form-core` tests and custom regression tests)

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Introduced Prioritized Default System that enforces consistent field-level default handling throughout form operations. Field-specific defaults now take precedence over form-wide defaults when resetting forms and deriving field state.

* **Bug Fixes**
  * Fixed improper default value detection when explicit values are specified, improving accuracy and reliability of default state tracking.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/form/pull/2006)


packages/form-core/src/FormApi.ts
```diff
@@ -2542,15 +2557,22 @@ export class FormApi<
    */
   resetField = <TField extends DeepKeys<TFormData>>(field: TField) => {
     this.baseStore.setState((prev) => {
+      const fieldDefault =
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        this.getFieldInfo(field)?.instance?.options.defaultValue
```
LeCarbonator: This lint error seems odd. Which part does it think does not need an optional chain?

From a surface level, this looks like a wrong type that should be adjusted to reflect (potential) runtime values.

packages/form-core/tests/FieldApi.spec.ts
```diff
@@ -120,7 +120,7 @@ describe('field api', () => {
     expect(field.getMeta().isDefaultValue).toBe(false)
 
     field.setValue('test')
-    expect(field.getMeta().isDefaultValue).toBe(true)
+    expect(field.getMeta().isDefaultValue).toBe(false)
```
LeCarbonator, line 123: Reminder to self: Check git blame.

I don't this change is good, but I want to know the context of why it was explicitly listed as unit test.

LeCarbonator (commented):
Looks clean! I'll need to do some additional checks though, since this is potentially a breaking change instead of a fix.

It may take a few days. Thanks for tackling the issue!

packages/form-core/src/FormApi.ts
kyu-rong: You're right! `getFieldInfo()` always returns a value due to `||=` initialization, so the first `?.` is indeed unnecessary.

I'll update it to: (without lint disable)
```typescript
this.getFieldInfo(field).instance?.options.defaultValue


```

packages/form-core/tests/FieldApi.spec.ts
kyu-rong: @LeCarbonator Thanks for taking the time to review this!

I checked the git blame - this test was intentionally written this way in PR #1456 It wasn't a mistake, so I want to be careful here.

But I think there's a philosophical question: **What should "default" mean when form-level and field-level disagree?**

### The Two Interpretations

**Original (OR logic)**: "Both are defaults"
```
isDefaultValue = matches(formDefault) || matches(fieldDefault)
```

**Proposed (?? logic)**: "The one `reset()` uses is THE default"
```
isDefaultValue = matches(fieldDefault ?? formDefault)
```

### Why I Lean Toward the Proposed Logic

In #1081, explained the purpose of `isDefaultValue`:

> "if you do something like `!meta.isDefaultValue` will give you the react ecosystem's isDirty"

RHF's `isDirty` is `true` when `currentValue !== defaultValue`. A form has exactly **one** clean state.

With the original OR logic:
- `isDefaultValue` is `true` for both `'test'` and `'another-test'`
- So `!isDefaultValue` is `false` for both values
- This means the form has **two** clean states

That feels inconsistent with RHF's model. A form should have one default state, not two.

### The Practical Problem

```typescript
// User checks before reset
if (!field.getMeta().isDefaultValue) {
  form.reset()  // "I'm not at default, let me reset"
}
```

With OR logic, if value is `'test'` (form-level default):
- `isDefaultValue` → `true` → user skips reset
- But `reset()` would actually change the value to `'another-test'`

The user gets misleading information.

### I Could Be Wrong

I understand the original design might have had reasons I'm not aware of. Maybe there are use cases where treating both as "default" makes sense. What's your take on this?

crutchcorn (approved):
This looks really good. Thank you for the thoughtful PR and description; sorry that it took me so long to review + merge.

Just as a heads up, @LeCarbonator is considering some changes to `defaultValue` in a field for v2. Namely removing it outright. Would love to hear your feedback on that idea as someone who's clearly thought about field-first default values.

Kyujenius:
@crutchcorn Thanks so much, and no worries on the timing!

I’ve picked up on bits of the related conversations around field-level vs form-level defaults, so dropping field-level defaultValue in v2 sounds like a reasonable direction from where I’m sitting. Would love to contribute to that effort in whatever capacity is useful — just let me know where the discussion lives and I’ll join in.

PR #2012, opened 2026-02-02
test(form-core): improve mergeForm coverage to 100%

Kyujenius:
## 🎯 Changes

This PR improves test coverage for `mergeForm.ts` from 87.09% to 100% by adding comprehensive tests for previously uncovered edge cases.

### Coverage Improvement

**Before:**
- Uncovered lines: 17, 19, 105-106

**After:**
- All lines covered

image

### Type System Note

The `mergeForm` tests use `as any` type assertion due to FormApi's 12 generic parameters where `TSubmitMeta` defaults to `never`, creating incompatibility with mergeForm's `any` constraint. This is a type system limitation only - runtime behavior is correct. In production usage via `useTransform`, this receives `AnyFormApi` and avoids the issue.

## ✅ Checklist

## 🚀 Release Impact

- [x] This change is docs/CI/dev-only (no release).

This PR only adds tests without modifying any implementation code or public APIs. No changeset is required.


crutchcorn:
There's some behavior here that I'm not confident we want to codify with a test. Like the `return an object when merging null`. I'm going to close this as 100% test coverage is explicitly not our goal at all times.

PR #2035, opened 2026-02-13
Move to Alien-Signals Based Store

crutchcorn:
This PR introduces the new changes made by TanStack Store to rebase our internals on top of a faster more stable store package. It should bring massive performance improvements and enable further improvements

### DIFF
--- a/.changeset/silly-swans-take.md
+++ b/.changeset/silly-swans-take.md
@@ -1,0 +1,11 @@
+---
+'@tanstack/angular-form': patch
+'@tanstack/svelte-form': patch
+'@tanstack/react-form': patch
+'@tanstack/solid-form': patch
+'@tanstack/form-core': patch
+'@tanstack/lit-form': patch
+'@tanstack/vue-form': patch
+---
+
+Refactor internals for substancially faster performance


packages/form-core/src/FieldApi.ts
```diff
@@ -1194,7 +1193,10 @@ export class FieldApi<
               TFormOnDynamic,
               TFormOnDynamicAsync
             >
-          | undefined
+          | undefined,
+      ) => {
+        // Temp hack to subscribe to form.store
+        this.form.store.get()
```
LeCarbonator, line 1199: This temp hack will stay for a while, right? Or should this be removed for the PR?

crutchcorn: This will need to stay for a while

Lukas-Kullmann:
For what it's worth, this was a breaking change since it changed the API for `form.store.subscribe`. 😕

Previously:

```ts
const unsubscribe = form.store.subscribe(({ currentVal, prevVal }) => { ... });
```

Now:

```ts
// unsubscribe has to be destructured, `prevVal` is missing entirely and `currentVal` does not need to be destructured any more
const { unsubscribe } = form.store.subscribe(currentVal => { ... });
```

PR #2043, opened 2026-02-15
Add preact integration for tanstack form

JoviDeCroock:
## 🎯 Changes

- Add new `@tanstack/preact-form` package to the monorepo.
- Use `@tanstack/preact-store` instead of `@tanstack/react-store`.
- Use `useId` from `preact/hooks` for form ID generation.
- Set `preact` peer dependency to `>10.11.0` for useId support.

## Summary by CodeRabbit

* **New Features**
  * Official Preact support: new Preact form package with hooks, field components, form composition helpers, and reactive subscriptions.

* **Documentation**
  * Extensive Preact docs: Quick Start, Guides (validation, arrays, async initial values, listeners, reactivity, debugging, focus management, composition, UI libs, submission, etc.), and full API reference.

* **Examples**
  * Added a runnable Preact example demonstrating common form patterns.

* **Tests**
  * Added runtime and type-level tests exercising Preact form APIs.

* **Chores**
  * Workspace, lint, build and test config updates plus package metadata and docs generation.


packages/preact-form/src/useFieldGroup.tsx
```diff
@@ -0,0 +1,265 @@
+'use client'
```
theVedanta: can we remove these please

packages/preact-form/src/useFieldGroup.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+import { useState } from 'preact/compat'
```
theVedanta: Do we wanna bring in compat? I feel like it defeats the purpose

packages/preact-form/src/useForm.tsx
```diff
@@ -0,0 +1,282 @@
+'use client'
```
theVedanta: Also here

packages/preact-form/src/useForm.tsx
```diff
@@ -0,0 +1,282 @@
...
+  in out TOnServer extends undefined | FormAsyncValidateOrFn<TFormData>,
+  in out TSubmitMeta,
+> {
+  /**
+   * A React component to render form fields. With this, you can render and manage individual form fields.
+   */
+  Field: FieldComponent<
+    TFormData,
+    TOnMount,
+    TOnChange,
```
theVedanta, line 43: I feel like fields should have an onInput prop instead of onChange for it to be consistent with Preact. If we can do that, it'd be awesome

theVedanta (changes requested):
compat and use-client are the only things I wanna flag, since they're in multiple files. Apart from that, looks okay. The onInput prop isn't important as such, but would be a nice-to-have.

packages/preact-form/src/useForm.tsx
JoviDeCroock: I don't think anyone would benefit from that, the end-user is still free to use onInput in exchange for onChange at their disposal

theVedanta: Oh, gotcha. But how would users use onInput in parts of the API that are not the children?

```tsx
          <form.Field
            name="firstName"
            validators={{
              onInput: ({ value }) =>
                !value
                  ? 'A first name is required'
                  : value.length < 3
                    ? 'First name must be at least 3 characters'
                    : undefined,
            }}
            ...
```


I can understand if you feel like they don't, I was only thinking about it because of consistency with Preact's naming.

theVedanta (commented):
Once we can finalize this, let's get the docs in place and merge! Super excited to use preact-form 🔥

packages/preact-form/src/useForm.tsx
JoviDeCroock: Ah I see what you mean now, sorry I didn't derive the validator enforcement from the comment

crutchcorn: Validators should use onChange even on adapters that use onInput. That comes from `core`. No need to change :)

JoviDeCroock: I'm a bit torn on this, it feels a bit too distinct from how tanstack-form works to change this. Might be better to leave this. I am not sure

theVedanta:
LGTM! @crutchcorn just needs to approve the worflows perhaps?

crutchcorn (approved):
Pushed some minor changes to sync with `main`. Otherwise LGTM. Waiting on @KevinVandy to help us publish before merging (he'll merge for us)

theVedanta (approved):

LeCarbonator:
Looks like the lock has some conflicts now

packages/preact-form/src/useStore.ts
```diff
@@ -0,0 +1,44 @@
+import { useStore as usePreactStore } from '@tanstack/preact-store'
+
+export type NoInfer<T> = [T][T extends any ? 0 : never]
+
+type EqualityFn<T> = (objA: T, objB: T) => boolean
+
+export interface UseStoreOptions<T> {
```
JoviDeCroock:
```suggestion
interface UseStoreOptions<T> {
```

JoviDeCroock:
I'm sorry but I was asked to rebase this multiple times now, just for this to go quiet again 😅 I am going to close this out as it looks like there is no interest in landing this

crutchcorn:
There is absolutely interest in landing this. As we mentioned privately, we were having difficulties with merging any new packages. I don't have permissions to with our new release setup.

I'll mention internally again.

PR #2044, opened 2026-02-17
fix(react-form): Make form.options reactive by adding getter

swushi:
## Summary

Fixes an issue where `form.options.defaultValues` was not reactive when initializing a form with async data.

## Problem

When using `useForm()` with async data:

```tsx
const { data } = useAsyncApiData();

const form = useForm({
  defaultValues: data ?? { foo: '' },
})

// When data loads, `form.state.values` updates correctly,
// but form.options.defaultValues is stale (i.e. `{ foo: '' }`)
const hasChanges = form.state.values.foo !== form.options.defaultValues.foo; // incorrectly true
```

## Root Cause

The `options` object was captured in the `useMemo` closure and never updated when `formApi.options` changed, even though `formApi.update()` was called in the effect lifecycle.

## Solution

Added a getter for `options` (similar to the existing `state` getter) to ensure the latest options are always returned:

```tsx
get options() {
  return formApi.options
},
```

This follows the same pattern as the `state` getter, as mentioned in the existing comment: _"We must add all `get`ters from `core`'s `FormApi` here, as otherwise the spread operator won't catch those"_

## Related Discussion

https://canary.discord.com/channels/719702312431386674/1471891391348215972


crutchcorn:
@swushi this look good! Can you add a test in either `form-core` or `react-form` that validates this change? Would help us not regress in the future.

PR #2057, opened 2026-03-03
fix: deepkeyandvalue behavior in union types

### LINKED ISSUE #1813: Union types are not handled correctly
### Describe the bug

The internal types of Tanstack Form, `DeepValue` and `DeepKeysAndValues` don't handle union types correctly and seem to collapse them.

```ts
type Example = { id: string } | { id: undefined };

type Test = DeepKeysAndValues<Example>
//   ^? { key: "id"; value: string; }
```

```ts
const testObj: { id: string } | { id: undefined } = Math.random()
    ? { id: undefined }
    : { id: '' }

type idField = DeepValue<typeof testObj, 'id'>
// Type error - TS2322: Type undefined is not assignable to type string
const a: idField = undefined
const b: idField = 'zzz'
```

My use case was using a union type with `withForm`,  this also happens with `withFieldGroup` 

```ts
import type { MergeExclusive, RequiredDeep } from 'type-fest'

export type EditType = ContactType & ContactPatchType // ContactType has readonly properties like `id`, ContactPatchType has the properties that are editable
export type CreateType = RequiredDeep<ContactDataType> // ContactDataType contains properties that are used for initial create, so no `id` for example
type EitherType = MergeExclusive<CreateType, EditType>

export const ContactFormFields = withForm({
    defaultValues: {} as EitherType,
...
```

`EditType` contains `{id: string}` while `CreateType` does not. You would expect a form to understand that `id` may be `string | undefined`

Instead it becomes just `{id: string}` 

Due to the incorrect type, this does not produce a type error but can fail in runtime since `field.state.value` may be `undefined`
```ts
<group.AppField name="id">
  {(field) => <div>ID: {field.state.value.length}</div>}
</group.AppField>
```

N/A

### Steps to reproduce

^ See Above

### Expected behavior

Union types to be handled properly

Every time

### Screenshots or Videos

_No response_

### Platform

`"@tanstack/react-form": "1.23.7",`

### TanStack Form adapter

react-form

### TanStack Form version

1.23.7

### TypeScript version

5.9.3

### Additional context

_No response_

mdm317:
## 🎯 Changes
fixes #1813

**Fix DeepKeysAndValues to return nullish values instead of never when the value type is nullish**

- main
```
type Test = DeepKeysAndValues<{ id : undefined }>
//        ^? never
```
[ts playground](https://www.typescriptlang.org/play/?ssl=3&ssc=8&pln=1&pc=1#code/JYWwDg9gTgLgBAbzgEQKarAaVQTwM4CCAdgCYBqAhgDYCuqecAvnAGZQQhwDkAAjBUTz8AxgGsA9FFQVhMALQtoILgCgYOMKjgAVevAC8KdFlyFSlWvQA8SYCTgAuODVKoWwIqnuMAfCvHicEEAegD8QA)

 - this branch
```
type Test = DeepKeysAndValues<{ id : undefined }>
//        ^? { key : 'id', value: undefined}
```
---

**Due to the above change, I replaced DeepKeysOfType with DeepKeysOfNonNullableType in the group API.**

> https://github.com/TanStack/form/pull/2057#discussion_r2876652617

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed type behavior so unions keep nullish members where appropriate and field-group selections reject paths that are nullish-only.

* **Tests**
  * Added type-level tests covering nullish preservation, tuple/array index cases, discriminated-union nulls, and rejection of nullish-only field-group paths.

* **Chores**
  * Bumped form-related packages with a patch changeset.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/form/pull/2057?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/form-core/src/util-types.ts
```diff
@@ -194,3 +196,14 @@ export type FieldsMap<TFormData, TFieldGroupData> =
             TFieldGroupData[K]
           >
         }
+
+type ExtractByNonNullableValue<T, TValue> = T extends { value: infer V }
+  ? [NonNullable<V>] extends [never]
+    ? never
```
mdm317, line 219: Since the previous behavior inferred never when the value was `nullish`, 
the changes were updated to continue inferring never for `nullish` value types.

packages/form-core/src/util-types.ts
```diff
@@ -194,3 +196,14 @@ export type FieldsMap<TFormData, TFieldGroupData> =
...
+    : NonNullable<V> extends TValue
+      ? T
+      : never
+  : never
+
+export type DeepKeysOfNonNullableType<TData, TValue> =
+  ExtractByNonNullableValue<DeepKeysAndValues<TData>, TValue>['key']
```
mdm317, line 226: After changing `DeepKeysAndValues` to include `nullish` values, a bug was introduced in the group API when using `DeepKeysOfType`.

When the value type is nullish, it previously evaluated as:
```
Extract<never, nullish | TValue>
```
However, after the change, it became:
```
Extract<nullish, nullish | TValue>
```

which now returns nullish.
To resolve this issue, I implemented a new utility type to correctly infer `TFields` in the `group API`.

crutchcorn:
The code looks good to me and makes sense, but given that @LeCarbonator has to remind me every week how FieldGroup actually works, I'm going to pass it to him 😅

Pinged internally so response time is usually quicker that way.

packages/form-core/src/util-types.ts
```diff
@@ -135,21 +135,23 @@ export type DeepKeysAndValuesImpl<
...
+        ? TAcc
+        : T extends ReadonlyArray<any>
+          ? number extends T['length']
+            ? DeepKeyAndValueArray<TParent, T, TAcc>
+            : DeepKeyAndValueTuple<TParent, T, TAcc>
+          : keyof T extends never
+            ? TAcc | UnknownDeepKeyAndValue<TParent>
+            : T extends object
+              ? DeepKeyAndValueObject<TParent, T, TAcc>
+              : TAcc
```
LeCarbonator, line 171: This particular change addresses not just `DeepKeysOfType`, but also `DeepKeys` and `DeepValue`. 

I see one test was adjusted to acknowledge this, but some additional LOC for `DeepKeys` and `DeepValue` would be fantastic. Not merely for this PR, but also for future typings to follow the spec.

packages/form-core/src/util-types.ts
LeCarbonator, line 1: How's type instantiation treating this type change? Have you tested them before / after? Did they explode or is it only a bit more?

I would be surprised if it did, but you never know with some of those seemingly simple changes.

If you haven't already, I can help out with testing it. Just let me know.

LeCarbonator (changes requested):
Looks interesting! Reported cause and the reasoning for the change make sense to me! As far as the changes go, there's only some regression tests missing as well as an instantiation check. Refer to the two open threads for more details.

Thanks for tackling this!

packages/form-core/src/util-types.ts
mdm317: Measured with the following command.
  `Instantiations` increased slightly on this branch's HEAD `c0d762dd` compared to base `3886dcc5`.

  ```bash
npx tsc --extendedDiagnostics
  ```

  ### `packages/form-core`
  | metric | base | HEAD |
  |---|---:|---:|
  | Types | 84,335 | 86,803 |
  | Instantiations | 436,577 | 449,014 |

  ### `packages/react-form`
  | metric | base | HEAD |
  |---|---:|---:|
  | Types | 58,924 | 60,824 |
  | Instantiations | 435,282 | 447,536 |

  ### `packages/preact-form`
  | metric | base | HEAD |
  |---|---:|---:|
  | Types | 51,525 | 52,567 |
  | Instantiations | 322,435 | 326,877 |

  ### `packages/vue-form`
  | metric | base | HEAD |
  |---|---:|---:|
  | Types | 25,531 | 26,099 |
  | Instantiations | 113,263 | 114,659 |

  ### `packages/solid-form`
  | metric | base | HEAD |
  |---|---:|---:|
  | Types | 37,855 | 39,301 |
  | Instantiations | 203,009 | 210,621 |

packages/form-core/src/util-types.ts
mdm317: Thanks for review!
I added case in this commit [2ee891c](https://github.com/TanStack/form/pull/2057/commits/2ee891ceb0575a97043224133c3444349f311bc6)

packages/preact-form/src/useFieldGroup.tsx
```diff
@@ -91,7 +91,7 @@ export function useFieldGroup<
   TFormData,
   TFieldGroupData,
   TFields extends
-    | DeepKeysOfType<TFormData, TFieldGroupData | null | undefined>
+    | DeepKeysOfNonNullableType<TFormData, TFieldGroupData>
```
mdm317, line 94: Apply the same non-nullable field group fix to preact-form.

packages/form-core/src/util-types.ts
LeCarbonator: That's only a minor amount because this would only account for a few unit tests.

I took the liberty to give it a try in my work space. We use TanStack Form a lot there and we have way more types, so the impact should be a bit clearer.

  ### `packages/react-form`
  | metric | v1.28 | HEAD | Diff (~) |
  |---|---:|---:|---:|
  | Types | 439'839 | 439'089 | - 0.17% |
  | Instantiations | 2'983'698 | 2'978'088 | - 0.18% |
  | Memory | 1'406'720K | 1'412'563K | + 0.41% |
  
  
  Looks clean!

mdm317: Thanks for checking — appreciate it.
Let me know if there’s anything else I should work on.

PR #2065, opened 2026-03-06
fix(react-form): default selector to identity in Subscribe component

### LINKED ISSUE #2063: <form.Subscribe> allows missing selector in types, but crashes at runtime (selector is not a function)
### Describe the bug

`form.Subscribe` currently allows omitting `selector` at type level, but in runtime it passes `undefined` into `useStore`, which throws:

`TypeError: selector is not a function`

So there is no TypeScript warning, but the app crashes when rendering `<form.Subscribe>` without `selector`.

https://tanstack.com/form/latest/docs/examples/react/simple

### Steps to reproduce

```tsx
const form = useForm({
  defaultValues: { name: "" },
  onSubmit: async () => {},
})

return (
  <form.Subscribe>
    {(s) => <button disabled={s.isSubmitting}>Save</button>}
  </form.Subscribe>
)
```

### Expected behavior

Either:
- selector should be optional and default internally to identity ((state) => state), or
- selector should be required in types so TS catches this at compile time.

None

### Screenshots or Videos

_No response_

### Platform

macOS

### TanStack Form adapter

None

### TanStack Form version

1.28.4

### TypeScript version

5.9.3

### Additional context

_No response_

isaackaara:
When `<form.Subscribe>` is rendered without a `selector` prop, it crashes at runtime:

```
TypeError: selector is not a function
```

The types already mark `selector` as optional (`selector?`), so the runtime should handle the missing case. This adds a default identity function `(state) => state` for the `selector` parameter in both `LocalSubscribe` implementations (`useForm` and `useFieldGroup`).

**Before (crashes):**
```tsx
<form.Subscribe>
  {(state) => <button disabled={state.isSubmitting}>Save</button>}
</form.Subscribe>
```

**After (works, returns full form state):**
```tsx
<form.Subscribe>
  {(state) => <button disabled={state.isSubmitting}>Save</button>}
</form.Subscribe>
```

Fixes #2063

### DIFF
--- a/packages/react-form/tests/createFormHook.test.tsx
+++ b/packages/react-form/tests/createFormHook.test.tsx
@@ -731,7 +731,9 @@ describe('createFormHook', () => {
             />
             <group.Subscribe
               children={(state) => (
-                <span data-testid="state-lastName">{state.values.lastName}</span>
+                <span data-testid="state-lastName">
+                  {state.values.lastName}
+                </span>
               )}
             />
           </div>


LeCarbonator:
dw about the coverage failing, as long as the lines you changed are covered.

I'll review in a bit. Thanks!

packages/react-form/src/useFieldGroup.tsx
```diff
@@ -23,11 +23,11 @@ import type { LensFieldComponent } from './useField'
 
 function LocalSubscribe({
   lens,
-  selector,
+  selector = (state) => state,
   children,
 }: PropsWithChildren<{
   lens: AnyFieldGroupApi
-  selector: (state: FieldGroupState<any>) => FieldGroupState<any>
```
LeCarbonator, line 30: You mentioned that "the types already mark the selector as optional", but it wasn't marked here.

A subscription without a selector (or an identity selector) makes no sense, so we shouldn't encourage it. For the sake of semver, the identity function fallback should stay, but the types should enforce that a selector is required.

packages/react-form/tests/createFormHook.test.tsx
```diff
@@ -699,4 +699,63 @@ describe('createFormHook', () => {
     const button = getByText('Testing')
     expect(button).toBeInTheDocument()
   })
+
+  it('should render FieldGroup Subscribe without selector (default identity)', async () => {
```
LeCarbonator, line 703: Of course, if the selector will be marked as required, these tests would type error.

Adjust the title to mention the related GH issue or that this is for semver.

LeCarbonator:
I checked through some source code again, and it looks like it is explicitly marked as optional in multiple places.

This should still be changed at some point, but since it looks intentional and has runtime implications, we'll keep it as optional.

Thanks for the adjustment!

packages/react-form/src/useFieldGroup.tsx
ShrJamal: Could you explain why using a subscription without a selector makes no sense? In some cases, I only want the content inside a `Subscribe` component (usually a submit button) to react to multiple form states like `isValid`, `isSubmitting`, `isTouched`.....
Yes, I could return all of these from a selector, but then I’d have to keep updating the selector whenever I want to listen to additional states. Personally, I don’t mind if that button re-renders more often than strictly necessary.

isaackaara:
Good question. The reasoning is about preventing unintentional performance issues by default.

Without a selector, `Subscribe` re-renders on *every* state change across the entire form/field group. For small forms that's fine, but as forms grow (nested fields, arrays, complex validation), this can create cascading re-renders that are hard to trace.

Your use case is valid: a submit button watching `isValid`, `isSubmitting`, `isTouched`, etc. But you're consciously choosing to accept those re-renders. The default selector approach forces that decision to be explicit:

```tsx
<Subscribe selector={(state) => state}>
  {/* I know this re-renders often */}
</Subscribe>
```

vs. accidentally forgetting a selector and wondering why performance degrades.

That said, if the maintainers prefer allowing `undefined` and trusting users to know what they're doing, I'm happy to revert this change. Just seemed safer to make performance opt-out rather than opt-in.

LeCarbonator:
I think the decision is clear that it should always be granular and specified.

However, given the semantic versioning, that's not going to be possible in v1. It will be noted for v2.

PR #2068, opened 2026-03-06
fix(core): field unmount

TkDodo:
Add unmounting to fields. `field.mount` now returns a cleanup function for unmounting.

Without unmounting, conditionally rendered fields will keep their field-level validations running even though the field isn’t rendered anymore.

## Summary by CodeRabbit

* **New Features**
  * Added form option cleanupFieldsOnUnmount (default: false) to control clearing field meta/state on unmount.
  * Field mount now returns an explicit cleanup/unmount function.

* **Bug Fixes**
  * Unmount now cancels in-flight async validations and debounced listeners to avoid stale callbacks.
  * Hidden/conditionally rendered fields no longer retain submit errors after unmount when cleanup is enabled.

* **Tests**
  * Expanded coverage for mount/unmount, value/meta preservation, async cancellation, and remount behavior.

### DIFF
--- a/.changeset/red-hats-jam.md
+++ b/.changeset/red-hats-jam.md
@@ -1,6 +1,6 @@
 ---
-"@tanstack/form-core": patch
-"@tanstack/react-form": patch
+'@tanstack/form-core': patch
+'@tanstack/react-form': patch
 ---
 
 fix(core): field unmount
--- a/.changeset/red-hats-jam.md
+++ b/.changeset/red-hats-jam.md
@@ -1,0 +1,6 @@
+---
+"@tanstack/form-core": patch
+"@tanstack/react-form": patch
+---
+
+fix(core): field unmount
--- a/docs/framework/angular/guides/listeners.md
+++ b/docs/framework/angular/guides/listeners.md
@@ -15,10 +15,11 @@ In this example, when the user changes the country, the selected province needs 
 
 Events that can be "listened" to are:
 
-- onChange
-- onBlur
-- onMount
-- onSubmit
+- `onChange`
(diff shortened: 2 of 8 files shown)


harry-whorlow:
This is a really solid PR! 🚀

harry-whorlow (approved):

Dakkers:
@harry-whorlow would syncing this branch with `main` fix the CI failures?

examples/react/simple/src/index.tsx
```diff
@@ -30,6 +30,7 @@ export default function App() {
     },
   })
 
+  const [hide, setHIde] = React.useState(true)
```
LeCarbonator: typo
```suggestion
  const [hide, setHide] = React.useState(true)
```

examples/react/simple/src/index.tsx
```diff
@@ -40,42 +41,50 @@ export default function App() {
           form.handleSubmit()
         }}
       >
+        <button onClick={() => setHIde((prev) => !prev)}>hide</button>
```
LeCarbonator:
```suggestion
        <button onClick={() => setHide((prev) => !prev)}>hide</button>
```

LeCarbonator (approved):
Looks good! Did a quick sanity check, only noticed a small typo.

LeCarbonator:
@Dakkers No it would not. But it'll be fixed soon.

examples/react/simple/src/index.tsx
harry-whorlow: yeah, its in a test commit... its been cut now

PR #2073, opened 2026-03-10
fix(react-form): re-render mode="array" fields after swapValues/moveValue

### LINKED ISSUE #2018: Field with `mode="array"` does not re-render after `swapValues` / `moveValue` since version 1.27.4
### Describe the bug

When using Field with `mode="array"`, calling `swapValues` or `moveValue` updates the internal form state, but the component does not re-render.

In the example below, the fruits array is updated internally, but `field.state.value` does not reflect the change in the render output until another event triggers a re-render.

```tsx
const defaultValues = {
  fruits: ["apple", "banana", "cherry"],
};

const form = useForm({
  defaultValues,
});

return (
  <form.Field name="fruits" mode="array">
    {(field) => (
      <div>
        <pre>{JSON.stringify(field.state.value, null, 2)}</pre>
        <button onClick={() => field.swapValues(0, 2)}>
          array move
        </button >
      </div>
    )}
  </form.Field>
);
```

https://codesandbox.io/p/sandbox/black-bush-7dnvmp

### Steps to reproduce

1. Render an `Field` with `mode="array"`.

2. Call` swapValues` or `moveValue` on the field.

3. The array value updates internally but the UI does not re-render.

### Expected behavior

Calling swapValues or moveValue should trigger a re-render and update `field.state.value` accordingly.

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/c0ee0613-312a-40ec-8847-f2371baf0b07

### Platform

WSL2 Ubuntu 24.04
Browser: Chrome 144.0.7559.110

### TanStack Form adapter

react-form

### TanStack Form version

1.28.0

### TypeScript version

_No response_

### Additional context

_No response_

isaackaara:
Closes #2018

## What broke

When using `<Field mode="array">`, calling `swapValues` or `moveValue` updated the internal form state correctly but the component did not re-render. The UI kept showing the old order until something else triggered a re-render.

Root cause: PR #1930 (fixing #1925) changed the `useStore` selector for `mode="array"` to only track array length, not full value. That stopped the unwanted re-renders from child property changes — but as a side effect, reordering operations that keep the array length the same (swap, move) also stopped triggering re-renders.

## The fix

Added a `useReducer` version counter (`arrayStructuralVersion`) that increments every time `swapValues` or `moveValue` is called. The counter is included in the `useMemo` dependency array, so any reorder forces a re-render.

The `mode="array"` length-only optimization from #1930 is kept intact — child property changes still do not cause unnecessary array field re-renders.

## Testing

- Added a test that calls `swapValues` and `moveValue` on a `mode="array"` field and verifies the component re-renders and displays the updated order.
- Existing test for #1925 (`should not rerender array field when child field value changes`) still passes.

### DIFF
--- a/.changeset/fix-array-swap-rerender.md
+++ b/.changeset/fix-array-swap-rerender.md
@@ -1,0 +1,16 @@
+---
+'@tanstack/react-form': patch
+---
+
+fix(react-form): re-render mode="array" fields after swapValues/moveValue
+
+When `swapValues` or `moveValue` was called on a `mode="array"` field, the
+array field did not re-render because the array length didn't change and the
+selector only tracked length (to avoid re-renders on child property changes,
+see #1925).
+
+The fix introduces a small version counter via `useReducer` that bumps
+whenever `swapValues` or `moveValue` is called. This forces a re-render so
+the displayed item order stays in sync with the form state.
+
+Fixes #2018
--- a/packages/react-form/src/useField.tsx
+++ b/packages/react-form/src/useField.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { useMemo, useRef, useState } from 'react'
+import { useMemo, useReducer, useRef, useState } from 'react'
 import { useStore } from '@tanstack/react-store'
 import { FieldApi, functionalUpdate } from '@tanstack/form-core'
 import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect'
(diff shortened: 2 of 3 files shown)


crutchcorn:
We should not overwrite the methods in the specific adapters. Instead, we should either expose some kind of state from `core` and consume that or find another method to reflect the changes. Closing for now

PR #2077, opened 2026-03-10
docs(react): add Getting and Setting Form Values guide

### LINKED ISSUE #478: [Docs] Guide page about getting and setting field values

mixelburg:
## Summary

Adds a new guide page (React) covering the frequently-asked question of how to read and write form values in TanStack Form.

Closes #478

## What's included

The guide walks through all the scenarios raised in the issue:

| Topic | What the guide covers |
|---|---|
| Initializing the form | `defaultValues` (static & shared via `formOptions`) |
| Async / API values | Plain `fetch` + state pattern; TanStack Query (links to existing async-initial-values guide) |
| Reading values reactively | `useStore`, `form.Subscribe` (links to Reactivity guide) |
| Reading values imperatively | `form.getFieldValue()`, `form.state.values` |
| Setting a single field | `form.setFieldValue()`, `field.setValue()` / `field.handleChange()` |
| Setting all fields at once | `form.reset(newValues)` |
| Validation when setting programmatically | Runs by default; can suppress with `{ dontValidate: true }`; manual trigger via `form.validateField()` |
| Marking a field as touched | `field.setMeta()` alongside the value update |
| Listening for changes | Snippet + link to the Listeners guide |

The guide is also added to the navigation in `docs/config.json` right after "Basic Concepts".

## Checklist

- [x] Docs only change (no source code touched)
- [x] All code snippets are consistent with existing guides (same import style, same naming conventions)
- [x] Added entry in `docs/config.json` under the React Guides section

### DIFF
--- a/docs/config.json
+++ b/docs/config.json
@@ -104,6 +104,10 @@
               "to": "framework/react/guides/basic-concepts"
             },
             {
+              "label": "Getting and Setting Values",
+              "to": "framework/react/guides/getting-and-setting-values"
+            },
+            {
               "label": "Form Validation",
               "to": "framework/react/guides/validation"
             },
--- a/docs/framework/react/guides/getting-and-setting-values.md
+++ b/docs/framework/react/guides/getting-and-setting-values.md
@@ -1,0 +1,264 @@
+---
+id: getting-and-setting-values
+title: Getting and Setting Form Values
+---
+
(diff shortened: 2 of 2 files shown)


crutchcorn:
This was a fairly outdated issue that we closed before this PR. I think this guide introduces too much duplication from other guides IMO

PR #2083, opened 2026-03-15
fix(react-form): restore optional selector in re-exported useStore

mixelburg:
## Summary

Fixes TanStack/store#323

## Problem

Prior to `@tanstack/react-store@0.9.1`, `useStore` could be called with just a store reference:

```ts
const formState = useStore(form.store) // ✅ worked before
```

The `0.9.1` update to `react-store` made the `selector` argument required in its TypeScript types. Because `@tanstack/react-form` re-exported `useStore` directly:

```ts
export { useStore } from '@tanstack/react-store'
```

…any project that had previously used the single-argument form now sees a TypeScript build error:

```
error TS2554: Expected 2-3 arguments, but got 1.
```

## Fix

Instead of re-exporting `useStore` verbatim, this PR adds a thin overload wrapper in `packages/react-form/src/useStore.ts` that:

1. **No selector** – returns the full atom snapshot (restores pre-1.28.4 behaviour)
2. **With selector** – passes through to the underlying `@tanstack/react-store` implementation

At runtime the identity function `(s) => s` is used as the default selector, which matches what the underlying implementation would have done before.

## Summary by CodeRabbit

* **Refactor**
  * Improved the useStore hook with stronger typing, overloads for returning full snapshots or selector-derived slices, and preserved comparison behavior for selectors.
* **Chore**
  * Switched the public export source of useStore to the package's local implementation (no public API name changes).

### DIFF
--- a/packages/react-form/src/index.ts
+++ b/packages/react-form/src/index.ts
@@ -1,6 +1,6 @@
 export * from '@tanstack/form-core'
 
-export { useStore } from '@tanstack/react-store'
+export { useStore } from './useStore'
 
 export * from './createFormHook'
 export * from './types'
--- a/packages/react-form/src/useStore.ts
+++ b/packages/react-form/src/useStore.ts
@@ -1,0 +1,44 @@
+import { useStore as _useStore } from '@tanstack/react-store'
+import type { AnyAtom } from '@tanstack/store'
+
+type AtomSnapshot<TAtom> = TAtom extends { get: () => infer TSnapshot }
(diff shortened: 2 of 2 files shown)


LeCarbonator:
Apart from the TypeScript error, does it cause errors at runtime when omitted?

crutchcorn:
I agree this should be fixed, but it should be done upstream in TanStack Store. Closing here, but would review a PR to re-introduce it to `useStore` in that library

PR #2103, opened 2026-03-31
fix(listeners): rename onFieldUnmount to match that of the other listeners

harry-whorlow:
## Summary by CodeRabbit

* **Breaking Changes**
  * Renamed form-level listener key from onFieldUnmount to onUnmount for consistency with other listeners.

* **New Features**
  * Built-in debouncing for listeners via onChangeDebounceMs and onBlurDebounceMs.

* **Documentation**
  * Added or updated comprehensive listener guides across frameworks covering events (onChange/onBlur/onMount/onSubmit/onUnmount), debouncing, field- and form-level listener examples, and propagated listener behaviors.

### DIFF
--- a/.changeset/rename-onFieldUnmount-to-onUnmount.md
+++ b/.changeset/rename-onFieldUnmount-to-onUnmount.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/form-core": minor
+---
+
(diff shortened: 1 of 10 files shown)


crutchcorn (changes requested):
We can't rename this now that it's in v1 - that's a breaking change. Let's keep the docs but revert the rename

PR #2104, opened 2026-03-31
docs: add guide about getting and setting field values

### LINKED ISSUE #478: [Docs] Guide page about getting and setting field values

Partharsid:
Closes #478. Added the missing documentation guide for programmatically getting and setting field values, and linked it in the React guides configuration.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive guide on reading and setting field values, covering synchronous access methods, state access patterns, and component render prop usage
  * Included guidance for programmatically updating field values with optional configuration options

### DIFF
--- a/docs/config.json
+++ b/docs/config.json
@@ -132,6 +132,10 @@
               "to": "framework/react/guides/listeners"
             },
             {
+              "label": "Getting and Setting Values",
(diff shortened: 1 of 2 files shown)


crutchcorn:
These docs feel duplicative from other docs we have now. Closing

PR #2116, opened 2026-04-10
docs: add guide for getting and setting field values

### LINKED ISSUE #478: [Docs] Guide page about getting and setting field values

Vitalini:
Fixes #478

Adds a new React guide page at `docs/framework/react/guides/field-values.md` covering the topics requested in the issue:

- How to initialize a form with `defaultValues`
- How to initialize with values from an API (with and without TanStack Query)
- How to get a single field value or the entire form state (`form.getFieldValue`, `form.state.values`, `field.state.value`)
- How to reactively subscribe to values (`form.useStore`, `form.Subscribe`)
- How to set a single field (`form.setFieldValue`, updater functions, nested paths, `field.setValue`)
- How to set multiple fields / reset the form (`form.reset`)
- Whether `setFieldValue` triggers validation, and how to disable it (`dontUpdateMeta`, `dontValidate`) and trigger it manually (`form.validateField`)
- Array helpers (`pushFieldValue`, `insertFieldValue`, `removeFieldValue`, etc.)
- How to listen for value changes (field `listeners` prop, `form.useStore`)

Also adds the new guide to `docs/config.json` under the React Guides section.

## Summary by CodeRabbit

* **Documentation**
  * Added a new "Getting and Setting Field Values" guide to the React documentation.
  * Covers initializing field values (including async patterns), retrieving current values, updating values (with updater functions, nested/array paths, and meta/validation options), and reactive listening/subscriptions for value changes.

### DIFF
--- a/docs/config.json
+++ b/docs/config.json
@@ -104,6 +104,10 @@
               "to": "framework/react/guides/basic-concepts"
             },
             {
+              "label": "Getting and Setting Field Values",
+              "to": "framework/react/guides/field-values"
+            },
+            {
               "label": "Form Validation",
               "to": "framework/react/guides/validation"
             },
--- a/docs/framework/react/guides/field-values.md
+++ b/docs/framework/react/guides/field-values.md
@@ -1,0 +1,329 @@
+---
+id: field-values
+title: Getting and Setting Field Values
+---
+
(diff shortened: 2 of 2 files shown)


crutchcorn:
These docs feel duplicative from other docs we have now. Closing

PR #2118, opened 2026-04-10
docs: general docs updates

harry-whorlow:
## Summary by CodeRabbit

* **Documentation**
  * Replaced the old combined SSR/meta-framework guide with separate TanStack Start, Next.js, and Remix guides.
  * Added detailed integration guides for TanStack Start, Next.js (App Router + Server Actions), and Remix covering server/client validation and state merging.
  * Added “Why Index as Key?” for array examples.
  * Added submission-handling docs describing canSubmitWhenInvalid across frameworks.
  * Updated React guides navigation and removed an outdated transform reference.


docs/framework/react/guides/arrays.md
```diff
@@ -52,6 +52,10 @@ Finally, you can use a subfield like so:
 </form.Field>
 ```
 
+## Why Index as Key?
+
+You may notice that these examples use `key={i}` the array index as the key prop. React's documentation generally advises _against_ using array indices as keys when items can be reordered or deleted. TanStack Form is an exception to this rule. Because field names in TanStack Form arrays are index-based, using the array index as `key` is required, it keeps React component instances, form store state, and field names in sync.
```
LeCarbonator, line 57: This is only partially true. Yes, the field data is identified by the index, but the rendering usually isn't. Many reordering libraries break when using the index as key.

We addressed that for React, so while it may impact performance slightly, you can use arbitrary key props. I believe it's still broken in Vue.

docs/framework/react/guides/nextjs.md
```diff
@@ -0,0 +1,162 @@
...
+        )}
+      </form.Subscribe>
+    </form>
+  )
+}
+```
+
+### useTransform
+
+you may have noticed util function `useTransform` being used throughout these examples, it's primary responsibility is the merging of the server and client state. Under the hood it is a useCallback whose deps are that of the server state, when the server state changes it will automatically patch the client state.
```
LeCarbonator, line 150:
```suggestion
you may have noticed the util function `useTransform` being used throughout these examples. Its primary responsibility is the merging of the server and client state. Under the hood, it is a useCallback whose deps are that of the server state, when the server state changes it will automatically patch the client state.
```

This change needs to be applied to the other SSR docs.

LeCarbonator (commented):
Looks solid! The array explanation might need some rephrasing with the new context, hopefully my explanation is clear enough.

docs/framework/react/guides/arrays.md
harry-whorlow: ah when was that changed?

PR #2120, opened 2026-04-11
fix(form-core): clear stale onBlur errors on re-submission

### LINKED ISSUE #2034: Field-level validation does not run on submit.
### Describe the bug

Field-level validation does not run on submit.

https://codesandbox.io/p/sandbox/tanstack-bug-jydplk

### Steps to reproduce

1. Set select to PIN
2. Press Submit
3. Notice how REQUIRED error shows up
4. Change select to Card
5. Press Submit
6. Notice how REQUIRED is still there and the form cannot be submitted

### Expected behavior

As a user, I expected the form to be able to be submitted when I press Submit for the second time since the field state should be valid on the second submit.

Every time

### Screenshots or Videos

_No response_

### Platform

- Browser: Chrome
- OS: MacOS
- Browser Version: 145.0.7632.45

### TanStack Form adapter

react-form

### TanStack Form version

v1.28.0

### TypeScript version

v4.4.4

### Additional context

_No response_

restareaByWeezy:
## Changes

Fixes #2034

When a field has an `onBlur` validator that produces an error, subsequent submit attempts get blocked. The `canSubmit` check in `_handleSubmit` returns early before `validateAllFields` has a chance to re-run and clear the stale error.

This patch skips the early return on re-submission (`submissionAttempts > 1`) so that `validateAllFields` actually runs again. First-submission behavior is unchanged.

Added a regression test that reproduces the exact scenario from the issue — conditional field validation with type switching.

## Checklist

## Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Stale field-level validation errors from prior onBlur validation are now cleared on resubmission; fields are re-validated so subsequent submits reflect current errors.

* **Tests**
  * Added a regression test ensuring onBlur validators re-run and stale errors don't block a later successful submit.

* **Chores**
  * Prepared a patch release note marking this fix.


etodanik:
For context: This is actually fixing a nasty bug that cost us an App Review submission cycle. Would love to see a maintainer take a look.

Issue #419, opened 2023-08-30
[Feature Request]: Form Groups


crutchcorn:
### Description

When building a form stepper, like so:

![image](https://user-images.githubusercontent.com/9100169/228072732-7514d64c-b6f5-471e-9b0b-e598597202a7.png)

It's common for each step to have its own form. However, this complicates the form submission and validation process by requiring you to add complex logic.

Ideally, it would be nice to have a FormGroup where you could validate the group, but not the form itself - `submit` the value and move on to the next step.

### Describe the solution you'd like

What if there was something like this psuedo-API?

```tsx
import { useState } from "react";
import { useForm } from "@tanstack/react-form";
// Import might change
import { za } from "@tanstack/zod-form-adapter";

const StepperForm = () => {
  const [step, setStep] = useState(0);
  const form = useForm({ onSubmit: () => {} });

  return (
    <form.Provider>
      <form {...form.getFormProps()}>
        <div>
          {step === 1 && (
            <form.FormGroup
              onGroupSubmit={() => {
                setStep(step + 1);
              }}
              preserveState={true}
            >
              <form.Field
                name="step1.name"
                onSubmit={za(z.string().thing())}
                onGroupSubmit={za(z.string().thing())}
              >
                <Input />
              </form.Field>
            </form.FormGroup>
          )}

          {step === 2 && (
            <form.FormGroup onGroupSubmit={submit} preserveState={true}>
              <form.Field
                name="step2.name"
                onSubmit={za(z.string().thing())}
                onGroupSubmit={za(z.string().thing())}
              >
                <Input />
              </form.Field>
            </form.FormGroup>
          )}
        </div>
      </form>
    </form.Provider>
  );
};
```

Christian24:
Just wanted to leave some additional context: We sometimes have quite complex forms that have a lot of fields that are only visible on a certain condition, say for example a billing and a delivery address, they are basically the same, but one is only used under the condition that the two are not the same. So FormGroups can be a great way to structure a form as well.

timothyac:
+1

At AWS, a common pattern we implement is a “Wizard” pattern which guides a user through a multi-step form. 

Example: https://cloudscape.design/examples/react/wizard.html

francesconi:
I'm currently working on a multi-step form, similar to a wizard. The proposed `form.FormGroup` component would simplify the code a lot.

crutchcorn:
[...]

PR #2128, opened 2026-04-16
Form Groups

crutchcorn:
This PR implements #419 as outlined in the issue.

> This feature might take us some time to implement due to the sheer amount of edgecases and API surface area we have to cover to mark this idea as stable and ready-for-usage.
>
> We will very likely ask community members to help us test when we get farther along, but will do so in #419.
>
> Thank you for your patience!

# Core TODOs

> This is what will take up a majority of our time.

- [x] Core FormGroupAPI class and base usage
- [x] Fixed types with FormGroupAPI
- [x] `core` package integration tests
- [x] `core` package type tests
- [x] Refactor state to live in the form instead of in `FormGroup`
- [ ] Add the ability to `getFormGroupStore` or something like it so users can access group validity externally

# Framework support

> This list looks long and intimidating, but once we have a single framework's support checkboxed, it's mostly boilerplate for the rest of the frameworks.
>
> As a result, we will be prioritizing React adapter support as our first target.

- [x] `react` adapter base code
- [x] `react` adapter integration tests
- [x] `react` adapter type tests
- [x] `react` adapter examples
- [x] `react` adapter docs

---

- [x] `preact` adapter base code
- [x] `preact` adapter integration tests
- [x] `preact` adapter type tests
- [x] `preact` adapter examples
- [x] `preact` adapter docs

---

- [x] `angular` adapter base code
- [x] `angular` adapter integration tests
- [x] `angular` adapter examples
- [x] `angular` adapter docs

---

- [x] `lit` adapter base code
- [x] `lit` adapter integration tests
- [x] `lit` adapter examples
- [x] `lit` adapter docs

---

- [x] `solid` adapter base code
- [x] `solid` adapter integration tests
- [x] `solid` adapter type tests
- [x] `solid` adapter examples
- [x] `solid` adapter docs

---

- [x] `svelte` adapter base code
- [x] `svelte` adapter integration tests
- [x] `svelte` adapter examples
- [x] `svelte` adapter docs

---

- [x] `vue` adapter base code
- [x] `vue` adapter integration tests
- [x] `vue` adapter examples
- [x] `vue` adapter docs

## Summary by CodeRabbit

* **New Features**
  * Form Groups: sub-forms with independent submit flow, group-level validators, lifecycle/state flags, and group-scoped error distribution.
  * Field-level onGroupSubmit listener and per-call validation controls for selective field/group validation and optional skipping of form-level error-map updates.

* **Documentation**
  * Added "Form Groups" guides and Multi‑Step Wizard examples across multiple frameworks.

* **Tests**
  * Expanded tests covering group submission, validation, error propagation, and lifecycle behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/form/pull/2128?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


crutchcorn:
Note to self: We need to extend `FormGroupApi` to add `implements FieldManipulator` as part of our Form-like contract.

crutchcorn:
Another note to self: Work on the validators for `FormGroupApi` live in this branch: [form-group-validation](https://github.com/TanStack/form/tree/form-group-validation)

To ensure that this PR is clean enough to review by other maintainers. Will merge these two branches when I'm more confident about that code.

gaganref:
I’m trying out `FormGroup` from the PR build for a multi-step onboarding flow where each step calls `group.handleSubmit()` and the Continue button uses `group.formState.isSubmitting`, similar to how I’d use form-level `isSubmitting` in a normal TanStack Form.

The submit itself seems to work, but the React state for `group.formState.isSubmitting` looks stale/flaky.

Here is a reduced version of the pattern I’m using. My real flow has multiple fields and step validators, but the submit-state issue reproduces with one grouped field and an async `onGroupSubmit`.

```tsx
<form.FormGroup
  name="profile"
  onGroupSubmit={async () => {
    await new Promise((resolve) => setTimeout(resolve, 1200))
  }}
>
  {(group) => (
    <form
      onSubmit={(event) => {
        event.preventDefault()
        event.stopPropagation()
        void group.handleSubmit()
      }}
    >
      <form.Field name="profile.name">
        {(field) => (
          <input
            name={field.name}
            value={field.state.value}
            onBlur={field.handleBlur}
            onChange={(event) => field.handleChange(event.target.value)}
          />
        )}
      </form.Field>

      <button disabled={group.formState.isSubmitting} type="submit">
        {group.formState.isSubmitting ? "Saving..." : "Continue"}
      </button>
    </form>
  )}
</form.FormGroup>
```

What I’m seeing:

- On a valid submit, `group.formState.isSubmitting` does not reliably rerender while the async `onGroupSubmit` is running.
- With validators in the real flow, the first invalid submit can show the loading state and stay there longer than expected.
- Subsequent invalid submits behave differently, which made it feel like the state was only updating when some other field/group meta update happened.

From reading the source, it looks like `FormGroupApi._handleSubmit` updates submit lifecycle state on `formStateStore`, but `useFormGroup` only subscribes to `formGroupApi.store` value/meta state. The adapter exposes `formState` through a getter, but React is not subscribed to `formStateStore`.

This local patch fixed the behavior for me:

```ts
const reactiveFormState = useStore(
  formGroupApi.formStateStore,
  (state) => state,
)

...

get formState() {
  return reactiveFormState
}
```

After that, `group.formState.isSubmitting` rerenders on both edges of the submit lifecycle and behaves much closer to form-level `isSubmitting`.

Does that seem like the right fix for the React adapter?

crutchcorn:
@gaganref WOW is this helpful!!! That does indeed seem like the right fix to me. Thank you so much for helping test and debug this!!

gaganref:
I also hit one related default-value edge case while testing the same reduced `profile.name` example above.

If I use the group-level `defaultValue`:

```tsx
<form.FormGroup
  name="profile"
  defaultValue={{
    name: "",
  }}
>
```

with this setup, child field changes did not reliably stick for me. I first noticed it with grouped select fields, where selecting an option immediately returned to the placeholder. I also reproduced it with text fields in the first step: typing into the field was overwritten, as if the group default was being reapplied.

From the source, it looks like `FormGroup.update()` checks the group field meta:

```ts
if (!this.state.meta.isTouched && this.options.defaultValue !== undefined) {
  const formField = this.form.getFieldValue(this.name)

  if (!evaluate(formField, opts.defaultValue)) {
    this.form.setFieldValue(this.name, opts.defaultValue as never, {
      dontUpdateMeta: true,
      dontValidate: true,
      dontRunListeners: true,
    })
  }
}
```

Moving the default back to the parent form fixed it for me, and that also matches the wizard example in this PR:

```ts
useForm({
  defaultValues: {
    profile: {
      name: "",
    },
  },
})
```

That said, it would be nice if group-level defaults worked too, since the API shape mirrors the existing form-level and field-level default patterns.

Small style question: for object groups, would `defaultValues` be a better prop name than `defaultValue`, since the group represents a subtree of form values rather than one scalar field value?

Is `FormGroup defaultValue` intended to be supported for object groups with nested fields? If yes, should it only initialize when the group value is `undefined`, or should it use the derived group touched state instead of only `state.meta.isTouched`?

crutchcorn:
@gaganref sorry, had a lot of loose threads to hunt down related to this PR.

Regarding:

```tsx
<form.FormGroup
  name="profile"
  defaultValue={{
    name: "",
  }}
>
```

We're likely not going to support this. @LeCarbonator has made some pretty compelling private comments about removing `defaultValue` even from `field` in Form v2. Would love to hear your thoughts/concerns about that though, feel free to drop the feedback on that idea here and/or in our Discord.

> FWIW if we ever rethought that idea, we can always improve `defaultValue` in FormGroup in a minor (feature) release of Form

crutchcorn:
Also @gaganref, the issue about `isSubmitting` you reported should be fixed across all framework adapters. Thank you for the help again!

gaganref:
Thanks for the response,

Re: group-level `defaultValue` no strong feelings here, I almost always set defaults at the form level anyway. I only flagged it because the current behavior felt broken, and it stood out as inconsistent with the field-level defaultValue pattern. If the v2 direction is to drop defaultValue from field entirely, that works for me too .

monarcode:
Any chance we can see a basic example of this new form-group api in action? To get an idea of how it'll work

gaganref:
There is an example for each framework in `examples/{framework}/multi-step-wizard` in the repo

PR #2133, opened 2026-04-18
fix(form-core): prevent removeValue from auto-touching shifted siblings

### LINKED ISSUE #2131: removeValue on array field incorrectly sets isTouched=true on all shifted siblings
### Describe the bug

Calling `form.removeValue(index)` on an array field flips `isTouched` to `true` on every sibling at index ≥ removed index. Even when the user has never interacted with those fields.

The bug lives in `@tanstack/form-core`, so every adapter (react, vue, solid, svelte, angular, lit) is affected.

https://github.com/ntatoud/repro-tanstack-form-removevalue-touched.git

### Steps to reproduce

1. `git clone https://github.com/ntatoud/repro-tanstack-form-removevalue-touched.git`
2. `npm install && npm run test` (deterministic vitest assertion)
3. `npm run dev`
4. Click 'remove index 2' button.
5. Notice how all shifted elements are not marked as touched. 

### Expected behavior

`isTouched` stays `false` on every remaining item: The user never touched anything.

```
values  = ['a', 'b', 'd', 'e']
touched = [false, false, false, false]
```

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/3b884c33-a58f-4892-90ab-86c547180d1e

### Platform

- OS: macOS 26.2
- Node: 24.15.0
- Browser: any (not related to browser)

### TanStack Form adapter

react-form

### TanStack Form version

1.29.0

### TypeScript version

6.0.2

### Additional context

Adapter-agnostic: bug is in `form-core`, not the React bindings

## Proposed fix

In `validateArrayFieldsStartingFrom`, bypass `validateField` when the field already has a mounted instance and call `fieldInstance.validate(cause)` directly. This preserves validation, skipping the auto-touch:

```ts
batch(() => {
  fieldsToValidate.forEach((nestedField) => {
    fieldValidationPromises.push(
      Promise.resolve().then(() => {
        const fieldInstance = this.fieldInfo[nestedField]?.instance
        if (fieldInstance) {
          return fieldInstance.validate(cause)
        }
        return this.validateField(nestedField, cause)
      }),
    )
  })
})
```

Verified locally by patching `node_modules/@tanstack/form-core/dist/esm/FormApi.js`.

The same fix likely applies to `swapFieldValues` / `moveFieldValues` / `insertFieldValue` if they call `validateArrayFieldsStartingFrom` (worth auditing as a follow-up)

mixelburg:
### What changed

In `validateArrayFieldsStartingFrom`, fields with an existing instance now call `fieldInstance.validate(cause)` directly instead of going through `this.validateField()`. This avoids the auto-touch side-effect in `validateField` that marks fields as touched even when the user never interacted with them.

### Why

Calling `form.removeValue(index)` on an array field was flipping `isTouched` to `true` on every sibling at index ≥ removed index. The bug lives in `validateField` which auto-touches any field that has an instance but isn't yet touched.

### How to reproduce

1. Create a form with an array field
2. Call `form.removeValue(0)` without interacting with other fields
3. Observe that `isTouched` becomes `true` on all shifted siblings

Closes #2131

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation behavior for array fields so removing or reordering items no longer incorrectly marks sibling fields as “touched.”
  * Preserves fields' prior touched state when validating already-mounted nested fields, reducing unexpected validation side effects during array operations.

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -1582,6 +1582,7 @@ export class FormApi<
     field: TField,
     index: number,
     cause: ValidationCause,
+    options?: { avoidTouch?: boolean },
   ) => {
     const currentValue = this.getFieldValue(field)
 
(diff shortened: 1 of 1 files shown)


LeCarbonator:
Hmm ... this has probably been here for a long while, but you're right that it makes little sense to touch the fields. However, the implementation seems to either be overcomplicated for it, or more likely, it should target a different function of the library.

I'll look into it and let you know.

mixelburg:
Closing due to inactivity. As noted, the implementation may be overcomplicated for the change. Happy to reopen with a simpler approach.

PR #2136, opened 2026-04-21
fix(form-core): clear stale onMount error on linked-field revalidation

### LINKED ISSUE #2124: Field-level `onMount` errors are never cleared on linked-field revalidation
### Describe the bug

When a field has a field-level `onMount` validator and is linked to another field, the onMount error is never cleared, even when the other validators on that field run successfully via the linked-field mechanism and report the field as valid.

The result is that `field.state.meta.errors` keeps surfacing the stale mount-time error and `isValid` stays false indefinitely or until you touch the field. This also propagates to form.state.isValid, breaking submit-button enablement and any UI that depends on per-field validity.

Possibly related to #689, #903 

https://stackblitz.com/edit/vitejs-vite-ywiic5qt?file=src%2FApp.tsx

### Steps to reproduce

1. Open the StackBlitz repro.
2. Observe the initial state: password is empty, confirmPassword defaults to "password". The confirmPassword field's onMount validator correctly reports "Passwords do not match" in `errorMap.onMount`, and isValid is false. This is expected.
3. Type "password" into the Password input. Do not touch the Confirm Password input.
4. The onChange validator on confirmPassword runs via `onChangeListenTo: ["password"]` and correctly returns undefined — `errorMap.onChange` is now undefined.
5. Bug: `errorMap.onMount` still shows `"Passwords do not match"`, and both `confirmPassword.isValid` and `form.isValid` remain false, even though both validators on the field, given the current values, return undefined.

### Expected behavior

As a user, I expected the `onMount` error to clear once the field's other validators run successfully via the linked-field mechanism, but I am seeing the `onMount` error persist indefinitely (or until it's touched), leaving the field marked as invalid even though all of its validators agree that it is valid.

Every time

### Platform

- OS: macOS
- Browser: Google Chrome
- Version: 146.0.7680.178

### TanStack Form adapter

react-form

### TanStack Form version

v.1.29.0

### TypeScript version

v6.0.2

### Additional context

Not sure whether to classify this as a bug or a design choice. `onMount` only clearing on direct value changes could be a deliberate "the user must acknowledge this field" rule. But it does feel like an edge case worth revisiting, given that `onChangeListenTo` / `onBlurListenTo` exist specifically for cross-field patterns.

If the current behavior is intentional, an opt-in to let successful linked-field revalidation clear the stale onMount slot would be very helpful. The only workaround I've found is to manually clear the slot via setMeta from a listener on the source field, which is fragile and easy to forget.

mixelburg:
Fixes #2124

When a field has an  validator and is linked to another field via /, the onMount error was never cleared even when the linked-field validator confirmed the field was valid. This left  and  stuck.

The fix: after a sync or async validator runs on a field (or its linked field) and reports no error, if an  entry is present, clear it — the later validation supersedes the mount-time check. Same pattern already used for clearing submit errors on change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced form field validation to properly manage error states during synchronous and asynchronous validation cycles, ensuring outdated errors are cleared when fields become valid.
  * Improved error cleanup logic to prevent stale validation messages from one validation run from persisting into subsequent validation attempts, resulting in more accurate field status displays.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -1735,17 +1735,28 @@ export class FieldApi<
 
         // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
         if (field.state.meta.errorMap?.[errorMapKey] !== newErrorValue) {
-          field.setMeta((prev) => ({
-            ...prev,
-            errorMap: {
+          field.setMeta((prev) => {
+            const nextErrorMap = {
               ...prev.errorMap,
               [errorMapKey]: newErrorValue,
-            },
-            errorSourceMap: {
-              ...prev.errorSourceMap,
-              [errorMapKey]: newSource,
-            },
-          }))
+            }
+            // Clear stale onMount error when a later validator confirms the field is valid
+            if (
+              !newErrorValue &&
+              errorMapKey !== 'onMount' &&
+              prev.errorMap.onMount
+            ) {
+              nextErrorMap.onMount = undefined
+            }
+            return {
+              ...prev,
+              errorMap: nextErrorMap,
+              errorSourceMap: {
+                ...prev.errorSourceMap,
+                [errorMapKey]: newSource,
+              },
+            }
+          })
         }
         if (newErrorValue) {
           hasErrored = true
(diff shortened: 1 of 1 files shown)


mixelburg:
Closing due to inactivity. Over 2 months with no review.

PR #2140, opened 2026-04-30
fix(form-core): treat non-plain objects with no own enumerable keys as unequal in evaluate

### LINKED ISSUE #1628: `evaluate` does not correctly compare objects which breaks formOptions updates
### Describe the bug

This is essentially the same issue as https://github.com/TanStack/store/issues/218, but specific to `useForm`'s form options updates. If default values were changed and the changed value is an object that does not return property keys via `Object.keys`—such as `Date`, `File`, or `Temporal.Duration`/`Temporal.PlainDate`—then `evaluate` returns true and `shouldUpdateValues` returns false. This causes the form API to not be updated, so nothing rerenders despite a prop change.

https://codesandbox.io/p/devbox/modest-jang-4635xz

### Steps to reproduce

1. Create a field that uses one of these types of objects as a value. e.g. Temporal.Duration, or Date. 
2. Give it a default value
3. Update the default value (without touching the input)

### Expected behavior

The field should update its untouched default value accordingly to the updated default value.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Arc (Chromium)
- Version: 138.0

### TanStack Form adapter

react-form

### TanStack Form version

v1.12.3

### TypeScript version

v5.8.2

### Additional context

_No response_

Zelys-DFKH:
## 🎯 Changes

Fixes #1628.

`evaluate()` compares objects by iterating `Object.keys()`. That works fine for plain objects, but for anything whose state lives only in getters, `Object.keys()` returns `[]`. The key-iteration loop then vacuously succeeds and two distinct instances are treated as equal, regardless of what they actually contain. `Temporal.Duration`, `RegExp`, any class that exposes values through getters — they all hit this. The result is that form updates get dropped silently when a field value changes to a new instance of such a type.

The fix is a single guard placed after the key-count check:

```ts
if (
  keysA.length === 0 &&
  !Array.isArray(objA) &&
  !Array.isArray(objB) &&
  (Object.getPrototypeOf(objA) !== Object.prototype ||
    Object.getPrototypeOf(objB) !== Object.prototype)
) {
  return false
}
```

When both objects have zero own enumerable keys and at least one isn't a plain `{}`, they fall back to referential inequality (`Object.is` at the top already returned `false`). Arrays get an explicit carve-out so `evaluate([], [])` still returns `true`.

PR #1939 adds a specific `instanceof Blob` guard for `File`/`Blob`. This guard is more general and covers the same case, plus `Temporal`, `RegExp`, `Error`, and anything else with a getter-only design. If #1939 merges first, its `instanceof Blob` check fires before reaching this one, so there's no conflict. If this merges first, `File`/`Blob` is already covered.

One thing to flag: two `RegExp` literals with identical source and flags (like `/abc/g` vs `/abc/g`) now return `false`, since they're different instances. Previously they returned `true` vacuously, which was also wrong: it suppressed updates when a regex field actually changed. The behavior is now in the right direction. A dedicated `instanceof RegExp` handler (like the existing `Date` one) could add semantic equality if that turns out to be useful.

## ✅ Checklist

## 🚀 Release Impact

## Release Notes

* **Bug Fixes**
  * Fixed form validation comparison logic that was incorrectly treating distinct non-plain objects as equivalent. This affected special object types including Temporal, RegExp, and classes with getter-only properties.
  * Strengthened test coverage to ensure accurate object comparison behavior across all supported object types.

### DIFF
--- a/.changeset/fix-evaluate-getter-only-objects.md
+++ b/.changeset/fix-evaluate-getter-only-objects.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/form-core': patch
+---
+
+`evaluate()` incorrectly treated distinct non-plain objects with no own enumerable keys (Temporal types, RegExp, getter-only class instances) as equal because the key-iteration loop vacuously succeeded. A guard now returns `false` for such objects, falling back to referential inequality.
--- a/packages/form-core/src/utils.ts
+++ b/packages/form-core/src/utils.ts
@@ -464,6 +464,20 @@ export function evaluate<T>(objA: T, objB: T) {
     return false
   }
 
+  // Two distinct non-plain, non-array objects with no own enumerable keys cannot
+  // be compared by key iteration — the loop below would vacuously succeed and
+  // treat them as equal regardless of their internal state. This covers Temporal
+  // types, RegExp, and any class that exposes values only through getters.
+  if (
+    keysA.length === 0 &&
+    !Array.isArray(objA) &&
+    !Array.isArray(objB) &&
+    (Object.getPrototypeOf(objA) !== Object.prototype ||
+      Object.getPrototypeOf(objB) !== Object.prototype)
+  ) {
+    return false
+  }
+
   for (const key of keysA) {
     // performs recursive search down the object tree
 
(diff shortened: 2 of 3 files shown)


crutchcorn (approved):
This is great, thanks!

PR #2143, opened 2026-05-04
fix(FieldApi): fix race condition when using onChangeListenTo

Pascalmh:
## 🎯 Changes

Repro: https://stackblitz.com/edit/github-lxc6tmac?file=src%2Findex.tsx

ToDo:
- [x] fixes a bug where `isValidating` remians `true` when multiple fields are set simultaniasly in while using `onChangeListenTo` on multiple fields and running `onChangeAsync` checks

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Tests**
  * Added a Vitest regression test simulating rapid/autofill updates across linked fields to ensure async validation runs only once and clears correctly.
* **Bug Fixes**
  * Introduced per-field validation counting so "validating" flags are set/cleared only for fields with async validators, preventing stuck validating states and restoring accurate form validity and submitability.
* **Chores**
  * Added a changeset to publish a patch.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -1854,18 +1854,25 @@ export class FieldApi<
     // Check if there are actual async validators to run before setting isValidating
     // This prevents unnecessary re-renders when there are no async validators
     // See: https://github.com/TanStack/form/issues/1130
-    const hasAsyncValidators =
-      validates.some((v) => v.validate) ||
-      linkedFieldValidates.some((v) => v.validate)
+    const hasAsyncValidators = validates.some((v) => v.validate)
+    const linkedFieldsWithAsyncValidators = linkedFieldValidates.some(
+      (v) => v.validate,
+    )
+      ? Array.from(
+          new Set(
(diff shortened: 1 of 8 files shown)


packages/form-core/src/FieldApi.ts
```diff
@@ -631,6 +631,12 @@ export type FieldMetaBase<
    * A flag indicating whether the field is currently being validated.
    */
   isValidating: boolean
+  /**
+   * @private
+   * Counter for tracking active async validations to prevent race conditions
+   * when multiple validations finish at the same time.
+   */
+  validationCount: number
```
LeCarbonator: Would be good to prepend a `_` to make it clearer it's not for user consumption. Otherwise, people might get the idea that this is some tracked state for the amount of validation attempts or something similar.


```suggestion
  _pendingValidationsCount: number
```

LeCarbonator (approved):
The implementation isn't the prettiest, but considering the bug it's fixing, it's quite small! Beside the comment, I think this is alright to go.


Generally, a React adapter unit test would also be great to make sure that this error isn't adapter-specific.

PR #2149, opened 2026-05-06
fix: remove errant `Field.Field` usage and `useForm().useField()`

crutchcorn:
This PR removes the ability to do:

```
<form.Field children={field => <field.Field/>}/>
```

And:

```
const {useField} = useForm();
```

Which I think was errantly added, as I can't think of any reliable reason why we'd want to support this.

## Summary by CodeRabbit

* **Breaking Changes**
  * Field hook no longer exposes a Field sub-component on its return value — update code that accessed Field via the hook.
  * Form API no longer provides a useField helper on the extended surface; only Field component and other helpers remain.
  * Several exported type aliases and the augmented field API were removed — update affected type usages.
  * Svelte form API now exposes Field as a component-style member instead of the previous factory-style helper.

* **Chores**
  * ESLint configs updated to disable the no-use-context rule for TS/TSX.


crutchcorn:
I'm very suspicious about removing these changes. Because we have so many of these instances in our code it makes me think that we added this purposefully at some point, but I can't fathom why.

packages/react-form/src/createFormHook.tsx
```diff
@@ -66,7 +66,7 @@ type UnwrapDefaultOrAny<DefaultT, T> = [DefaultT] extends [T]
   : T
 
 function useFormContext() {
-  const form = useContext(formContext)
+  const form = use(formContext)
```
LeCarbonator: We can just do that? Does that not risk React 17 / 18 compat?

LeCarbonator (changes requested):
Looks good! Resolve the note since it may affect React 17/18. Otherwise, feel free to merge anytime.

PR #2190, opened 2026-05-21
fix(form-core): bump _arrayVersion when async defaultValues update array fields

### LINKED ISSUE #2178: Async initial array values are not set
### Describe the bug

Until version 1.31.0, it's been possible to set async initial values that are in the form of an array. In version 1.32.0, this does not seem to work anymore. I assume [this PR](https://github.com/TanStack/form/pull/2172) is the reason, however I've not looked into it. 

https://stackblitz.com/edit/tanstack-form-skfjtgfa?file=src%2Findex.tsx

### Steps to reproduce

1. Open the StackBlitz
2. Notice that no person is displayed after 3 seconds, when the async state is updated
3. You may click the "Add person"-button to notice that there suddenly is 2 persons in the list

### Expected behavior

When the async state is resolved, I would expect the state to be updated in the form as well. This works up until version 1.31.0.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome

### TanStack Form adapter

react-form

### TanStack Form version

1.32.0

### TypeScript version

6.0.3

### Additional context

_No response_

mixelburg:
Fixes #2178

When async `defaultValues` resolve and update the form, array fields rendered with `mode="array"` were not re-rendering. This is a regression from #2172 which changed the array field reactivity signal from `Object.keys(value ?? []).length` to `state.meta._arrayVersion`. The `_arrayVersion` counter is bumped on explicit mutations (push/insert/remove/swap/move) but not when `defaultValues` change asynchronously.

The fix: after the store update in `FormApi.update()`, iterate over registered fields and bump `_arrayVersion` for any whose current value is an array. Added a regression test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Array-mode fields now reliably update and re-render when their values change (including when async/default values are applied), fixing stale array renders.

* **Tests**
  * Added a regression test verifying array fields re-render when default values are provided after initial render.

* **Chores**
  * Added a patch changeset recording the fix.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/form/pull/2190?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/nine-beds-shout.md
+++ b/.changeset/nine-beds-shout.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/form-core': patch
+---
+
+Fix array async default values not updating
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -1512,6 +1512,17 @@ export class FormApi<
       )
     })
 
+    if (shouldUpdateValues) {
+      const helper = metaHelper(this)
+      for (const fieldKey of Object.keys(
+        this.fieldInfo,
+      ) as DeepKeys<TFormData>[]) {
+        if (Array.isArray(this.getFieldValue(fieldKey))) {
+          helper.bumpArrayVersion(fieldKey)
+        }
+      }
+    }
+
     formEventClient.emit('form-api', {
       id: this._formId,
       state: this.store.state,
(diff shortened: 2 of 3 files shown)


crutchcorn (approved):
This fix makes a lot of sense. Thanks for the PR and tests

PR #2197, opened 2026-05-28
fix(form-core): use equals() for value-type equality (Temporal, Decimal.js)

### LINKED ISSUE #2195: Value equality does not handle Temporals
### Describe the bug

The `evaluate` method in form-core, which does deep equality checking, does not work for Temporals.  Previously, it always treated two Temporal instances (e.g., two `Temporal.PlainDate`) as equal, because they have no enumerable properties.  With the recent fix #2140, it will now always treat temporal instances as *not* equal.

The new behavior is clearly better, but it's still incorrect.  A date field backed by a Temporal instance will be considered "dirty" even if the user resets it back to the original date.

`evaluate` already special-cases some built-in types, like `Date`.  With Temporal advancing to Stage 4 and shipping in browsers, it might make sense to special-case them as well.

OTOH, given that people will be using polyfills for a while yet, it may be best to duck-type.  Here's one snippet that could be added to `evaluate`:

```ts
const objAToStringTag = objA[Symbol.toStringTag];
if (objAToStringTag.startsWith('Temporal.') && objAToStringTag === objB[Symbol.toStringTag]) {
  // Handle Temporal types (they all have .equals())
  return objA.equals(objB);
}
```

https://stackblitz.com/edit/vitejs-vite-1zaxj72u?file=src%2Fmain.ts

### Steps to reproduce

The simple repro is:

```ts
evaluate(Temporal.PlainDate.from('2026-01-01'), Temporal.PlainDate.from('2026-01-01')) === false
```

I think any form that has a field with a Temporal value (e.g., a `Temporal.PlainDate` instance) will be subject to dirty-checking issues because of this.

### Expected behavior

```ts
evaluate(Temporal.PlainDate.from('2026-01-01'), Temporal.PlainDate.from('2026-01-01')) === true
```

None

### Screenshots or Videos

_No response_

### Platform

This should be universal!

### TanStack Form adapter

None

### TanStack Form version

1.32.1

### TypeScript version

_No response_

### Additional context

Currently we are using this patch (`pnpm patch`) of `@tanstack/form-core`.  It's a bit more general, and a bit less safe, because it just checks for the `.equals()` method.  But it has the benefit of working for other types, like `Decimal.js`.

```patch
diff --git a/dist/cjs/utils.cjs b/dist/cjs/utils.cjs
--- a/dist/cjs/utils.cjs
+++ b/dist/cjs/utils.cjs
@@ -198,6 +198,10 @@ function evaluate(objA, objB) {
   if (objA instanceof Date && objB instanceof Date) {
     return objA.getTime() === objB.getTime();
   }
+  // Handle Temporal types (they all have .equals())
+  if (typeof objA.equals === "function" && objA.constructor === objB.constructor) {
+    return objA.equals(objB);
+  }
   if (objA instanceof Map && objB instanceof Map) {
     if (objA.size !== objB.size) return false;
     for (const [k, v] of objA) {
diff --git a/dist/esm/utils.js b/dist/esm/utils.js
--- a/dist/esm/utils.js
+++ b/dist/esm/utils.js
@@ -196,6 +196,10 @@ function evaluate(objA, objB) {
   if (objA instanceof Date && objB instanceof Date) {
     return objA.getTime() === objB.getTime();
   }
+  // Handle Temporal types (they all have .equals())
+  if (typeof objA.equals === "function" && objA.constructor === objB.constructor) {
+    return objA.equals(objB);
+  }
   if (objA instanceof Map && objB instanceof Map) {
     if (objA.size !== objB.size) return false;
     for (const [k, v] of objA) {
```

I wish JavaScript had a convention like for a method that checks value equality, like `Symbol.for('valueEquals')` or something

mixelburg:
Fixes #2195

The `evaluate` function in form-core used key-based deep equality for all objects. This broke for value types like `Temporal` (TC39 Stage 4, now in browsers/Deno/Node) and `Decimal.js`, which have no enumerable properties — two instances with the same value would always be considered unequal.

This PR adds a duck-typed check: if both objects have the same constructor and the left-hand object has an `equals()` method, we delegate to it. This covers `Temporal.PlainDate`, `Temporal.Instant`, `Temporal.Duration`, `Decimal.js`, and similar value types.

The change is placed before the Map/Set checks and after the Date check, so more specific `instanceof` checks take precedence.

```ts
evaluate(Temporal.PlainDate.from('2026-01-01'), Temporal.PlainDate.from('2026-01-01'))
// was: false
// now: true
```

All 439 existing tests pass; 1 new test added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved equality comparison to properly handle custom objects with custom equality methods.

* **Tests**
  * Added test coverage for custom object equality comparison.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/form/pull/2197?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/form-core/src/utils.ts
+++ b/packages/form-core/src/utils.ts
@@ -498,6 +498,13 @@ export function evaluate<T>(objA: T, objB: T) {
     return objA.getTime() === objB.getTime()
   }
 
+  if (
+    typeof (objA as any).equals === 'function' &&
+    objA.constructor === objB.constructor
+  ) {
+    return (objA as any).equals(objB)
+  }
+
   if (objA instanceof Map && objB instanceof Map) {
     if (objA.size !== objB.size) return false
     for (const [k, v] of objA) {
--- a/packages/form-core/tests/utils.spec.ts
+++ b/packages/form-core/tests/utils.spec.ts
@@ -747,6 +747,18 @@ describe('evaluate', () => {
     expect(dateObjectFalse).toEqual(false)
   })
 
+  it('should test equality for objects with an equals() method (e.g. Temporal, Decimal.js)', () => {
+    class ValueType {
+      constructor(private val: number) {}
+      equals(other: ValueType) {
+        return this.val === other.val
+      }
+    }
+
+    expect(evaluate(new ValueType(1), new ValueType(1))).toEqual(true)
+    expect(evaluate(new ValueType(1), new ValueType(2))).toEqual(false)
+  })
+
   it('should test equality between Map objects', () => {
     const map1 = new Map([
       ['a', 1],


mixelburg:
Closing due to inactivity — over 5 weeks with no review or feedback. Happy to reopen if there's interest.

PR #2206, opened 2026-06-04
fix: re-export useSelector and document useStore migration

hemraj-007:
- Re-export `useSelector` from TanStack Store in react/preact/vue/solid/svelte form packages
- Keep deprecated `useStore` export for backward compatibility
- Add `form.useSelector` on Vue, Solid, and Svelte (deprecate `form.useStore`)
- Update docs, examples, changeset, and add `exports.test.ts` for react-form

## Summary by CodeRabbit

* **New Features**
  * Added selector-based subscriptions via `form.useSelector` across React, Preact, Vue, Solid, and Svelte adapters, and re-exported `useSelector` at the package level where applicable.

* **Deprecations**
  * `form.useStore` is deprecated and remains as an alias; use `form.useSelector` instead.

* **Documentation**
  * Updated reactivity, validation, SSR, and reference guides/examples to use `useSelector`, including migration notes and selector/compare guidance.

* **Tests**
  * Added coverage to verify `useSelector`/`useStore` export surface.

### DIFF
--- a/.changeset/use-selector-export.md
+++ b/.changeset/use-selector-export.md
@@ -1,0 +1,9 @@
+---
+'@tanstack/react-form': patch
+'@tanstack/preact-form': patch
+'@tanstack/vue-form': patch
+'@tanstack/solid-form': patch
+'@tanstack/svelte-form': patch
+---
+
+Re-export `useSelector` from TanStack Store adapters and document migration from deprecated `useStore` (fixes #2203).
--- a/docs/framework/preact/guides/basic-concepts.md
+++ b/docs/framework/preact/guides/basic-concepts.md
@@ -236,7 +236,7 @@ function App() {
 
 ## Reactivity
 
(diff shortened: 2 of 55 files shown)


hemraj-007:
Updated prose to use useSelector(form.store, …) for consistency

crutchcorn:
Hey, this is genuinely helpful and self-starting; we appreciate the contribution. I should have clarified our position on `useStore` much earlier.

FWIW, I think this is not the path we were intending (`useStore` is safe for all v1 usage, but we **will** likely migrate to `useSelector` in v2), but is actually the path I think makes the most sense. I'll run tests and we can merge

mwmcode:
> Hey, this is genuinely helpful and self-starting; we appreciate the contribution. I should have clarified our position on `useStore` much earlier.
> 
> FWIW, I think this is not the path we were intending (`useStore` is safe for all v1 usage, but we **will** likely migrate to `useSelector` in v2), but is actually the path I think makes the most sense. I'll run tests and we can merge

so in future, react-store will be a peer dep?

PR #2211, opened 2026-06-14
fix(form-core): keep onSubmit error when blurring an unchanged field

### LINKED ISSUE #1242: Field submit error is removed on field blur without a change
### Describe the bug

A submit error is removed on blur of a field when the value didn't change.

https://codesandbox.io/p/devbox/hungry-sanderson-ycs6rd

### Steps to reproduce

1. Press submit without entering a name
2. Focus the name input field
3. Leave focus without a change

### Expected behavior

The error which occured on submit is not removed on blur when the value during focus didn't change.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: WIndows
- Browser: Chrome

### TanStack Form adapter

react-form

### TanStack Form version

1.0.2

### TypeScript version

_No response_

### Additional context

_No response_

chatman-media:
Closes #1242

## Problem

A field-level `onSubmit` error is wrongly cleared when the user focuses and blurs the field **without changing its value**. The submit error should only disappear once the user actually enters a (valid) value.

## Root cause

In `FieldApi`, the logic that clears a stored `submit` error fired for any cause other than `'submit'`:

```ts
this.state.meta.errorMap?.[submitErrKey] &&
cause !== 'submit' &&
!hasErrored
```

`handleBlur` runs validation with `cause: 'blur'`, which satisfied `cause !== 'submit'`, so a plain blur (no edit) cleared the submit error.

## Fix

Clear the stored submit error only on an actual value `change`:

```ts
cause === 'change' &&
```

A `blur` (or any non-value cause) no longer drops the submit error. Entering a valid value still clears it as documented.

## Tests

Two tests added to `packages/form-core/tests/FieldApi.spec.ts`:
- `should not clear onSubmit errors on blur when the value did not change` — **fails on `main`**, passes with the fix
- `should clear onSubmit errors once a valid value is entered` — guards the documented clear-on-change path

Full `form-core` suite green (107 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed form validation to properly preserve submit errors during blur and non-value interactions. Submit errors are now cleared only when valid values are entered, improving form error feedback accuracy.

* **Tests**
  * Extended test coverage for submit error behavior, verifying errors persist on blur without changes and clear when valid values are entered.

### DIFF
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -1347,14 +1347,17 @@ export class FieldApi<
 
     /**
      *  when we have an error for onSubmit in the state, we want
-     *  to clear the error as soon as the user enters a valid value in the field
+     *  to clear the error as soon as the user enters a valid value in the field.
+     *  This must only happen on a value `change` - clearing it on `blur` (or any
+     *  other non-value cause) would wrongly drop the submit error when the user
+     *  focuses and leaves the field without editing it.
      */
     const submitErrKey = getErrorMapKey('submit')
 
     if (
       // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
       this.state.meta.errorMap?.[submitErrKey] &&
-      cause !== 'submit' &&
+      cause === 'change' &&
       !hasErrored
     ) {
       this.setMeta((prev) => ({
--- a/packages/form-core/tests/FieldApi.spec.ts
+++ b/packages/form-core/tests/FieldApi.spec.ts
@@ -2098,6 +2098,64 @@ describe('field api', () => {
     expect(field.getMeta().errors).toStrictEqual(['first name is required'])
   })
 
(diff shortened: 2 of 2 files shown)


MILLERMARRU (commented):
Checked `ValidationCause` and it's `'change' | 'blur' | 'submit' | 'mount' | 'server' | 'dynamic'`, so the old `cause !== 'submit'` was clearing a stale onSubmit error on blur, mount, server and dynamic validations too, not just on an actual value change. `cause === 'change'` is the right fix, it only clears the error for the one cause that actually means "the user entered something new," which is what the surrounding comment already claimed the code was doing.

The two new tests cover the pair that matters, blurring an untouched field keeps the error, entering a valid value clears it, and neither the FieldApi type nor any caller elsewhere seems to rely on the old broader clearing behavior.

PR #2218, opened 2026-06-20
fix(form-core): preserve reset(newValues) when update() is called with stale options

### LINKED ISSUE #1681: form.reset during onSubmit ignores new default values

LeSingh1:
## Summary

Fixes #1681 — `form.reset(newValues)` called inside `onSubmit` is ignored and the form reverts to its original default values.

### Root cause

`reset(values)` (with explicit new values) does two things:
1. Updates `this.options.defaultValues = values` in-place
2. Calls `this.baseStore.setState(...)` with the new values

After the `onSubmit` promise resolves, React re-renders (because `isSubmitting` flipped back to `false`). The framework then calls `formApi.update(opts)` inside `useIsomorphicLayoutEffect` — where `opts` is the **original** options object captured in the render closure, still holding the pre-reset `defaultValues`.

Inside `update()`:
- `oldOptions = this.options` → has the **new** defaultValues (set by reset)
- `this.options = options` → overwrites back to **original** defaultValues
- `options.defaultValues !== oldOptions.defaultValues` → `true`
- `!this.state.isTouched` → `true` (reset cleared touched state)
- `shouldUpdateValues = true` → form values reset to original defaults ❌

### Fix

Add a private `_defaultValuesOverridden` flag on `FormApi`. When `reset(values)` is called with explicit values (and `!keepDefaultValues`), set the flag to `true`. In `update()`, if the flag is set, skip the `shouldUpdateValues` branch and preserve the `defaultValues` that `reset()` stored on `this.options`. The flag is cleared after one `update()` cycle so subsequent prop-driven `defaultValues` changes continue to work correctly.

## Test plan

- [x] Added a new unit test in `packages/form-core/tests/FormApi.spec.ts` that directly simulates the race condition: calls `reset(newValues)` then `update(originalOpts)` and asserts that values are **not** reverted
- [x] Existing test `form should reset default value when resetting in onSubmit` (both core and react-form suites) continues to pass
- [x] All 499 `@tanstack/form-core` tests pass
- [x] All 123 `@tanstack/react-form` tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition where resetting a form with new values could have those values unexpectedly reverted to stale defaults.

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -1066,6 +1066,12 @@ export class FormApi<
    * @private
    */
   private _devtoolsSubmissionOverride: boolean
+  /**
+   * @private
+   * Tracks whether `reset(values)` was called with explicit new default values.
+   * When true, `update()` must not overwrite those values with stale prop values.
+   */
+  private _defaultValuesOverridden: boolean = false
 
   /**
    * Constructs a new `FormApi` instance with the given form options.
@@ -1751,6 +1757,21 @@ export class FormApi<
     // Options need to be updated first so that when the store is updated, the state is correct for the derived state
     this.options = options
 
+    // If reset(newValues) was called, the incoming `options` still carries the
+    // original (pre-reset) defaultValues from the render closure. We must not
+    // let those stale values overwrite what reset() set, so we skip the
+    // shouldUpdateValues branch and clear the flag for the next render.
+    if (this._defaultValuesOverridden) {
+      this._defaultValuesOverridden = false
+      // Keep the defaultValues that reset() stored on this.options so that
+      // future renders compare against them correctly.
+      this.options = {
+        ...options,
+        defaultValues: oldOptions.defaultValues,
+      }
+      return
+    }
+
     const shouldUpdateValues =
       options.defaultValues &&
       !evaluate(options.defaultValues, oldOptions.defaultValues) &&
@@ -1815,6 +1836,7 @@ export class FormApi<
         ...this.options,
         defaultValues: values,
       }
+      this._defaultValuesOverridden = true
     }
 
     this.baseStore.setState(() => {
--- a/packages/form-core/tests/FormApi.spec.ts
+++ b/packages/form-core/tests/FormApi.spec.ts
@@ -144,6 +144,33 @@ describe('form api', () => {
     form.handleSubmit()
   })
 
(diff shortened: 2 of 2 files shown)


MILLERMARRU (commented):
The direction is right, but the one-shot flag reverts if `update()` fires twice in a row with the same stale props before a genuinely new set arrives, which is a realistic case (a parent re-rendering for an unrelated reason re-runs the child's layout effect with the same props reference, or StrictMode double-invoking it).

Reimplemented both this PR's logic and #2235's (which targets the same #1681, comparing against the previous *update's* props instead of a one-shot flag) side by side and ran the exact scenario, reset, then two consecutive `update()` calls with unchanged stale props:

```
[PR#2218 (flag)] after reset: { name: 'new-default' }
[PR#2218 (flag)] after update #1 (stale): { name: 'new-default' }
[PR#2218 (flag)] after update #2 (same stale props again): { name: 'original' }
[PR#2218 (flag)] reset value survived double stale update: false

[PR#2235 (tracking)] after reset: { name: 'new-default' }
[PR#2235 (tracking)] after update #1 (stale): { name: 'new-default' }
[PR#2235 (tracking)] after update #2 (same stale props again): { name: 'new-default' }
[PR#2235 (tracking)] reset value survived double stale update: true
```

The flag gets consumed by the first `update()` call after `reset()`, so by the second consecutive call with the same (unchanged) props, `_defaultValuesOverridden` is already `false` and `update()` falls through to the normal `shouldUpdateValues` comparison, comparing the stale incoming props against the `defaultValues` that `reset()` set, which reads as a real change and reverts.

#2235 doesn't have this gap because it tracks what the *previous update() call* actually saw, not a one-shot signal, so a repeated identical `update()` call keeps comparing against the same baseline instead of consuming a flag. It also doesn't need the early `return` this PR adds (which skips the rest of `update()`, including `shouldUpdateState`, on the cycle it fires), since it folds the reset-preservation into the same `this.options` assignment that runs either way.

Since #2235 is already open against the same issue and (at least in this scenario) doesn't have the gap, might be worth comparing the two directly before merging either.

PR #2219, opened 2026-06-20
fix(form-core): export FormListenerFieldProps to fix TS4023 declaration emit

### LINKED ISSUE #2210: Exported formOptions with listeners fails declaration emit: FormListenersPropsField cannot be named
### Describe the bug

When exporting an inferred `formOptions(...)` value that includes `listeners`, TypeScript fails during declaration emit with `TS4023`.

The inferred exported type references TanStack Form's internal `FormListenersPropsField` type from `@tanstack/form-core/dist/esm/FormApi`. Since that type is not publicly exported/nameable, TypeScript cannot serialize the exported `formOptions(...)` value into a `.d.ts` file.

This can happen even without explicitly setting `declaration: true`. For example, `composite: true` is enough to reproduce because composite TypeScript projects emit declarations by default.

Minimal failing code:

```ts
import { formOptions } from '@tanstack/react-form'

export const reproFormOptions = formOptions({
  defaultValues: {
    country: '',
    province: '',
  },
  listeners: {
    onChange: ({ formApi }) => {
      formApi.setFieldValue('province', '')
    },
  },
})
```

The error is:

```text
src/form-options.ts(12,14): error TS4023: Exported variable 'reproFormOptions' has or is using name 'FormListenersPropsField' from external module ".../node_modules/.pnpm/@tanstack+form-core@1.33.0/node_modules/@tanstack/form-core/dist/esm/FormApi" but cannot be named.
```

https://codesandbox.io/p/github/jzxhuang/tanstack-form-form-options-type-bug/main?import=true

### Steps to reproduce

See codesandbox, snippet above, or https://github.com/jzxhuang/tanstack-form-form-options-type-bug

### Expected behavior

TypeScript should be able to emit declarations for an exported `formOptions(...)` value that uses form `listeners`.

In particular, the inferred public type of `formOptions(...)` should not require consumers' declaration emit to name an internal, non-exported TanStack Form type.

None

### Screenshots or Videos

_No response_

### Platform

```text
Reproduces with TypeScript 6.0.3.

Also verified with TypeScript 5.8.3 using:

pnpm --package typescript@5.8.3 dlx tsc -p tsconfig.app.json --pretty false --noEmitOnError
```

### TanStack Form adapter

None

### TanStack Form version

1.33.0

### TypeScript version

6.0.3

### Additional context

_No response_

mixelburg:
Fixes #2210

Extract the inline `{formApi, fieldApi}` callback props type from `FormListeners` into a named exported `FormListenerFieldProps` interface.

Previously, exporting a `formOptions()` result that included `listeners` failed with TS4023 because TypeScript could not serialize the anonymous inline type into a `.d.ts` file. By giving the type an explicit name and exporting it, declaration emit can now reference it directly.

All 438 tests pass across TS 5.4–5.8.

## Summary by CodeRabbit

* **Refactor**
  * Form listener callback types have been updated to use a unified type structure, ensuring consistent generic type parameters across `onChange`, `onBlur`, and `onFieldUnmount` callbacks. This improves overall type safety and consistency for developers working with form event handlers.

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -229,7 +229,7 @@ export interface FormValidators<
   onDynamicAsyncDebounceMs?: number
 }
 
-interface FormListenersPropsGroup<
+export interface FormListenersPropsGroup<
   TFormData,
   TOnMount extends undefined | FormValidateOrFn<TFormData>,
   TOnChange extends undefined | FormValidateOrFn<TFormData>,
@@ -260,7 +260,7 @@ interface FormListenersPropsGroup<
   groupApi: AnyFormGroupApi
 }
 
-interface FormListenersPropsField<
+export interface FormListenersPropsField<
   TFormData,
   TOnMount extends undefined | FormValidateOrFn<TFormData>,
   TOnChange extends undefined | FormValidateOrFn<TFormData>,


mixelburg:
Rebased on latest main ✅

Note: Upstream has added group-related listeners since this PR was opened, so I adapted the fix to export both  and  (keeping upstream's naming convention) rather than renaming. This still resolves the TS4023 declaration emit issue.

PR #2235, opened 2026-07-09
fix(form-core): keep reset(values) baseline on update() with unchanged props defaultValues

### LINKED ISSUE #1681: form.reset during onSubmit ignores new default values

xianjianlf2:
Fixes #1681

## Problem
Calling `form.reset(newValues)` inside `onSubmit` while a component subscribes to `isDirty` reverts the form to the **original** default values instead of the new ones.

The `isDirty` change re-renders the component, which re-runs the `formApi.update(opts)` layout effect with the unchanged props `defaultValues`. `update()` compared the incoming `defaultValues` against `this.options.defaultValues` — but `reset(values)` had just mutated that field, so the unchanged props looked like a real change and (because `reset` also cleared `isTouched`) clobbered the freshly-reset values back to the original defaults.

## Fix
Track the `defaultValues` seen on the *previous* `update()` call and compare against that instead. When the props `defaultValues` did not actually change, preserve the current baseline (e.g. one set by `reset(values)`) so the values — and `isDirty` / `isDefaultValue` — stay correct after a reset. Fixed in `form-core`, so the fix covers all framework adapters.

## Tests
- **`form-core`**: regression test for the `update()` / `reset()` race.
- **`react-form`**: integration test reproducing the exact issue (component-level `isDirty` subscription + `reset` in `onSubmit`).

Both tests fail on `main` and pass with this change. Full `form-core` (499) and `react-form` (124) suites pass with no type errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where form values could revert after resetting during submit and then updating the form.
  * Preserved reset values so the form stays in the expected state after submission.
  * Improved dirty-state handling so forms return to pristine correctly after a reset.

### DIFF
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -1068,6 +1068,15 @@ export class FormApi<
   private _devtoolsSubmissionOverride: boolean
 
   /**
+   * The `defaultValues` carried by the options on the previous `update()` call
+   * (i.e. the last values that came from the framework adapter / props). Used
+   * to detect a genuine props change without being fooled by `reset(values)`,
+   * which legitimately mutates `this.options.defaultValues`.
+   * See https://github.com/TanStack/form/issues/1681
+   */
+  private _prevUpdateDefaultValues: TFormData | undefined = undefined
+
+  /**
    * Constructs a new `FormApi` instance with the given form options.
    */
   constructor(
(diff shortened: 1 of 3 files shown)


MILLERMARRU (commented):
The root cause here is that `update()` used to compare incoming `defaultValues` against `this.options.defaultValues`, but `reset(values)` also writes to that exact same field. In React, `update(opts)` runs in a layout effect on every render with whatever `defaultValues` the component's props currently hold, so after `reset()` inside `onSubmit` triggers a re-render, the next `update()` call still carries the *original* props value, which by then looks different from the just-reset `this.options.defaultValues`, and the old code read that as a genuine props change and overwrote the reset.

Introducing `_prevUpdateDefaultValues` to track what `update()` itself last saw, separately from `this.options.defaultValues`, is the right fix, since it's comparing against the thing that actually represents "props," not the internal baseline that other calls are also allowed to mutate. The ternary that decides the next `this.options` is a bit dense but it holds up when traced through: if props genuinely changed (or happen to already match `oldOptions.defaultValues`, which is just a no-op either way), take the incoming options as-is, otherwise keep the current `defaultValues` on `this.options` instead of the incoming one, so a same-props re-render after `reset()` can't clobber it.

I checked `evaluate()`'s handling of the very first call, where `_prevUpdateDefaultValues` is still `undefined`. `Object.is` fails first since one side is a real object, then the `typeof x !== 'object'` guard on `undefined` (`typeof undefined === 'undefined'`) returns `false` immediately, so the first `update()` always reads as a real change, which is what you want for the initial sync.

Both regression tests are good, the form-core one exercises the store directly and the react-form one goes through an actual `isDirty` subscription and `handleSubmit`, which matters here specifically because that's what re-triggers the layout effect that was clobbering the reset in the original bug report. Testing it end to end through a real re-render is the only way this particular bug would have shown up in the first place.

PR #2243, opened 2026-07-13
fix: bump array version on reset so array fields re-render

### LINKED ISSUE #2228: Resetting array fields does not trigger full re-render
### Describe the bug

When using `form.reset({  fieldName: arr })` to set the state of an array mode field to an array with fewer elements than the field's `defaultValue`, the elements `defaultValue[arr.length]` through `defaultValue[defaultValue.length - 1]` will still render.

**Example**: 
* defaultValue = [1, 2, 3]
* form.reset({ fieldName: [4] })
* rendered values: 4, 2, 3.

**Notes**:
* This only applies if the value of `fieldName` has not been altered since the initial page load. If you alter the value (e.g. by adding another element) before resetting, the reset value will be properly rendered.
* The actual state of the form is correctly updated by the reset. You can see in my provided code that if you click reset, then submit, the submitted value is as expected (and the form re-renders to show the correct state).

https://codesandbox.io/p/sandbox/36vt5h

### Steps to reproduce

In the example sandbox, click the reset button and observe that 3 elements are still rendered instead of one.

### Expected behavior

Only the elements present in the value passed to reset should render.

Every time

### Screenshots or Videos

_No response_

### Platform

* OS: macOS
* Browser: chrome

### TanStack Form adapter

React

### TanStack Form version

1.33.0

### TypeScript version

_No response_

### Additional context

_No response_

hemraj-007:
## Summary

- Bump `_arrayVersion` after `form.reset()` and `form.resetField()` when the target field is an array.
- Fix React array fields not re-rendering when a reset shortens the array (e.g. `[1,2,3]` → `[4]` previously still rendered 3 items).
- Fixes #2228.

## Test Plan

- `pnpm run test:lib` in `packages/form-core`
- `pnpm run test:lib` in `packages/react-form`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed array-based form fields not re-rendering correctly after resetting the form or an individual field.
  * Array fields now immediately reflect changes in length and updated values after a reset.

### DIFF
--- a/.changeset/fix-array-reset-rerender.md
+++ b/.changeset/fix-array-reset-rerender.md
@@ -1,0 +1,8 @@
+---
+'@tanstack/form-core': patch
+'@tanstack/react-form': patch
+---
+
+Bump `_arrayVersion` when resetting array fields so adapters re-render after `form.reset()` or `form.resetField()` changes array length.
+
+Fixes #2228
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -1847,6 +1847,15 @@ export class FormApi<
         fieldMetaBase,
       })
     })
+
+    const helper = metaHelper(this)
+    for (const fieldKey of Object.keys(
+      this.fieldInfo,
+    ) as DeepKeys<TFormData>[]) {
+      if (Array.isArray(this.getFieldValue(fieldKey))) {
+        helper.bumpArrayVersion(fieldKey)
+      }
+    }
   }
 
   /**
@@ -2954,6 +2963,10 @@ export class FormApi<
             : prev.values,
       }
     })
+
+    if (Array.isArray(this.getFieldValue(field))) {
+      metaHelper(this).bumpArrayVersion(field)
+    }
   }
 
   /**
--- a/packages/form-core/tests/FormApi.spec.ts
+++ b/packages/form-core/tests/FormApi.spec.ts
@@ -125,6 +125,23 @@ describe('form api', () => {
     })
   })
 
+  it('should bump array version when resetting to a shorter array', () => {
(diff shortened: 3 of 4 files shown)


MILLERMARRU (commented):
Went to check why bumping `_arrayVersion` specifically is what's needed here, and it's because `react-form`'s `useField` deliberately subscribes to `state.meta._arrayVersion` instead of `state.value` for `mode="array"` fields, there's already a comment there referencing #1925 explaining it's an intentional optimization to avoid re-rendering on every child property change. That means for an array field, React genuinely does not know the array changed unless `_arrayVersion` itself changes, no matter what happened to the actual array reference or length. `form.reset()`/`resetField()` changing the array without touching that counter is exactly the kind of thing that selector is blind to, so the fix is targeting the right mechanism, not just papering over a symptom.

Checked the ordering in both call sites too. In `reset()`, the new loop runs after the `baseStore.setState` that actually commits the new values, so `this.getFieldValue(fieldKey)` inside the loop is reading the post-reset array, not stale state. Same story in `resetField`.

One minor thing, not a correctness issue: `resetField` isn't wrapped in `batch()`, so setting `fieldMetaBase[field]` to `defaultFieldMeta` (which resets `_arrayVersion` to `0`) and then calling `bumpArrayVersion` right after are two separate store commits rather than one. Since the array value itself is already correct as of the first commit, this shouldn't produce wrong output, just possibly one extra render pass in cases where `_arrayVersion` wasn't already `0` going in. Worth a look if this file already batches similar multi-step meta+value updates elsewhere, but not something I'd block on.

Good that the React test renders through an actual `mode="array"` field and checks the DOM list length after reset, rather than asserting on internal state, since the whole point of the fix is what the UI shows, not just what the store holds.

PR #2244, opened 2026-07-13
fix(form-core): clear stale onMount error on linked-field revalidation

### LINKED ISSUE #2124: Field-level `onMount` errors are never cleared on linked-field revalidation

xianjianlf2:
## Problem
A field-level `onMount` error is cleared when the field's own value changes, but not when the field is revalidated indirectly through `onChangeListenTo` / `onBlurListenTo`. After a passing cross-field revalidation the stale mount-time error lingers, so the field (and the whole form) stays permanently invalid (#2124).

## Root cause
`FormApi.setFieldValue` clears `errorMap.onMount` on a direct value change, treating onMount as a mount-time snapshot. Linked-field revalidation re-runs the field's validators without changing its value, so that clearing never runs.

## Fix
Thread an `isLinkedField` flag into `validateFieldFn`; on a passing linked-field revalidation, clear the stale `onMount` error/source slots. This mirrors the existing value-change clear and the adjacent "clear stale submit error on successful revalidation" logic in the same function.

## Testing
Added a regression test in `packages/form-core/tests/FieldApi.spec.ts` (red before, green after). Full form-core suite (504 tests) passes with no type errors. Changeset included.

Closes #2124

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed stale field errors that could remain after linked-field validation succeeds.
  - Fields now correctly clear outdated mount-time errors when related field changes trigger successful revalidation.
  - Form and field validity states now update correctly without requiring the affected field’s value to change.

- **Tests**
  - Added coverage for linked-field revalidation and stale error removal.

### DIFF
--- a/.changeset/clear-stale-onmount-linked-revalidation.md
+++ b/.changeset/clear-stale-onmount-linked-revalidation.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/form-core": patch
+---
+
+fix(form-core): clear a stale field-level `onMount` error when a linked-field revalidation (`onChangeListenTo`/`onBlurListenTo`) passes, so the field is no longer left permanently invalid
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -1292,6 +1292,7 @@ export class FieldApi<
       const validateFieldFn = (
         field: AnyFieldApi,
         validateObj: SyncValidator<any>,
+        isLinkedField = false,
       ) => {
         const errorMapKey = getErrorMapKey(validateObj.cause)
 
@@ -1334,6 +1335,32 @@ export class FieldApi<
         if (newErrorValue) {
           hasErrored = true
         }
+
+        // A direct value change clears the mount-time error in
+        // `setFieldValue`. When a field is revalidated indirectly via
+        // `onChangeListenTo`/`onBlurListenTo`, its own value never changed, so
+        // that clearing never ran and a now-stale `onMount` error would linger
+        // forever (leaving the field permanently invalid even though its
+        // validators now pass). Clear it once a passing linked revalidation
+        // supersedes it.
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        if (
+          isLinkedField &&
+          !newErrorValue &&
+          field.state.meta.errorMap?.onMount
+        ) {
+          field.setMeta((prev) => ({
+            ...prev,
+            errorMap: {
+              ...prev.errorMap,
+              onMount: undefined,
+            },
+            errorSourceMap: {
+              ...prev.errorSourceMap,
+              onMount: undefined,
+            },
+          }))
+        }
       }
 
       for (const validateObj of validates) {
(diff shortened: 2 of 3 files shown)


MILLERMARRU (commented):
Went to check `setFieldValue` in FormApi.ts to confirm the asymmetry the changeset describes is real. It does clear `errorMap.onMount` unconditionally whenever a field's own value changes, regardless of whether the field's new value actually passes validation. A linked field never goes through that path since its own value doesn't change, only the field it's listening to does, so before this fix a stale `onMount` error on the listening field had no way to ever clear short of the user editing that field directly, which defeats the point of `onChangeListenTo` existing in the first place for things like password-confirmation pairs.

Worth noting the new code is actually more conservative than the thing it's filling a gap for: `setFieldValue` clears `onMount` on any edit whether or not the new value is valid, this only clears it when the linked revalidation for that specific cause has just passed (`!newErrorValue`). That's arguably the more correct behavior of the two, since it won't clear a mount error out from under a field that's still genuinely invalid.

Scoping the clear to just the `onMount` key inside `errorMap`/`errorSourceMap` rather than resetting the whole map means any errors from other validation causes on that field survive, which matters since a field can accumulate errors from multiple sources independently.

The test builds the actual password/confirm-password scenario this is meant to fix rather than a synthetic minimal repro, mounts with a real mismatch so `onMount` genuinely fires first, then changes the linked field and checks the field is fully valid afterward (`errors` empty, `isValid` true, and the form's own `isValid` too), which is the right level to assert at since the bug's real symptom is a permanently-invalid form the user has no way out of.

PR #2246, opened 2026-07-16
fix(form-core): apply async defaultValues to untouched fields after other fields are edited

### LINKED ISSUE #2229: Async default values not applied to untouched fields after other fields are edited
### Describe the bug

When a form is created with `defaultValues` that include a field whose value arrives asynchronously, that field does not receive its updated default value after the async data loads if the user has already edited another field in the meantime.

The async field stays empty instead of being populated with the loaded default.

This is important to make forms usable while small parts of it are still loading, avoiding the need to block the entire form to workaround the issue. In our case, the user should be able to modify other fields, even if the form is not ready for submission yet (we disable the submit button and the pending fields).

https://codesandbox.io/p/sandbox/goofy-chatelet-q8r3ff

### Steps to reproduce

In the reproduction example:
1. Render the form (the async field starts with an empty default).
2. Before the async value resolves (3s), type into the `name` field.
3. Wait for the async value to load.
4. Observe that `asyncField` remains empty instead of showing `"loaded default"`.

### Expected behavior

When async data arrives and `defaultValues` are updated on re-render, untouched fields should receive their new default values even if other fields have already been edited.

### Screenshots or Videos

https://github.com/user-attachments/assets/f2892c24-d4a9-4a6e-9fee-b8800d3d44db

### Platform

TanStack Form adapter: React (@tanstack/react-form)
Browser: Chrome (latest)
OS: macOS

### TanStack Form adapter

None

### TanStack Form version

1.33.0

### TypeScript version

_No response_

### Additional context

_No response_

xianjianlf2:
## 🎯 Changes

Fixes #2229. `FormApi.update()` previously gated default-value merging on the form-level touched state, so touching one field prevented asynchronously arriving defaults from applying to every other untouched field.

- Drop the form-level touched gate and merge updated defaults per field instead.
- Preserve user-edited values for touched fields.
- Apply new default values to untouched fields, including fields that have not mounted yet.
- Add regression coverage for both scalar fields and array fields.

## ✅ Checklist

- [x] I have tested this code locally.

## 🚀 Release Impact

## Notes

The existing behavior of not wiping an edited field during update is preserved. The new behavior only applies newly arrived defaults to fields that are still untouched.

### DIFF
--- a/.changeset/async-default-untouched-fields.md
+++ b/.changeset/async-default-untouched-fields.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/form-core': patch
+---
+
+fix(form-core): apply updated async `defaultValues` to untouched fields even after another field has been edited, instead of skipping the whole update once any field is touched (#2229)
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -1754,14 +1754,32 @@ export class FormApi<
 
     const shouldUpdateValues =
       options.defaultValues &&
-      !evaluate(options.defaultValues, oldOptions.defaultValues) &&
-      !this.state.isTouched
+      !evaluate(options.defaultValues, oldOptions.defaultValues)
 
     const shouldUpdateState =
       !evaluate(options.defaultState, oldOptions.defaultState) &&
       !this.state.isTouched
 
     if (!shouldUpdateValues && !shouldUpdateState) return
+
+    // Fields the user has already edited must keep their edited values, but
+    // untouched fields (including ones that have not mounted yet) should still
+    // receive their newly-arrived default values — even when a *different*
+    // field has already been touched. See #2229.
+    let nextValues = options.defaultValues
+    if (shouldUpdateValues && this.state.isTouched) {
+      for (const [fieldName, fieldMeta] of Object.entries(
+        this.state.fieldMeta,
+      )) {
+        if ((fieldMeta as AnyFieldLikeMeta | undefined)?.isTouched) {
+          nextValues = setBy(
+            nextValues,
+            fieldName,
+            getBy(this.state.values, fieldName),
+          )
+        }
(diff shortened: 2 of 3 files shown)


MILLERMARRU (commented):
Checked `setBy`/`getBy` in utils.ts to confirm the merge approach actually does what the changeset claims. `setBy` rebuilds the object path-segment by path-segment and only touches the exact path you give it, everything else comes from whatever you passed as the base object. So starting `nextValues` from the full incoming `options.defaultValues` and then overwriting just the touched field paths with their current live values via `getBy(this.state.values, fieldName)` is exactly the right shape for "untouched fields get the new default, touched fields keep what the user typed," and it composes fine for nested/array paths since `setBy` doesn't care about depth.

The loop iterates `this.state.fieldMeta`, which only has entries for fields that have actually mounted or been interacted with, so a field that hasn't mounted yet simply has no entry and falls through to the new default untouched, which matches the changeset's claim about fields that "have not mounted yet." That's a detail that's easy to get wrong (checking some kind of "isMounted" flag instead of just relying on absence from the map), and this uses the simpler and correct one.

Good that the second test specifically covers an array field (`asyncItems: []` to `['a', 'b']`) rather than just a scalar, since arrays are exactly the shape most likely to be loaded asynchronously in a real form and where "touched" tracking has more edge cases than a single string field.

One thing worth a second look, not a blocker: this only preserves values for fields with an existing `fieldMeta` entry marked `isTouched`. If a field gets marked touched through something other than direct user edits (an array `insert`/`move` marking a parent array field touched as a side effect, for instance) rather than the field's own value being edited, this would also preserve that field's current value over the incoming default. That seems like the right call for an array a user has already manipulated, but wanted to flag it since it's the one place where "touched" doing double duty for both "edited" and "structurally mutated" could matter for less obvious field shapes than the ones in the tests.

PR #2247, opened 2026-07-16
fix(form-core): reset stale validation counter when shifting array meta

### LINKED ISSUE #2234: Using onChangeAsync in an array field causes isFieldsValidating to always be true after calling removeValue
### Describe the bug

When using a `validators` config which has both `onChangeListenTo` and `onChangeAsync` in a subfield of an array mode field, removing any entry but the last of the array field will cause the listened-to fields in each shifted entry to be stuck on `isValidating: true`.

This does not happen when `onChangeListenTo` is included with `onChange`.
This does not happen when `onChangeListenTo` is *not* included with `onChangeAsync`. 

Possibly related to https://github.com/TanStack/form/issues/2131 ?

https://codesandbox.io/p/devbox/array-mode-rerender-forked-56d2np?workspaceId=ws_XStUxd2y69i1A18WTfWXJw

### Steps to reproduce

In the example sandbox, click the add button twice, then remove the first row and click submit.

Notice in the console logs that all errors are empty, and that the field meta for `list[0].operator` shows `isValidating: true`, even though it does not have a validator.

Comment out line 70, repeat the above steps, and notice that the form submits as normal.

Uncomment line 70, change `onChangeAsync` on line 71 to `onChange`, repeat the above steps, and notice that the form submits as normal.

### Expected behavior

On aform submit, the listened-to field should show `isValidating: false`, and the form submission should succeed.

Every time

### Screenshots or Videos

_No response_

### Platform

* OS: macOS
* Browser: chrome

### TanStack Form adapter

react-form

### TanStack Form version

1.33.0

### TypeScript version

_No response_

### Additional context

_No response_

xianjianlf2:
## 🎯 Changes

Fixes #2234. When an array item is removed (or inserted/moved) while an async validator is in flight, `shiftMeta` copied the source field's transient `_pendingValidationsCount`/`isValidating` down to the shifted index. The pending promise is owned by the original field instance and settles its counter against the original index, so the shifted index's counter is orphaned and never decremented — leaving `form.state.isFieldsValidating` stuck `true` and blocking submission.

- `shiftMeta` now resets the transient validation state (`isValidating`, `_pendingValidationsCount`) when shifting meta between indices. Array mutations already re-trigger validation on the shifted fields, so this is safe.
- `endValidation` bails out if the field's instance is no longer registered, so a validation that resolves after its field was removed no longer resurrects deleted meta (previously threw a `TypeError`).

## ✅ Checklist

- [x] I have tested this code locally.

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where async validation state could remain stuck after removing an item from an array field during in-flight validation.
  - Ensured transient async-validation indicators are properly reset or preserved when array field metadata shifts due to move, swap, or shift operations.
- **Tests**
  - Added coverage for array remove/move/swap scenarios with in-flight async validation, verifying field-level and form-level validating status behavior.

### DIFF
--- a/.changeset/fix-array-removevalue-isvalidating-stuck.md
+++ b/.changeset/fix-array-removevalue-isvalidating-stuck.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/form-core': patch
+---
+
+Fix `isFieldsValidating` getting stuck `true` after `removeValue` on an array field while an async validation is in flight
--- a/packages/form-core/src/FieldApi.ts
+++ b/packages/form-core/src/FieldApi.ts
@@ -1394,6 +1394,11 @@ export class FieldApi<
    * Ends tracking an async validation, decrementing the counter and clearing isValidating if no validations remain.
    */
   private endValidation() {
+    // The field may have been removed (e.g. via an array `removeValue`) while
+    // this async validation was still in flight. In that case its meta has
+    // already been deleted, so bail out instead of resurrecting empty meta.
+    // See https://github.com/TanStack/form/issues/2234
+    if (this.getInfo().instance !== this) return
     this.setMeta((prev) => {
       const newCount = Math.max(0, prev._pendingValidationsCount - 1)
       return {
--- a/packages/form-core/src/metaHelper.ts
+++ b/packages/form-core/src/metaHelper.ts
@@ -110,7 +110,10 @@ export function metaHelper<
 
         const fromMeta = fromFields.get(fromKey)
         if (fromMeta) {
-          formApi.setFieldMeta(fieldKey as DeepKeys<TFormData>, fromMeta)
+          formApi.setFieldMeta(
+            fieldKey as DeepKeys<TFormData>,
+            resetTransientValidationMeta(fromMeta),
+          )
         }
       })
   }
@@ -153,8 +156,10 @@ export function metaHelper<
         formApi.getFieldMeta(swappedKey),
       ]
 
-      if (meta1) formApi.setFieldMeta(swappedKey, meta1)
-      if (meta2) formApi.setFieldMeta(fieldKey, meta2)
+      if (meta1)
+        formApi.setFieldMeta(swappedKey, resetTransientValidationMeta(meta1))
+      if (meta2)
+        formApi.setFieldMeta(fieldKey, resetTransientValidationMeta(meta2))
     })
   }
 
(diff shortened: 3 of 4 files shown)


xianjianlf2:
Addressed the CodeRabbit follow-up in 8488fc3. Move and swap now use the same transient async-validation reset as shifted array meta, so `isValidating` / `_pendingValidationsCount` are not copied to a different index. Added regression coverage for both `moveValue` and `swapValues`.

Local checks:
- `PATH=/Users/even/.nvm/versions/node/v24.18.0/bin:$PATH pnpm --dir packages/form-core test:lib tests/FieldApi.spec.ts`
- `PATH=/Users/even/.nvm/versions/node/v24.18.0/bin:$PATH pnpm --dir packages/form-core test:types:ts59`

xianjianlf2:
Addressed the same-index move/swap follow-up in `860ab18`.

What changed:
- `moveValue(i, i, { dontValidate: true })` now returns after bumping the array version, before meta shifting/resetting.
- `swapValues(i, i, { dontValidate: true })` does the same.
- Added regression coverage to verify `isValidating` and `_pendingValidationsCount` remain intact for both no-op operations.

Local check:

```sh
pnpm --filter @tanstack/form-core exec vitest run tests/FieldApi.spec.ts
```

Result: 1 file / 110 tests passed, no type errors.

MILLERMARRU (commented):
Checked how `this.getInfo().instance !== this` is used elsewhere in `FieldApi.ts` before judging this guard, and it's not a new idea being introduced here, the async validator resolution path around line 1554 already bails out of writing `setMeta` with that exact same check when a field's identity has been superseded. So `endValidation()` picking up the same guard is consistent with how the file already treats "this instance no longer owns this name" elsewhere, rather than a one-off patch.

What makes this actually close the bug rather than just move it is that `endValidation()` runs unconditionally after `Promise.all(validatesPromises)` in `validateAsync`, regardless of that existing instance check further up (which only guards the error-writing `setMeta`, not the counter decrement). So before this fix, even in the path that correctly skipped writing a stale error, the pending-validation counter on whatever now occupies that field name would still get decremented by an instance that no longer owns it, which is exactly how `isFieldsValidating` could end up permanently wrong after `removeValue` shifts indices.

The `resetTransientValidationMeta` addition in `metaHelper.ts` is the other half of the same problem: when array mutation copies old meta onto a new key, carrying over `isValidating`/`_pendingValidationsCount` would leave a counter at the new key that nothing will ever decrement, since the instance that started that validation is bailing out via the guard above instead of settling it. Resetting it at copy time and letting the shift naturally re-trigger validation on the moved field is the right call.

The `fromIndex === toIndex` / `index === secondIndex` early returns are a nice touch beyond the core fix, and the test for the same-index case (`should preserve transient async validation state when moving array field meta to the same index`) is what actually proves the reset only fires on a real move and not on a no-op call, which is easy to get backwards.

The main regression test reconstructs the real scenario end to end, an async validator on a listened-to field, in flight, then removing a different array index while it's pending, then resolving it and checking the counter settles at zero rather than asserting some internal method got called. That's the right level to test this at since the bug only shows up through that exact interleaving.

PR #2251, opened 2026-07-21
docs(philosophy): clarify 'Three.js framework adapters' to 'Three.js renderers'

kushaim:
Three.js framework adapters → Three.js renderers. Three.js is a WebGL renderer, not a framework, and the surrounding sentence uses 
enderer for everything else.


LeCarbonator:
Thanks for the amendment!

PR #2255, opened 2026-07-23
fix: require compatible withFieldGroup field types

### LINKED ISSUE #1855: BUG: Reusable forms do not have safe TypeScript types
### Describe the bug

When types of reusable form component are wider than types of parent form - it is not longer safe to use...

https://stackblitz.com/edit/tanstack-form-zwgf8aqd?file=src%2Findex.tsx

### Steps to reproduce

Run this code with React.js and click `X` button

```tsx
import { createFormHook, createFormHookContexts } from "@tanstack/react-form"

const { fieldContext, formContext } =
  createFormHookContexts()

const { useAppForm, withFieldGroup } = createFormHook({
  fieldComponents: {},
  formComponents: {},
  fieldContext,
  formContext,
})

type PasswordFields = {
  password: string | null
}

const reusableDefaultValues: PasswordFields = { password: '', }

export const FieldGroupPasswordFields = withFieldGroup({
  defaultValues: reusableDefaultValues,
  render: function Render({ group }) {
    return (
      <div>
        <h2>Reusable Fields</h2>
        <group.AppField name="password">
          {(field) => <>
            <input
              value={field.state.value ?? ""}
              onChange={(e) => field.handleChange(e.target.value)}
            />
            <button onClick={() => field.handleChange(null)}>X</button>
          </>}
        </group.AppField>
      </div>
    )
  },
})


// ======================= App Form ======================

type FormValues = {
  reusable: { password: string }
}

  reusable: { password: '', },
}

export function App() {
  const form = useAppForm({ defaultValues })

  return (
    <form.AppForm>
      <FieldGroupPasswordFields
        form={form}
        fields="reusable"
      />
      <form.Subscribe
        selector={f => f.values.reusable.password}
        children={p => <div>
          {
            // Here it CRASHES when password is null
          }
          {p.startsWith("1") ? "Valid" : "Invalid"}
        </div>}
      />
    </form.AppForm>
  )
}

```

### Expected behavior

I should see the TypeScript error when types do not match exactly!

None

### Screenshots or Videos

_No response_

### Platform

Does not matter

### TanStack Form adapter

react-form

### TanStack Form version

1.25.0

### TypeScript version

5.9.3

### Additional context

At least ts check like this should be sufficient for this issue:
```ts
type Equals<T, S> =
	[T] extends [S] ? (
		[S] extends [T] ? true : false
	) : false
```

Check this out:
```ts
// this function just reads the value
function read(arg: string) {}
read("" as ""); // valid, narrower type
read("" as string); // valid, exact type
// type 'string | null' is not assignable to parameter of type 'string'
read("" as string | null); // invalid, too wide

// this function just writes the value, type checking works in inverse way
function write(setter: (arg: string) => void) {}
// type '(arg: "") => void' is not assignable to parameter of type '(arg: string) => void'
write((arg: "") => {}); // invalid, too narrow
write((arg: string) => {}); // valid, exact type
write((arg: string | null) => {}); // valid, wider type


// this function both reads and writes the value, type checking is strict
function readWrite(arg: string, setter: (arg: string) => void) {}
readWrite("" as "", (arg: "") => {}); // invalid, too narrow for the setter
readWrite("" as string, (arg: string) => {}); // valid, exact type
readWrite("" as string | null, (arg: string | null) => {}); // invalid, too wide for the getter
```

As you can see for the `read` getter function wider types are not accepted which is perfectly valid, but narrower type are accepted.

And for the `write` setter function, vice versa narrower type are not accepted but wider types are accepted, which is also perfectly logical.

The `readWrite` function only accepts exact type, which is kinda intersection of both types.

The `FieldGroupPasswordFields` component is basically `readWrite` function, it can read the form field values and write to them. So to maintain type safety it should show typescript error when parent form part does not have exactly the same types for the fields.

xianjianlf2:
Fixes #1855

## Changes
- Add field-group-specific deep key helpers that allow parent objects to have extra fields while requiring exposed field values to be compatible.
- Use the stricter helper for `withFieldGroup` component `fields` in React, Preact, and Solid.
- Document the intentional nullability asymmetry between whole-group field matching and mapped leaf-field matching.
- Add React, Preact, and Solid type regression coverage for reusable groups that can write `null` into string-only parent fields.

## Checklist
- [x] Added or updated type regression coverage.
- [x] Kept runtime behavior unchanged; this is a type-level compatibility fix.
- [x] Verified affected package type tests locally.

## Tests
- `pnpm prettier --check packages/form-core/src/util-types.ts packages/react-form/src/createFormHook.tsx packages/react-form/tests/createFormHook.test-d.tsx packages/preact-form/src/createFormHook.tsx packages/preact-form/tests/createFormHook.test-d.tsx packages/solid-form/src/createFormHook.tsx`
- `pnpm --filter @tanstack/react-form test:types`
- `pnpm --filter @tanstack/preact-form test:types`
- `pnpm --filter @tanstack/solid-form test:types`
- `pnpm --filter @tanstack/form-core test:types`

## Release Impact
TypeScript-only fix for `withFieldGroup` field compatibility. It should prevent unsafe nullable field-group mappings without changing runtime behavior.

## Summary by CodeRabbit

* **Enhancements**
  * Improved TypeScript type checking for field-group field mappings across React, Preact, and Solid integrations.
  * Field-group mappings now support compatible nested object and array structures, while preventing unsafe parent/leaf type mismatches.
  * Mapping constraints are tightened to avoid cases where a field group could write `null`/other incompatible values into a non-compatible parent field.

* **Tests**
  * Added TypeScript type-tests covering accepted valid mappings and rejecting incompatible nullable/non-nullable and path-based vs object-based mappings.

### DIFF
--- a/packages/form-core/src/util-types.ts
+++ b/packages/form-core/src/util-types.ts
@@ -196,6 +196,60 @@ export type DeepKeysOfType<TData, TValue> = Extract<
   AnyDeepKeyAndValue<string, TValue>
 >['key']
 
+type IsEqual<TValue, TOther> = [TValue] extends [TOther]
+  ? [TOther] extends [TValue]
+    ? true
+    : false
+  : false
+
+/**
+ * Checks whether the form value can safely back a field group value.
+ * Extra form fields are allowed, but fields exposed by the field group must match exactly.
+ */
+type IsFieldGroupCompatible<TFormValue, TFieldGroupValue> = [
+  TFieldGroupValue,
+] extends [ReadonlyArray<infer TFieldGroupItem>]
+  ? [TFormValue] extends [ReadonlyArray<infer TFormItem>]
+    ? IsFieldGroupCompatible<TFormItem, TFieldGroupItem>
+    : false
+  : [TFieldGroupValue] extends [object]
+    ? [TFormValue] extends [object]
+      ? false extends {
+          [K in keyof TFieldGroupValue]-?: K extends keyof TFormValue
+            ? IsFieldGroupCompatible<TFormValue[K], TFieldGroupValue[K]>
+            : false
+        }[keyof TFieldGroupValue]
+        ? false
+        : true
+      : false
+    : IsEqual<TFormValue, TFieldGroupValue>
+
+/**
+ * The keys of an object or array, deeply nested and compatible with a field group value.
+ */
+export type DeepKeysOfFieldGroupType<TData, TValue> = unknown extends TData
+  ? string
+  : DeepKeysAndValues<TData> extends infer TDeepKeyAndValue
+    ? TDeepKeyAndValue extends AnyDeepKeyAndValue<infer TKey, infer TDeepValue>
+      ? IsFieldGroupCompatible<NonNullable<TDeepValue>, TValue> extends true
+        ? TKey
+        : never
+      : never
+    : never
+
+/**
+ * The keys of an object or array, deeply nested and compatible with a mapped field group value.
+ */
+export type DeepKeysOfFieldGroupFieldType<TData, TValue> = unknown extends TData
+  ? string
+  : DeepKeysAndValues<TData> extends infer TDeepKeyAndValue
+    ? TDeepKeyAndValue extends AnyDeepKeyAndValue<infer TKey, infer TDeepValue>
+      ? IsFieldGroupCompatible<TDeepValue, TValue> extends true
+        ? TKey
+        : never
+      : never
+    : never
+
 /**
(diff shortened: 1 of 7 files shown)


xianjianlf2:
Addressed the two CodeRabbit nits in a follow-up.

Changes:
- documented the intentional `NonNullable` asymmetry between whole-group and mapped-field deep key helpers
- added Solid Form type coverage for nullable `withFieldGroup` field mappings, matching the React/Preact regression shape

Local checks:
- `pnpm --filter @tanstack/solid-form test:types`
- `pnpm --filter @tanstack/form-core test:types`

PR #2259, opened 2026-07-29
fix(vue-form): generate an SSR-safe default formId

### LINKED ISSUE #2254: TanStack Form Vue - default formId is not SSR-safe
### Describe the bug

When binding the default form ID (in e.g. the Vue template) that TanStack Form produces using `uuid()`, there is a hydration mismatch due to this value *not* being SSR-safe. It would be great to replace the `uuid()` call with `useId()` Vue provides (see https://vuejs.org/api/composition-api-helpers.html#useid).

n/a

### Steps to reproduce

Create simplest possible form using `useForm`, add a form tag in the Vue template and bind default form ID to it (e.g. `<form :id="myForm.formId">...</form>`). Observe how you are getting hydration errors as a different value is computed on the server vs on the client.

### Expected behavior

SSR-safe default form ID generated by TanStack Form (in case user does not provide their own)

Every time

### Screenshots or Videos

_No response_

### Platform

n/a

### TanStack Form adapter

vue-form

### TanStack Form version

1.33.2

### TypeScript version

6.0.3

### Additional context

_No response_

harshit-d3v:
## 🎯 Changes

Fixes #2254.

`FormApi` falls back to `uuid()` when no `formId` is passed, and the Vue adapter never passes one — so the id differs between the server render and the client render, and binding it (`<form :id="form.formId">`) produces a hydration mismatch.

This uses Vue's `useId` for the default instead, which is exactly what the other adapters already do:

- `react-form/src/useFormId.ts` — `React.useId`, falling back to `useUUID` on React 17
- `preact-form/src/useFormId.ts` — `useId` from `preact/hooks`

Vue had no equivalent, so `packages/vue-form/src/useFormId.ts` adds one and `useForm` now passes `formId: opts?.formId ?? fallbackFormId`, mirroring `react-form/src/useForm.tsx`.

### The version constraint

`useId` was added in **vue@3.5**, but `@tanstack/vue-form` declares `peerDependencies: { "vue": "^3.4.0" }`. So a straight `import { useId } from 'vue'` would break 3.4 consumers — and a named import of a missing export is a bundler resolution error, not a runtime `undefined`. It is therefore read off the namespace and typed as optional, with the same reasoning as the existing React comment (webpack/webpack#14814):

```ts
const vueUseId: { useId?: () => string } = Vue

export function useFormId(): string {
  const useId = vueUseId.useId
  if (useId && Vue.getCurrentInstance()) {
    return useId()
  }
  return uuid()
}
```

The `getCurrentInstance()` guard matters too: outside a component instance `useId` cannot produce a stable value and warns in dev, so `useForm` called outside `setup` keeps the old uuid behaviour rather than emitting a warning.

If you would rather bump the peer range to `^3.5.0` and call `useId` unconditionally, happy to switch — this way keeps 3.4 working.

## ✅ Checklist

## 🚀 Release Impact

## Verification

`packages/vue-form/tests/useFormId.test.tsx` covers this. Run against `main` with only the test applied (i.e. reverting `useForm.tsx`), it fails on the two default-id cases while the explicit-id control still passes:

| Test | Without the fix | With the fix |
| --- | --- | --- |
| uses the provided `formId` when one is given | ✅ (control) | ✅ |
| same default `formId` across identical renders | ❌ `2902fb86-…` vs `752902fb-…` | ✅ |
| default `formId` matches between server and client | ❌ `fb8639db-…` vs `02fb8639-…` | ✅ |

The server/client case renders through `renderToString` from `vue/server-renderer` (reached via the `vue` subpath export, so it adds no new dependency) and compares against a client render.

Also run locally:

- `pnpm test:pr` → **passes** (`test:sherif`, `test:knip`, `test:docs`, `test:eslint`, `test:lib`, `test:types`, `test:build`, `build`)
- full `@tanstack/vue-form` suite → **31 passed**, no type errors

---

Disclosure: I used an AI assistant while investigating and preparing this change, including running the verification above.

## Summary by CodeRabbit

* **New Features**
  * Added a new `useFormId` composable for generating SSR-safe, stable form identifiers.
  * `useForm` now defaults to an auto-generated `formId` when none is provided.
* **Bug Fixes**
  * Prevented mismatches between server-rendered and client-hydrated form IDs.
  * Ensured safe ID generation when `useId` isn’t available (including non-component usage).
* **Tests**
  * Added Vitest coverage for provided IDs, deterministic defaults, and SSR/client consistency (including hydration without warnings).

### DIFF
--- a/.changeset/lucky-donkeys-shout.md
+++ b/.changeset/lucky-donkeys-shout.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/vue-form': patch
+---
+
+Generate the default `formId` with Vue's `useId` so it is SSR-safe, falling back to a random uuid on `vue@3.4` (where `useId` does not exist) and outside of a component instance.
+
+Fixes #2254
--- a/packages/vue-form/src/useForm.tsx
+++ b/packages/vue-form/src/useForm.tsx
@@ -3,6 +3,7 @@ import { useSelector } from '@tanstack/vue-store'
 import { defineComponent, h, onMounted } from 'vue'
 import { Field } from './useField'
 import { FormGroup } from './useFormGroup'
+import { useFormId } from './useFormId'
 import type {
   FormAsyncValidateOrFn,
   FormOptions,
@@ -268,6 +269,8 @@ export function useForm<
     TSubmitMeta
   >,
 ) {
+  const fallbackFormId = useFormId()
+
   const formApi = (() => {
     const api = new FormApi<
       TParentData,
@@ -282,7 +285,7 @@ export function useForm<
       TFormOnDynamicAsync,
       TFormOnServer,
       TSubmitMeta
-    >(opts)
+    >({ ...opts, formId: opts?.formId ?? fallbackFormId })
 
     const extendedApi: typeof api &
       VueFormApi<
--- a/packages/vue-form/src/useFormId.ts
+++ b/packages/vue-form/src/useFormId.ts
@@ -1,0 +1,24 @@
+import * as Vue from 'vue'
+import { uuid } from '@tanstack/form-core'
+
+/**
+ * `useId` was added in vue@3.5, but this package supports `vue@^3.4.0`. It is
+ * read off the namespace, typed as optional, rather than imported by name: a
+ * named import would make bundlers fail to resolve it on 3.4.
+ * Read more: https://github.com/webpack/webpack/issues/14814
+ */
+const vueUseId: { useId?: () => string } = Vue
+
+/**
+ * Returns an SSR-safe id for the form, so the server-rendered markup and the
+ * client hydration agree on it. Falls back to a random uuid when `useId` is
+ * unavailable, or when called outside of a component instance, where `useId`
+ * cannot produce a stable value.
+ */
+export function useFormId(): string {
+  const useId = vueUseId.useId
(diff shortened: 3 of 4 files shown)


harshit-d3v:
Thanks — both points addressed.

**On the hydration test (nitpick):** fair, the test compared SSR html against a fresh client render and never actually hydrated. `f3474c79` mounts the server markup with `createSSRApp().mount()` and asserts Vue raises no hydration mismatch. It's a stronger test: reverting `useForm.tsx` now fails 3 of the 4 cases (the new hydration one included) while the explicit-`formId` control still passes.

**On vue@3.4 still producing mismatched ids:** that's accurate, and it is a deliberate limitation rather than something the patch can solve. 3.4 has no `useId`, and Vue exposes nothing else that is stable across the server and client render, so on 3.4 there is no correct id to fall back to — only the existing `uuid()` behaviour, which this PR leaves untouched. The net effect is 3.5+ gets fixed and 3.4 is no worse than today.

The alternative is bumping `peerDependencies.vue` to `^3.5.0` and calling `useId` unconditionally, which fixes it everywhere the package claims to support at the cost of dropping 3.4. That's a maintainer call on the support window, so I kept the compatible option and am happy to switch if you'd prefer the bump.

crutchcorn (approved):
Thanks for this!

PR #2260, opened 2026-07-31
fix(form-core): re-run form-level validator on resubmit to clear its own stale field errors

### LINKED ISSUE #2248: Form-level validator doesn't re-run after first submit while fields hold errors — bug or by design?
### Describe the bug

With a form-level onSubmit validator only (no field-level validators) and the default validation logic:

The first submit runs the validator and populates all field errors.
After that, canSubmit is false (since `submissionAttempts` > 0 and `isValid` === false), so every subsequent submit hits the guard at the top of `_handleSubmit` and returns early — no further validation runs until the user has cleared all outstanding errors.
That part I understand is likely intended (button-disable pattern). My real question is about the next layer:

The specific question
When I set canSubmitWhenInvalid: true to allow re-submitting while invalid, the form-level validator still doesn't re-run. Tracing _handleSubmit:

```
await this.validateAllFields("submit");     // field-level only; no-op when I have no field validators
if (!this.state.isFieldsValid) return;       // <-- bails here because stale field errors remain
await this.validate("submit");               // <-- form-level schema; never reached
```
Because we only have a form-level validator, validateAllFields can't clear or recompute anything, so isFieldsValid stays false from the previous submit, and validate("submit") (the call that re-runs the form schema) is skipped.

Is this intended? Specifically: should validateAllFields returning stale field errors prevent the form-level validator from re-running on submit? It means a form configured with only a form-level validator can never re-validate on a second submit while any error is outstanding.

see below

### Steps to reproduce

```
const form = useForm({
  defaultValues: { name: '', other: '' },
  validators: {
    onSubmit: ({ value }) => {
      const errors = {};
      if (!value.name) errors.name = 'Required';
      if (!value.other) errors.other = 'Required';
      return Object.keys(errors).length ? { fields: errors } : undefined;
    },
  },
  canSubmitWhenInvalid: true,
});
```
1. submit empty -> both fields error
2. type into `name`, clear it (error clears on change)
3. submit again -> `name` error does NOT return; `other` still errors

### Expected behavior

On each submit, the form-level validator re-runs and reflects the current values (so name would error again in step 3).

None

### Screenshots or Videos

_No response_

### Platform

macOS, chrome

### TanStack Form adapter

None

### TanStack Form version

v1.29.1

### TypeScript version

v^6.0.3

### Additional context

_No response_

FrancoKaddour:
Closes #2248

## Problem

With only a form-level `onSubmit` validator (no field-level validators), the form can never re-validate on a second submit while any error is outstanding.

`_handleSubmit` runs `validateAllFields('submit')` (field-level only), then bails at `if (!this.state.isFieldsValid) return` — before `this.validate('submit')` (the form-level validator) is reached. A form-level validator writes its errors onto fields (`{ fields }`), but `validateAllFields` can't clear or recompute them, so `isFieldsValid` stays `false` from the previous submit and the form-level validator never re-runs — gating the form off permanently (even with `canSubmitWhenInvalid: true`).

## Fix

Run the form-level validator right after `validateAllFields`, guarded by whether any field-level validator errored this submit (`validateAllFields` returns those errors). When none errored, it re-runs and recomputes/clears its own field errors; when field-level validators did error, it's skipped as before (preserving the existing "don't trigger form validators if field validators errored" behavior).

## Tests

Added a `form-core` unit test reproducing #2248 (form-level-only validator, fill fields, resubmit → errors clear and `onSubmit` runs). Full `form-core` (504) and `react-form` (126) suites pass; the existing field-validator-gating test still passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Form-level submit validation now reruns on each submission when field-level validation passes.
  * Stale field errors are cleared after previously invalid fields are corrected.
  * Valid forms can now submit successfully on a subsequent attempt.

### DIFF
--- a/.changeset/form-level-validator-resubmit.md
+++ b/.changeset/form-level-validator-resubmit.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/form-core": patch
+---
+
+Re-run a form-level `onSubmit` validator on every submit (when no field-level validator errored) so it can recompute and clear its own field errors. Previously a form configured with only a form-level validator could be permanently gated by stale errors, because `_handleSubmit` bailed on `isFieldsValid` before the form-level validator re-ran (#2248).
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -2465,7 +2465,17 @@ export class FormApi<
       this.baseStore.setState((prev) => ({ ...prev, isSubmitting: false }))
     }
 
-    await this.validateAllFields('submit')
+    const fieldErrors = await this.validateAllFields('submit')
+
+    // A form-level validator owns its field errors (via `{ fields }`), so it must
+    // re-run on every submit to recompute/clear them; otherwise stale errors from
+    // a prior submit keep `isFieldsValid` false and gate the form off permanently
+    // when there are no field-level validators to clear them (#2248). Only run it
+    // here when no field-level validator errored this submit — if the fields are
+    // already invalid on their own, the form validator is skipped as before.
+    if (fieldErrors.length === 0) {
+      await this.validate('submit')
+    }
 
     if (!this.state.isFieldsValid) {
       done()
@@ -2487,8 +2497,6 @@ export class FormApi<
       })
       return
     }
-
-    await this.validate('submit')
 
     // Fields are invalid, do not submit
     if (!this.state.isValid) {
--- a/packages/form-core/tests/FormApi.spec.ts
+++ b/packages/form-core/tests/FormApi.spec.ts
@@ -1944,6 +1944,43 @@ describe('form api', () => {
     )
   })
 
+  it('re-runs a form-level onSubmit validator on re-submission when only a form-level validator exists (#2248)', async () => {
+    const onSubmit = vi.fn()
+    const form = new FormApi({
+      defaultValues: { name: '', other: '' },
+      validators: {
+        onSubmit: ({ value }) => {
+          const fields: Record<string, string> = {}
+          if (!value.name) fields.name = 'Required'
+          if (!value.other) fields.other = 'Required'
+          return Object.keys(fields).length ? { fields } : undefined
+        },
+      },
+      canSubmitWhenInvalid: true,
+      onSubmit,
+    })
+    form.mount()
+
+    // 1) Submit empty -> the form-level validator flags both fields.
+    await form.handleSubmit()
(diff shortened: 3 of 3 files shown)


MILLERMARRU (commented):
Traced through `validateAllFields` and `validate` on main to make sure I understood what each one actually touches before judging the fix. `validateAllFields('submit')` calls every field's `validate` with `skipFormValidation: true`, so `fieldErrors` here is strictly field-level validator output, it never includes anything the form-level `{ fields }` validator wrote. `validate('submit')` is the one that actually runs `options.validators.onSubmit` against the whole form and applies its `fields` map onto each field's error map. That separation is what makes the fix work.

The actual deadlock in #2248 makes sense once you see that `isFieldsValid` is a stale aggregate: after the first submit, the form-level validator writes `Required` into `fieldMeta.name.errorMap.onSubmit` and `fieldMeta.other`. On the second submit, there are no field-level validators to touch those entries, so `isFieldsValid` is still reading last submit's form-level output at the point the old code checked it, before `validate('submit')` ever got a chance to re-run and overwrite it. Gating the early-return on `this.state.isFieldsValid` was gating on data the form-level validator itself was supposed to refresh.

Switching the condition to `fieldErrors.length === 0` fixes it because that's always this submit's fresh result, not a carried-over flag. And behaviorally nothing changes for the case that matters, when a field-level validator does fail, `fieldErrors.length > 0` skips the form-level validator exactly like the old `isFieldsValid` gate did, since a field-level failure alone was already enough to fail `isFieldsValid` before. The only path that changes is the one that was actually broken, field-level clean but form-level state stale, and now the form-level validator gets to run and clear its own mess before the gate checks anything.

I also want to flag that the second `await this.validate('submit')` further down wasn't duplicated, it was moved, so the form-level validator still runs at most once per submit attempt in the passing path. That's an easy thing to get subtly wrong when relocating a call like this, and it's clean here.

Test is solid, it reproduces the exact regression path (submit empty, see fields get flagged by the form-level validator, fill the fields, submit again) and asserts both that the stale errors clear and that `onSubmit` actually fires, which is the part that would have silently failed forever under the old code without ever throwing.

FrancoKaddour:
Thanks for the careful trace @MILLERMARRU — you nailed exactly why gating on this submit's fresh `fieldErrors` instead of the stale `isFieldsValid` aggregate is what lets the form-level validator clear its own errors. Appreciate the second set of eyes.

PR #2261, opened 2026-07-31
fix(form-core): support value types with a custom equals() method in deep equality

### LINKED ISSUE #2195: Value equality does not handle Temporals

FrancoKaddour:
Closes #2195

## Problem

`evaluate` (the deep-equality helper used for dirty-checking) reports two equal value-type instances as unequal. Value types like `Temporal.PlainDate`, Luxon `DateTime`, etc. expose their state through private fields/getters, so they have no own enumerable keys and a non-plain prototype — which hits the no-enumerable-keys guard added in #2140 and returns `false`. A field backed by such a value is then considered dirty even after it's reset to its original value.

## Fix

Before that guard, delegate to a custom `equals()` method when both operands share a constructor that exposes one. This covers Temporal types, Luxon `DateTime`, Immutable collections, and any value type following the `equals()` convention, while leaving plain objects/arrays and the existing `Date`/`Map`/`Set` cases untouched.

## Tests

Added a `form-core` unit test using a value type with a private field and an `equals()` method (mirroring Temporal's shape — no own enumerable keys). Reverting the fix fails it; the full `form-core` suite passes, plus typecheck and lint.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved form value comparisons for objects with custom `equals()` methods.
  * Added support for value types such as Temporal and Luxon `DateTime` instances, including objects with private state.
  * Ensured equal instances compare correctly while differing instances remain distinct.

### DIFF
--- a/.changeset/evaluate-custom-equals.md
+++ b/.changeset/evaluate-custom-equals.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/form-core': patch
+---
+
+Support value types that expose a custom `equals()` method (Temporal types, Luxon `DateTime`, etc.) in the deep-equality check. Previously two equal instances were reported as unequal because they have no own enumerable keys, so a field backed by such a value stayed dirty even after being reset to its original value (#2195).
--- a/packages/form-core/src/utils.ts
+++ b/packages/form-core/src/utils.ts
@@ -519,9 +519,13 @@ export function evaluate<T>(objA: T, objB: T) {
   // exposes one. This must run before the no-enumerable-keys guard below, which
   // would otherwise treat every such value as unequal since they expose their
   // state through getters/private fields rather than own enumerable keys (#2195).
+  // Both operands must expose the method: plain object literals also share the
+  // `Object` constructor, so a one-sided `equals` (e.g. `{ equals: () => true }`
+  // vs `{}`) must fall through to structural comparison instead of delegating.
   if (
     objA.constructor === objB.constructor &&
-    typeof (objA as { equals?: unknown }).equals === 'function'
+    typeof (objA as { equals?: unknown }).equals === 'function' &&
+    typeof (objB as { equals?: unknown }).equals === 'function'
   ) {
     return (objA as unknown as { equals: (other: unknown) => boolean }).equals(
       objB,
--- a/packages/form-core/src/utils.ts
+++ b/packages/form-core/src/utils.ts
@@ -514,6 +514,20 @@ export function evaluate<T>(objA: T, objB: T) {
     return true
   }
 
+  // Delegate to a custom value-equality method (Temporal types, Luxon DateTime,
+  // Immutable collections, etc.) when both operands share a constructor that
+  // exposes one. This must run before the no-enumerable-keys guard below, which
+  // would otherwise treat every such value as unequal since they expose their
+  // state through getters/private fields rather than own enumerable keys (#2195).
+  if (
+    objA.constructor === objB.constructor &&
+    typeof (objA as { equals?: unknown }).equals === 'function'
+  ) {
+    return (objA as unknown as { equals: (other: unknown) => boolean }).equals(
+      objB,
+    )
+  }
+
   const keysA = Object.keys(objA)
   const keysB = Object.keys(objB)
 
--- a/packages/form-core/tests/utils.spec.ts
+++ b/packages/form-core/tests/utils.spec.ts
@@ -677,6 +677,30 @@ describe('evaluate', () => {
     expect(undefinedFalse).toEqual(false)
   })
 
+  it('uses a custom equals() method for value types like Temporal (#2195)', () => {
+    // Value types (Temporal.PlainDate, Luxon DateTime, ...) expose their state
+    // through private fields/getters, so they have no own enumerable keys and a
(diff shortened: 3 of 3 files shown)


MILLERMARRU (commented):
Pulled the current `evaluate()` from main to check the guard ordering, since that's what actually determines whether this fixes #2195 or not. The zero-own-keys guard a few lines below (`keysA.length === 0 && ... prototype !== Object.prototype`) is exactly what was returning `false` for two equal Temporal/Luxon instances before this patch, they expose state through private fields and getters so `Object.keys()` is empty on both, and that guard was written assuming empty keys on a non-plain object means "can't compare, treat as unequal." Putting the `equals()` delegation above that guard is the only place it can go for the fix to matter, since by the time you reach the empty-keys check it's already too late. Confirmed that's where it landed.

I see CodeRabbit's earlier comment about requiring both `objA.equals` and `objB.equals` to be callable before delegating, that's already handled in the version I'm looking at, both are checked with `typeof ... === 'function'`, and there's a dedicated test for the one-sided case now (`{ equals: () => true }` vs `{}` in both directions). Good that it's covered both ways since plain objects share the `Object` constructor, so the constructor check alone wouldn't have caught a lopsided `equals`.

Requiring `objA.constructor === objB.constructor` in addition to both having `equals` is the right conservative choice, it avoids letting an `equals()` from one class accidentally get called with an incompatible instance of a different class that happens to also define `equals` with different semantics.

The test class using a private field (`#iso`) rather than a plain property is a good choice too, it's what actually reproduces the "no own enumerable keys" condition instead of just simulating it, so the test would have caught a regression in the old code path honestly rather than by coincidence.

FrancoKaddour:
Thanks for the careful trace @MILLERMARRU — you're right that the `equals()` delegation only works above the empty-keys guard, and that requiring both operands' `equals` plus the matching constructor is what keeps it conservative. Appreciate the second set of eyes.

PR #2269, opened 2026-08-03
perf(form-core): extract meta markers to state object (v2)

43081j:
In order to compute if we can re-use the previous meta, we currently tag
two properties onto the meta (`derivedMetaSourceKey`,
`derivedMetaCanDisplayErrorsKey`).

This is fairly expensive because it means we change the shape of an
otherwise consistently structured object each time the value changes.

We only make use of this when the value changes, so this change switches
to using a `markers` object which transitions between values.

Basically this pattern:

```ts
const markers = { ... };

createAtom((prev) => {
  // in here, mutate `markers` when prev != curr
});
```

This means the underlying meta object never changes shape.

Also important to note, this now means `getFieldSnapshot` doesn't even
pass through this code path anymore.

NOTE: there's also a bit of a drive-by in here. I updated
`nameToFieldNodeSegments` to slice between separators instead of
appending characters one-by-one. Much faster, but can be split into its
own PR if needed.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/easy-rice-live.md
+++ b/.changeset/easy-rice-live.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/form-core': patch
+---
+
+Improve field snapshot and metadata re-use strategy.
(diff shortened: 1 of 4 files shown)


LeCarbonator:
Thanks!

PR #2273, opened 2026-08-06
[WIP] [v2] svelte adapter

crutchcorn:
This PR adds an initial Svelte adapter:

- [x] Impl
- [x] Tests
- [x] Docs
- [x] Examples
- [ ] Manual review


LeCarbonator (approved):
LGTM! General thoughts after all adapters are merged:

Field group types, functions and contract are mostly repeats from React <-> Svelte. Makes sense, but it should be centralized.

PR #2307, opened 2026-08-09
[v2] Bump devtools dependencies

LeCarbonator:


LeCarbonator:
This seems very flaky with the current setup of sherif. Canceling it for now.

PR #2332, opened 2026-08-13
fix: type serverValidate decoded values

### LINKED ISSUE #1438: Type inference from serverValidate is misleading
### Describe the bug

The type of the response from `serverValidate` (from a `createServerValidate`) in the example next server actions app has the `age` field as `number`, but if you access age at runtime it will be a string.

https://stackblitz.com/github/tanstack/form/tree/main/examples/react/next-server-actions?embed=1&theme=light&preset=node&file=src/index.tsx

### Steps to reproduce

1. Open stackblitz
2. Check the type that ts expects for `validatedData`: 

Image

3. See at runtime that printed age is actually a string:

Image

### Expected behavior

I'd expect that the type of `validatedData` should be remain a `FormData` since it doesn't actually parse the fields into their typescript types.

Another option could be that all the fields of `validatedData` are strings?

None

### Screenshots or Videos

_No response_

### Platform

macOS
chrome 135

### TanStack Form adapter

None

### TanStack Form version

1.6.3

### TypeScript version

5.8.2

### Additional context

_No response_

kkkhs:
## 🎯 Changes

Fixes #1438. `createServerValidate` exposed form-value types instead of decoded form data.

Fix: Return decoded `decode-formdata` value types from the public adapter API, and keep form-shaped casts only for validator and `ServerFormState` internals.

Test: Added Next.js and Remix adapter type regression coverage.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/quiet-lamps-warn.md
+++ b/.changeset/quiet-lamps-warn.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/react-form-nextjs': patch
+'@tanstack/react-form-remix': patch
+---
+
+Fix `createServerValidate` return types to reflect decoded `FormData` values instead of inferred form values.
--- a/packages/react-form-nextjs/src/createServerValidate.ts
+++ b/packages/react-form-nextjs/src/createServerValidate.ts
@@ -99,12 +99,11 @@ export const createServerValidate =
       })
     }
 
-    const values = (info
-      ? decode(formData, info)
-      : decode(formData)) as never as TFormData
+    const values = info ? decode(formData, info) : decode(formData)
+    const validationValues = values as never as TFormData
 
     const onServerError = (await runValidator({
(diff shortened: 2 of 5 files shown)


MILLERMARRU (commented):
I read through createServerValidate.ts in both react-form-nextjs and react-form-remix and the change is identical in each, which is good since they're meant to stay in sync. The split between `values` (untyped decoded data) and `validationValues` (cast for the validator call) is the right fix, the validator legitimately needs `TFormData` shape to type-check against the user's validation function, but claiming the same type for what callers get back from `serverValidate()` or catch in `ServerValidateError.formState.values` was misleading since `decode()` doesn't actually enforce that shape at runtime. The new `.test-d.ts` files cover exactly this distinction with `expectTypeOf`.

PR #2356, opened 2026-08-24
[v2] Fix: Keep ArrayField in sync after array replacement

scttcper:
`ArrayField` only reacts to array length and `_arrayVersion`, so replacing an array with a different array of the same length updated form state without rerendering the array container. UI rendered from the array could stay stale while submission used the new values.

This bumps `_arrayVersion` for direct same-length replacements. Array helpers now rely on length changes or this shared behavior instead of manually updating the version, while nested field updates still avoid rerendering the parent `ArrayField`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Array fields now update reliably when replaced with a new array of the same length.
  - Array components rerender when their array value changes, while avoiding unnecessary rerenders for nested field updates.
  - Swapping, moving, and clearing array items now maintain consistent field updates.

- **Tests**
  - Added coverage for array replacement through direct values and updater functions, nested updates, and array helper operations.
  - Added coverage confirming correct rerender behavior for array fields.

### DIFF
--- a/packages/form-core/src/FormApi/FormApi.lib.ts
+++ b/packages/form-core/src/FormApi/FormApi.lib.ts
@@ -502,7 +502,22 @@ export class InternalFormApi<
     updateOptions.fieldApiOverride = field
 
     batch(() => {
-      this._atoms.values.set((prev) => setBy(prev, fieldName, updater))
+      const previousValue = this.getFieldValue(fieldName)
+      const nextValue = callUpdater(updater, previousValue)
+      const replacedSameLengthArray =
+        Array.isArray(previousValue) &&
+        Array.isArray(nextValue) &&
+        previousValue !== nextValue &&
+        previousValue.length === nextValue.length
+
+      this._atoms.values.set((prev) => setBy(prev, fieldName, nextValue))
+
+      if (field && replacedSameLengthArray) {
+        field._setMeta((prev) => ({
+          ...prev,
+          _arrayVersion: prev._arrayVersion + 1,
+        }))
+      }
 
       this._notifyFieldChange(field, updateOptions)
     })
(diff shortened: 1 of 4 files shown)


packages/react-form/tests/FormField.spec.tsx
```diff
@@ -101,6 +101,53 @@ describe('Form fields', () => {
     expect(input).toHaveValue('new-value')
   })
 
+  it('rerenders an ArrayField for a same-length replacement', () => {
```
scttcper, line 104: this test does fail without the change

packages/form-core/src/FormApi/FormApi.lib.ts
```diff
@@ -502,7 +502,22 @@ export class InternalFormApi<
...
-      this._atoms.values.set((prev) => setBy(prev, fieldName, updater))
+      const previousValue = this.getFieldValue(fieldName)
+      const nextValue = callUpdater(updater, previousValue)
+      const replacedSameLengthArray =
+        Array.isArray(previousValue) &&
+        Array.isArray(nextValue) &&
+        previousValue !== nextValue &&
+        previousValue.length === nextValue.length
+
+      this._atoms.values.set((prev) => setBy(prev, fieldName, nextValue))
```
LeCarbonator: Minor correction, since `setBy` calls `callUpdater` already, we should wrap it so that it doesn't wrongly resolve function values. Doubt there's anyone that actually uses functions as field value, but oh well.


```suggestion
      this._atoms.values.set((prev) => setBy(prev, fieldName, () => nextValue))
```

LeCarbonator (approved):
Noticed a small correctness bug, but apart from that, looks good! It might change a bit in the future to make the work of the adapters a bit easier, but that's not a concern atm. Thanks for the fix!

PR #2358, opened 2026-08-24
[v2] Perf: Use one atom subscription per Field

scttcper:
Each React `Field` subscribed to the same atom twice, once for value and once for meta. The field atom already preserves state identity when neither changes, so use one whole-state subscription instead.

This cuts active subscriptions from two to one per field with the same render behavior. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved form field subscription handling so updates are tracked consistently.
  * Reduced redundant subscriptions while preserving existing field API behavior.
  * Improved subscription efficiency for more predictable form updates.

* **Tests**
  * Updated coverage to verify that each field uses a single active subscription.
  * Clarified subscription behavior in field-context tests.

### DIFF
--- a/packages/react-form/src/ReactForm/fieldSubscriptions.lib.ts
+++ b/packages/react-form/src/ReactForm/fieldSubscriptions.lib.ts
@@ -5,7 +5,7 @@ import type { AnyInternalFieldApi } from '@tanstack/form-core/internals'
 export function useValueFieldSubscription(
   fieldApi: AnyInternalFieldApi,
 ): AnyInternalFieldApi {
-  const state = useSelector(fieldApi.atom)
+  const state = useSelector(fieldApi.atom, state => state)
 
   return useMemo(() => {
     void state
--- a/packages/react-form/src/ReactForm/fieldSubscriptions.lib.ts
+++ b/packages/react-form/src/ReactForm/fieldSubscriptions.lib.ts
@@ -5,7 +5,7 @@ import type { AnyInternalFieldApi } from '@tanstack/form-core/internals'
 export function useValueFieldSubscription(
   fieldApi: AnyInternalFieldApi,
 ): AnyInternalFieldApi {
-  const state = useSelector(fieldApi.atom, state => state)
+  const state = useSelector(fieldApi.atom, (state) => state)
 
   return useMemo(() => {
     void state
--- a/packages/react-form/src/ReactForm/fieldSubscriptions.lib.ts
+++ b/packages/react-form/src/ReactForm/fieldSubscriptions.lib.ts
@@ -5,14 +5,12 @@ import type { AnyInternalFieldApi } from '@tanstack/form-core/internals'
 export function useValueFieldSubscription(
   fieldApi: AnyInternalFieldApi,
 ): AnyInternalFieldApi {
-  const value = useSelector(fieldApi.atom, (state) => state.value)
-  const meta = useSelector(fieldApi.atom, (state) => state.meta)
+  const state = useSelector(fieldApi.atom)
 
   return useMemo(() => {
-    void meta
-    void value
+    void state
     return Object.create(fieldApi)
-  }, [fieldApi, meta, value])
+  }, [fieldApi, state])
 }
 
 export function useArrayFieldSubscription(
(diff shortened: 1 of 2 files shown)


LeCarbonator:
You mention performance for the PR. Do you have some rough numbers for it? Memory / runtime?

scttcper:
@LeCarbonator all the perf is just from going from 2 to 1

I ran a focused benchmark using the exact before/after hooks with production React 19.2.8. Each variant ran in a fresh process 10 times, alternating order; results are medians.

### 1,000 mounted fields

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Active subscriptions | 2,000 | 1,000 | **−50%** |
| Retained heap | 7.50 MiB | 4.39 MiB | **−41.5%** |
| Mount time | 5.39 ms | 4.99 ms | **−7.4%** |
| Update all fields | 1.08 ms | 0.64 ms | **−40.1%** |
| Update one field | 4.8µs | 4.4µs | **−9%** |
| Renders per update | 1 | 1 | No change |

The harness mounts non-array fields backed by individual TanStack Store atoms, measures synchronous mount/update work, and measures retained heap after forced GC.

This isolates subscription overhead rather than measuring a complete application, so the absolute timings are directional. The consistent result is half the subscriptions and callbacks, with roughly 3.2 KiB less retained heap per mounted field.

crutchcorn:
Sorry, don't have a ton of cycles here to respond more in depth, but I'd attempted this in v1 and benchmarks showed MASSIVE improvements then (like 10x), but the UX of 1000 mounted fields slowed to a crawl for reasons I couldn't identify.

Might be helpful for us to all sit down and discuss and on our end manually test.

Not saying that this is a bad PR or approach, just flagging that benchmarks for this has been directionally wrong before.

scttcper:
Let sol on ultra try to figure out what you were referring to, obviously feel free to review/close whatever.

**Model used: OpenAI Codex — GPT-5.6 Sol**

Codex investigated the relevant v1 history and benchmarked the exact base and PR commits in production Chromium.

The closest v1 match is PR #2036. That implementation replaced seven subscriptions with one selector that created a new seven-property object and used a custom comparator. It was also mixed with substantial core store and notification changes, so the historical results do not isolate subscription consolidation as the cause. Its “1,000-field” benchmarks actually rendered 100 fields in jsdom and did not measure browser rendering or memory.

V2 has a different design. Before this PR, a normal field subscribes separately to `state.value` and `state.meta`. The PR subscribes once to the existing `{value, meta}` snapshot. V2 core preserves that snapshot’s identity whenever both references are unchanged:

```ts
if (prev?.meta === meta && prev.value === value) {
  return prev
}
```

Therefore, the rerender conditions are equivalent:

```text
before: value changed OR meta changed
after:  field snapshot changed
```

The PR does not create a projected object, add a custom comparator, or change core notification behavior. `ArrayField` is also untouched.

Real-browser confirmation with 1,000 controlled fields and 60 fresh contexts per variant found:

* Mount: **7.5% faster**
* Typing to commit: **12.2% faster**
* Typing p95: **12.7% faster**
* JavaScript heap: **15.4% lower**

crutchcorn:
@scttcper I'm an idiot and I wasted your tokens.

I just reviewed the diff quickly. For some reason I'd gotten a LOT wrong in my mind about what this PR was and assumed it was doing something more closely related to splitting one store into many for modularity. Frankly I don't know how I got things this mixed up.

So sorry.

This PR LGTM broadly, but might have weird consequences with Compiler. So long as it works well there we should merge.

Sorry again, that's embarrassing.

scttcper:
@crutchcorn work pays for it 🤷

LeCarbonator:
I think this should be compiler safe. Don't see how changing it from two to one could have consequences for it.

I wish I had better Wi-Fi to actually try this, but that'll have to wait until next monday.

packages/react-form/src/ReactForm/fieldSubscriptions.lib.ts
```diff
@@ -5,14 +5,12 @@ import type { AnyInternalFieldApi } from '@tanstack/form-core/internals'
 export function useValueFieldSubscription(
   fieldApi: AnyInternalFieldApi,
 ): AnyInternalFieldApi {
-  const value = useSelector(fieldApi.atom, (state) => state.value)
-  const meta = useSelector(fieldApi.atom, (state) => state.meta)
+  const state = useSelector(fieldApi.atom)
```
LeCarbonator: Nitpick, I prefer explicit selectors over the default fallback.
```suggestion
  const state = useSelector(fieldApi.atom, state => state)
```

LeCarbonator (approved):
LGTM!

PR #2362, opened 2026-08-29
fix(form-core): clear errorSourceMap.onMount alongside errorMap.onMount

### LINKED ISSUE #2294: setFieldValue clears errorMap.onMount but leaves errorSourceMap.onMount stale
### Describe the bug

`FormApi.setFieldValue` clears a field's stale `onMount` error when the field's own value changes, but only clears it from `errorMap`, not from `errorSourceMap`, so the two maps can end up out of sync.

```ts
// FormApi.ts, setFieldValue
  ...prev,
  isTouched: true,
  isDirty: true,
  errorMap: {
    ...prev?.errorMap,
    onMount: undefined,
  },
}))
```

`errorSourceMap.onMount` is never touched here. Both maps are surfaced to consumers directly, for example `useField` in `react-form` exposes `errorSourceMap: reactiveMetaErrorSourceMap` as part of `field.state.meta` alongside `errorMap`, so this isn't just internal bookkeeping, application code can read `meta.errorSourceMap.onMount` directly.

After a value change clears a field's `onMount` error, `meta.errorMap.onMount` correctly becomes `undefined`, but `meta.errorSourceMap.onMount` keeps whatever source (`'form'` or `'field'`) was last recorded there. Code that checks `errorSourceMap.onMount` to decide something (which validator layer an error came from, for instance) would see a source for an error that no longer exists in `errorMap`.

### Steps to reproduce

```ts
const form = new FormApi({
  defaultValues: { name: '' },
  validators: {
    onMount: () => 'Required',
  },
})
const field = new FieldApi({ form, name: 'name', validators: { onMount: () => 'Required' } })
form.mount()
field.mount()

console.log(field.state.meta.errorMap.onMount) // 'Required'
console.log(field.state.meta.errorSourceMap.onMount) // 'field'

field.setValue('something')

console.log(field.state.meta.errorMap.onMount) // undefined, cleared
console.log(field.state.meta.errorSourceMap.onMount) // still 'field', not cleared
```

### Expected behavior

`errorSourceMap.onMount` should be cleared alongside `errorMap.onMount` in `setFieldValue`, the same way #2244 (currently open) clears both together for the linked-field revalidation case.

### Actual behavior

`errorSourceMap.onMount` is left stale after `setFieldValue` clears the corresponding `errorMap.onMount`.

alliasgher:
## 🎯 Changes

Fixes #2294.

`FormApi.setFieldValue` clears `errorMap.onMount` on the field's meta but leaves `errorSourceMap.onMount` untouched, so `field.state.meta.errorSourceMap.onMount` keeps reporting a source (`'field'` or `'form'`) for an error that no longer exists in `errorMap`. Every other field-meta write in `FormApi.ts` updates the two maps together, so this reads as an omission rather than intent.

This adds the matching `errorSourceMap` update inside the same `setFieldMeta` call, and two regression tests in `FormApi.spec.ts` covering the `'field'` source (field-level `onMount` validator) and the `'form'` source (form-level `onMount` validator emitting a field error).

`alpha` is not affected: `errorSourceMap` does not appear in any package on that branch, so there is nothing to port to v2.

Note that #2245 edits the same block for an unrelated reason. There is no functional overlap; happy to rebase if it lands first.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearing a field value now consistently removes its initial validation errors, including errors originating from field- or form-level validation.
  * Updated field state now stays synchronized after values are changed.

### DIFF
--- a/.changeset/olive-forms-clear.md
+++ b/.changeset/olive-forms-clear.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/form-core': patch
+---
+
+Clear a field's `errorSourceMap.onMount` alongside `errorMap.onMount` when its value is set.
--- a/packages/form-core/src/FormApi.ts
+++ b/packages/form-core/src/FormApi.ts
@@ -2668,6 +2668,11 @@ export class FormApi<
             ...prev?.errorMap,
             onMount: undefined,
           },
+          errorSourceMap: {
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+            ...prev?.errorSourceMap,
+            onMount: undefined,
+          },
         }))
       }
 
--- a/packages/form-core/tests/FormApi.spec.ts
+++ b/packages/form-core/tests/FormApi.spec.ts
@@ -3233,6 +3233,70 @@ describe('form api', () => {
     expect(firstNameField.state.meta.errors).toEqual(['first name is required'])
   })
 
+  it("clears the onMount error source when a field's value is set", () => {
+    const form = new FormApi({
+      defaultValues: {
+        firstName: '',
+      },
+    })
+
+    form.mount()
+
+    const firstNameField = new FieldApi({
+      form,
+      name: 'firstName',
+      validators: {
+        onMount: () => 'first name is required',
+      },
+    })
+
+    firstNameField.mount()
+
+    expect(firstNameField.state.meta.errorMap.onMount).toBe(
+      'first name is required',
+    )
+    expect(firstNameField.state.meta.errorSourceMap.onMount).toBe('field')
+
+    form.setFieldValue('firstName', 'firstName')
+
+    expect(firstNameField.state.meta.errorMap.onMount).toBeUndefined()
+    expect(firstNameField.state.meta.errorSourceMap.onMount).toBeUndefined()
+  })
+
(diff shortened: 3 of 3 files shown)


MILLERMARRU (commented):
Fix looks right for #2294. I traced the other field-meta writes to check the "every other write updates both maps together" claim: `FieldApi.ts:889-890`, `1324-1328`, `1362-1366`, `1561-1566` all pair `errorMap` with `errorSourceMap`, and `FormApi.ts:2073-2077`, `2289-2293`, `2992-2996` do too. So `setFieldValue` was the one field-level omission and this closes it.

One spot that stays asymmetric after this PR: `FormApi.ts:1566`, `errorMap = Object.assign(errorMap, { onMount: undefined })` in the base-store derived state (`shouldInvalidateOnMount` branch). That clears the form-level `errorMap.onMount` with no matching `errorSourceMap` write. It's only fine if the form-level base store has no `errorSourceMap` field at all (the `errorSourceMap` writes I found are all field-meta, none form-level). Could you confirm that's the case, so it's clear this isn't a second instance of the same bug one scope up? Separately, that line mutates `currBaseStore.errorMap` in place rather than copying, which is a pre-existing smell in a `.state` read but not something this PR needs to touch.

PR #2380, opened 2026-09-11
fix(react-form,preact-form): prevent unnecessary FormGroup child re-renders

### LINKED ISSUE #2377: Field value change in FormGroup causes all fields in FormGroup to re-render
### Describe the bug

When using a `FormGroup`, changing the value of a single field within the group causes the entire FormGroup and all its child components to re-render, even if those components don't use or reference the changed field. This differs from the behavior of standalone `Field` components, where only the specific field re-renders when its value changes.

https://codesandbox.io/p/devbox/floral-dust-hlznnr?workspaceId=ws_C5KaZaVHbgfPuxgKAM9hsh

### Steps to reproduce

1. Preview the provided codesandbox which is based on the "React Example: Multi Step Wizard" example code.
2. Type in a value into one of the form fields of step 1.
3. View the console output or use the React dev tools setting "Highlight updates when components render" to see which components will re-render.

### Expected behavior

Similar to how individual `Field` components only re-render when their own value changes, a FormGroup should ideally only re-render the specific field that changed, not all child components within the group. This would allow heavy/complex components within a FormGroup to avoid unnecessary re-renders.

Every time

### Screenshots or Videos

Image

### Platform

OS: All platforms
Browser: All browsers
Version: N/A

### TanStack Form adapter

react-form

### TanStack Form version

1.33.5

### TypeScript version

_No response_

### Additional context

_No response_

dikshit-n:
## Summary

When a field value inside a `FormGroup` changes, all child components in that `FormGroup` were re-rendering — even components that had no dependency on the changed field. This is a significant performance regression for forms with many fields in a group.

## Root Cause

`useFormGroup` subscribed to the FormGroup's own store with the selector `(state) => state.value`. This selector returns the **entire group value object**, which gets a new reference every time **any** field in the group changes. Because the selector fires on every group-level store update, the `useMemo` that constructs `extendedFieldApi` always recomputes, causing a full re-render of all children.

## Fix

Subscribe to the **form's `baseStore`** with a **path-based selector** that reads only the specific group value at `opts.name`. Using `getBy(state.values, opts.name)` means:
- Changes to **fields inside this group** → selector returns a new reference → hook re-renders (correct)
- Changes to **fields in other groups** → selector returns the same value → no re-render (correct)

This mirrors the existing optimization pattern used for array-mode fields in `useField`.

## Changes

- **packages/react-form/src/useFormGroup.tsx**: Changed `useSelector` to read from `formGroupApi.form.baseStore` with a `getBy(state.values, opts.name)` selector
- **packages/preact-form/src/useFormGroup.tsx**: Same fix for the Preact adapter
- **packages/react-form/tests/useFormGroup.test.tsx**: Added regression test

## Testing

A new unit test (`should not re-render sibling field components when a field value changes`) verifies that changing a field in `step2` does not trigger a re-render of a field inside the `step1` FormGroup.

Closes [#2377](https://github.com/TanStack/form/issues/2377)

## Summary by CodeRabbit

* **Performance Improvements**
  * Form groups now update only when fields within the group change, reducing unnecessary re-renders caused by changes in sibling groups.

* **Bug Fixes**
  * Prevented unrelated field components from re-rendering when values change in a different form group.

* **Tests**
  * Added regression coverage to verify that sibling form groups remain unaffected by unrelated field updates.


PMudra:
This PR is clearly AI generated without a proper review and does not solve #2377.

PR #2385, opened 2026-09-11
fix(form-core): preserve validation counters across form reset

### LINKED ISSUE #2384: Pre-reset field validation completion clears a newer validation status
### Describe the bug

A field validation started before `form.reset()` can clear the validation status of a newer run when the older run finishes. A UI reading `field.state.meta.isValidating` can stop displaying its checking indicator while the new validation is still pending.

This is a follow-up to the [review discussion on #2382](https://github.com/TanStack/form/pull/2382#discussion_r3987479350). The counter race reproduces on that PR's base as well as its head; it is separate from the stale-error fix in #2381/#2382.

[Tests-only commit with unchanged main production code](https://github.com/lunaxislu/form/blob/ba6de4a83142fafcb74d5763f7e6574067019b14/packages/form-core/tests/FieldApi.spec.ts#L2257)

```sh
git clone https://github.com/lunaxislu/form.git
cd form
git checkout ba6de4a83142fafcb74d5763f7e6574067019b14
# Use Node 24.8.0 from .nvmrc and pnpm via Corepack.
corepack pnpm install --frozen-lockfile
corepack pnpm --dir packages/form-core exec vitest run tests/FieldApi.spec.ts -t 'should preserve pending validation when an older run settles'
```

The four cases use deferred promises and fake timers. The two reset cases fail; the two no-reset controls pass. Both direct and linked-field validation are covered.

### Steps to reproduce

1. Start validation A via a field's `onChangeAsync` and leave its promise pending.
2. Call `form.reset()`.
3. Blur the field to start validation B via `onBlurAsync` and leave B pending.
4. Resolve A.
5. Observe that the field's `isValidating` becomes false even though B is pending.

```text
A starts       count=1, isValidating=true
reset          count=0, isValidating=false
B starts       count=1, isValidating=true
A settles      count=0, isValidating=false  <-- B is still pending
B settles      count=0, isValidating=false
```

Using different validation causes avoids the same-cause debounce timer issue tracked by #2372/#2373. The linked-field variant starts A by changing another field listed in `onChangeListenTo`.

### Expected behavior

After reset, a completion belonging to a pre-reset run should not decrement the new run's counter. `field.state.meta.isValidating` should remain true until B settles. Without reset, overlapping runs should still decrement normally (2 → 1 → 0).

Every time with the controlled ordering above.

### Platform

macOS; reproduced in the form-core Vitest suite with jsdom, without a browser UI or server dependency.

### TanStack Form adapter

vanilla (`@tanstack/form-core`)

### TanStack Form version

1.33.5, main source at `57a855b42b9b99cee5ca12890b70c7df84e2982c`.

### TypeScript version

5.9.3

### Additional context

Reset replaces field metadata, including `_pendingValidationsCount`. The outer completion block of `FieldApi.validateAsync` still calls `endValidation()` against current metadata without distinguishing a pre-reset run. This applies to the main field and linked fields.

The report concerns completion accounting across a full form reset. It does not request changes to debounce scheduling, per-field reset, or form/group-level validation. Searches for reset/counter/isValidating and related reports did not reveal a separate report for this exact sequence beyond the #2382 review discussion.

lunaxislu:
## 🎯 Changes

Fixes #2384. Follow-up to the [counter race discussed in #2382](https://github.com/TanStack/form/pull/2382#discussion_r3987479350).

When validation A starts, the form resets, and validation B starts, A's eventual completion decrements the counter belonging to B. This makes the field report `isValidating: false` while B is still pending.

Track a private-by-convention validation generation on the form and advance it during full reset. Each field async run captures that generation at entry, before awaiting form validation. After that await, a stale run returns before discovering linked validators, registering counts, or replacing field controllers. Its final completion block only decrements the main and linked-field counters if the generation still matches. Each field validation timeout also records its ID and generation. Canceling a matching pre-reset timeout clears the timer without decrementing the new counter. Timers within the same generation keep their existing cancellation accounting. The ID check avoids assigning a field timeout generation to a timer installed by another path sharing the timeout slot, such as group validation.

This PR is based directly on main (`57a855b4`) and does not include or depend on the stale-error fix in #2382. It protects field validation continuation, completion, and pending-debounce cancellation across full form reset. The form validator itself, debounce timing, cleanup of already-fired timer IDs (#2373), per-field reset, and form/group-level validation are unchanged.

### Validation

Four parameterized completion cases cover direct and linked-field runs, each with and without reset. A uses `onChangeAsync` and B uses `onBlurAsync` to isolate this race from same-cause timer bookkeeping. They assert field validation status, the pending count, aggregate `isFieldsValidating`, and B's error when it finishes. Four additional cases cover replacing a still-pending same-cause debounce, again with direct/linked fields and with/without reset. They verify that only the replacement validator executes and its status stays pending through both debounce and async execution. Six cases cover a pending form-level validator: direct/linked fields with no reset, reset only, or reset followed by a new same-cause run. They verify that stale continuations do not start field validators or replace/abort the new controller, and that current work still applies its error and finishes its count.

| Verification | Result |
| --- | --- |
| Tests-only commit `ba6de4a8` on unchanged main production | 2 reset cases FAIL; 2 no-reset controls PASS |
| Completion cases with this fix | 4 PASS |
| Debounce cases before the review follow-up (head `7fbbc46d`) | 2 reset cases FAIL; 2 no-reset controls PASS |
| Debounce cases with the review follow-up | 4 PASS |
| Form-await cases before the latest follow-up (head `80628c69`) | 4 reset cases FAIL; 2 no-reset controls PASS |
| Form-await cases after the follow-up | 6 PASS |
| Full form-core suite | 519 PASS, 3 todo |
| `pnpm test:pr --parallel=3` | Passed for 58 affected projects, including adapter tests, TypeScript checks, lint and builds |
| `pnpm build:all` | All 14 package targets passed |
| Changed-file Prettier and `git diff --check` | PASS |

The required PR checks and builds used Node 24.8.0 and pnpm 11.21.0, with `NX_DAEMON=false NX_NO_CLOUD=true CI=true` for local execution. Existing warnings were emitted; no failed check remained. The changeset records the patch fix; no user-facing API usage changes require a guide update.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed asynchronous field validations incorrectly clearing the status of newer validation runs after a form reset.
  - Pending validation indicators now remain accurate during overlapping and debounced validations.
  - Stale validation results are discarded after resets and cannot trigger outdated field validation.
  - Validation errors remain associated with the latest field value and active validation run.


lunaxislu:
Addressed the [outside-diff review finding](https://github.com/TanStack/form/pull/2385#pullrequestreview-5177132116).

Field validation timers now record their ID and reset generation. Canceling a matching pre-reset timer still clears it, but does not decrement the new generation's counter. Same-generation debounce replacement retains its existing decrement. The ID comparison prevents attributing this bookkeeping to a different timer placed in the shared slot by another validation path.

Added four deferred-promise tests covering direct/linked fields with/without reset. Each replaces an onChange debounce before it fires, verifies that only the replacement validator runs, and checks the count/status through its debounce and async execution. Before this update the two reset cases fail while the two no-reset controls pass; afterward all four pass. The four existing completion tests also pass.

Validation: form-core 513 passing tests / 3 todo; full local `pnpm test:pr --parallel=3`, `pnpm build:all`, changed-file ESLint/Prettier, and `git diff --check` passed. The PR description and changeset now cover both completion and pending-debounce cancellation. Cleanup of already-fired timer IDs remains outside this change (#2373).

lunaxislu:
Addressed the [generation capture review](https://github.com/TanStack/form/pull/2385#pullrequestreview-5177371909).

`FieldApi.validateAsync` now captures the reset generation at entry, before awaiting `formValidationResultPromise`. It checks the generation immediately after that await and returns `[]` for stale work, before discovering linked validators, incrementing counts, or replacing/aborting field controllers. The existing completion and debounce-cancellation guards remain in place.

Added six tests using an actual deferred form-level `onChangeAsync`: direct/linked fields, each with no reset, reset only, or reset followed by a new same-cause validation. Before the change, the four reset cases fail because the stale continuation invokes a field validator; both no-reset controls pass. Afterward, all six pass, including assertions that the new controller remains unchanged and un-aborted, and that current validation applies its error and finishes its count normally.

All 14 tests added by this PR pass. Full validation: form-core 519 passing tests / 3 todo, `pnpm test:pr --parallel=3` across 58 affected projects, `pnpm build:all`, changed-file Prettier, and `git diff --check`. This guards field work waiting for form validation; it does not change the form-level validator's own result handling.

Issue #2390, opened 2026-09-14
[v2][vue] Field components registered with `createFormHook` don't rerender on field state changes, and lose their field API after `form.reset()`


awdr74100:
### Describe the bug

In `@tanstack/vue-form@2.0.0-alpha.2`, field components registered through `createFormHook` with `getFormHookHelpers().fieldComponent.strict()` (the pattern used in `examples/vue/large-form`) only render once:

1. **They don't rerender when the field state changes.** After a change that produces a validation error, the form state is correct (`form.getFieldMeta('name').errors` contains the error and `meta.isInvalid` is `true`), but the injected component keeps showing its first render: no error message and `aria-invalid="false"`. Rendering the same field directly inside the `<form.Field>` slot updates as expected.
2. **They lose their field API after `form.reset()`.** After a reset, typing into the injected component no longer updates the form value.

**Root cause**

- `wrapField` (`src/AppForm/fieldComponentHelpers.lib.ts`) injects the field API and passes the same object as a prop on every render. The parent `Field` rerenders its slot because of its own subscription, but Vue skips updating the injected component because its props are unchanged (same object reference), and the child's reads of the field API getters are not tracked by Vue.
- `createFieldComponent` (`src/VueForm/Components.lib.ts`) calls `provide(fieldContext, fieldApi.value)` once during setup. `useField` replaces `fieldApi.value` when `resetVersion` changes, so after `form.reset()` the injected component keeps the old instance. The comment above that line ("Field APIs are stable for a mounted name… the parent subscription handles state-driven renders") doesn't hold for injected components.

### Your minimal, reproducible example

Regression tests are included in PR #2392. Minimal Vue SFC repro:

```vue
<!-- TextField.vue — same as examples/vue/large-form/src/components/TextField.vue -->
<script setup lang="ts">
import type { FieldWithValue } from '@tanstack/vue-form'
defineProps<{ field: FieldWithValue<string>; label: string }>()
</script>

<template>
  <label :for="field.name">{{ label }}</label>
  <input
    :id="field.name"
    :value="field.value"
    :aria-invalid="field.meta.isInvalid"
    @input="field.handleChange(($event.target as HTMLInputElement).value)"
  />
  <p v-for="error in field.errors" :key="error.message">{{ error.message }}</p>
</template>
```

```vue
<!-- Form.vue -->
<script setup lang="ts">
import { createFormHook, getFormHookHelpers } from '@tanstack/vue-form'
import TextField from './TextField.vue'

[...]

Issue #2391, opened 2026-09-14
[v2][vue] `Field` keeps a stale field API after `ArrayField` `removeValue()` shifts items, so shifted errors aren't rendered


awdr74100:
### Describe the bug

With index keys (as in `examples/vue/array`), after `array.removeValue(0)` the form creates a new field API instance for `issues[0].title`, but the mounted `<form.Field :name="`issues[0].title`">` keeps the old instance. The form meta for `issues[0].title` contains the error that shifted from the removed row's neighbour, but the slot never renders it.

This happens with markup rendered directly inside the `<form.Field>` slot. It's unrelated to `createFormHook`.

Diagnosis (uses the internal `_tryGetFieldApi` only to compare instances):

| | Result |
|---|---|
| Before removal: slot `field` === `form._tryGetFieldApi('issues[0].title')` | `true` |
| After removal: internal instance replaced | `true` |
| After removal: slot `field` === internal instance | `false` |
| After removal: `form.getFieldMeta('issues[0].title').errors` | `["Required"]` |
| After removal: rendered errors | none |

`useField` (`src/VueForm/useField.lib.ts`) only recreates the field API when `form`, `name` or `resetVersion` changes. After `removeValue`, the name stays the same, so the component isn't moved to the new instance.

### Your minimal, reproducible example

```vue
<script setup lang="ts">
import { useForm } from '@tanstack/vue-form'

const form = useForm({
  defaultValues: { issues: [{ title: 'A' }, { title: '' }] },
  errorVisibility: () => true,
  validators: [
    {
      triggers: ['change'],
      run: ({ value, createErrorMap }) => {
        const errors = createErrorMap()
        value.issues.forEach((issue, i) => {
          if (!issue.title) (errors.fields as Record<string, string>)[`issues[${i}].title`] = 'Required'
        })
        return errors
      },
    },
  ],
})
</script>

<template>
  <form.ArrayField name="issues" v-slot="{ field: array }">
    <div v-for="(_, i) in array.value" :key="i">
      <form.Field :name="`issues[${i}].title`" v-slot="{ field }">
        <input :value="field.value" @input="field.handleChange(($event.target as HTMLInputElement).value)" />
        <p v-for="error in field.errors" :key="error.message">{{ error.message }}</p>
      </form.Field>
      <button type="button" @click="array.removeValue(i)">Remove</button>
    </div>
  </form.ArrayField>
</template>
```

### Steps to reproduce

[...]

PR #2392, opened 2026-09-14
[v2] Fix(vue): Rerender createFormHook field components and keep them connected after reset

awdr74100:
## 🎯 Changes

Fixes #2390.

Field components registered through `createFormHook` (`getFormHookHelpers().fieldComponent`) rendered once and then stopped updating:

- `wrapField` passed the same field API object on every render. The parent `Field` rerendered its slot, but Vue skipped updating the injected component because its props were unchanged, and the child's reads of the field API getters weren't tracked.
- `Field` provided `fieldApi.value` once during setup. `useField` replaces that API when `resetVersion` changes, so after `form.reset()` the injected component kept writing to the stale instance.

This PR:

- Makes `Field` provide a `computed` that returns a tracked view of the current field API (`trackFieldApi` in `fieldSubscriptions.lib.ts`) instead of the setup-time instance.
- `trackFieldApi` wraps the field API in a `Proxy`. Its `get` trap reads the field's existing subscription, so Vue tracks reads made while the injected component renders. Methods from the prototype chain are bound to the underlying field API, so they never run with the proxy as `this`. Own properties, including the attached field components, are returned unchanged. `set` forwards to the target.
- Makes `wrapField` read the computed while rendering, so a replaced field API (after reset) reaches the component as a new prop.
- Updates the comment in `createFieldComponent` that assumed field APIs were stable and that the parent subscription covered injected components.
- Adds two regression tests in `tests/adapter.spec.tsx`. Both fail on `alpha` and pass with this change:
  - `rerenders composed field components when field state changes`
  - `keeps composed field components connected after form reset`

Trade-off to review: the `field` prop that injected components receive is now a proxy, so `field === <slot field>` is `false`. Behaviour and types are unchanged. Other approaches I considered were forcing child updates with non-stable slots, or remounting with a changing `key`. Both felt more fragile.

I also verified the change against an external spike app (Vue 3.6.0-rc.8) that uses the `large-form` `TextField` pattern with no workarounds: submit-time errors, live revalidation, server errors from `createValidationError`, nested paths, SSR rendering and hydration all pass. One remaining failure there is a separate `ArrayField` removal issue, reported in #2391 and not addressed here.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Field components created with form hooks now update when field values or validation states change.
  - Field components remain connected to the latest field state after resetting a form.
  - Composed fields now keep values, errors, and accessibility states synchronized.

- **Tests**
  - Added coverage for reactive updates, form resets, validation errors, and `aria-invalid` state changes.


crutchcorn:
Superseded by https://github.com/TanStack/form/pull/2401

